PR #24, opened 2017-10-06
Adding node requirement

stevenernst131:


README.md
```diff
@@ -2,6 +2,7 @@
 # Azure Web Apps Support Center
 
 ## Prerequisites
+- [Node 6.*](https://nodejs.org/en/download/)
```
ShekharGupta1988: this is good. I guess my point was to add another section to troubleshoot build failures and add:
- try cleaning up node_modules and retry.
- make sure you are using node 6.* version or above.

These steps are not fairly obvious looking at the build errors we get from web pack.

PR #26, opened 2017-10-11
Add table view for event logs for aspnetcore

hforeste:
Add a table view of start up failures from the application event logs for aspnetcore applications.
image

### DIFF
--- a/app/availability/detector-view/detector-view.routeconfig.ts
+++ b/app/availability/detector-view/detector-view.routeconfig.ts
@@ -9,6 +9,7 @@ import { FrebAnalysisDetectorComponent } from './detectors/freb-analysis-detecto
 import { PhpLogAnalyzerComponent } from './detectors/php-log-analyzer-detector/php-log-analyzer-detector.component';
 import { CommittedMemoryUsageComponent } from './detectors/committed-memory-detector/committed-memory-detector.component';
 import { PageFileOperationsComponent } from './detectors/page-operations-detector/page-operations-detector.component';
+import { AspNetCoreComponent } from "./detectors/aspnetcore-detector/aspnetcore-detector.component";
 
 export const DetectorViewRouteConfig: Route[] = [
     {
@@ -58,6 +59,13 @@ export const DetectorViewRouteConfig: Route[] = [
         component: PageFileOperationsComponent,
         data: {
             navigationTitle: 'Page File Operations'
+        } 
+    },
+    { 
+        path: 'aspnetcore', 
+        component: AspNetCoreComponent,
+        data: {
+            navigationTitle: 'AspNetCore'
         } 
     },
     { 
--- a/app/availability/detector-view/detectors/aspnetcore-detector/aspnetcore-detector.component.html
+++ b/app/availability/detector-view/detectors/aspnetcore-detector/aspnetcore-detector.component.html
@@ -1,0 +1,17 @@
+<metric-graph [metricSets]="detectorMetrics" [title]="detectorMetricsTitle" [chartType]="metricsChartType" [yAxisLabel]="detectorMetricsYAxisLabel"></metric-graph>
+<span *ngIf="showMetadata">Startup failures captured in application event logs</span>
+<table *ngIf="showMetadata" class="table table-bordered" style="margin-top:8px;margin-bottom:5px;">
+    <thead>
+        <tr>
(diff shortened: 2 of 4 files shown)


ShekharGupta1988:
@hforeste - Hey buddy, why is the second graph empty in your screenshot? Shouldn't that contain error timeline (by error type)

app/availability/detector-view/detector-view.routeconfig.ts
```diff
@@ -61,6 +62,13 @@ export const DetectorViewRouteConfig: Route[] = [
         } 
     },
     { 
+        path: 'aspnetcore', 
+        component: AspNetCoreComponent,
+        data: {
+            navigationTitle: 'AspNetCore'
```
ShekharGupta1988: How about "AspNetCore Logs"

app/availability/detector-view/detectors/aspnetcore-detector/aspnetcore-detector.component.html
```diff
@@ -0,0 +1,17 @@
+<metric-graph [metricSets]="detectorMetrics" [title]="detectorMetricsTitle" [chartType]="metricsChartType" [yAxisLabel]="detectorMetricsYAxisLabel"></metric-graph>
+<span *ngIf="showMetadata">Startup failures captured in application event logs</span>
```
stevenernst131: What is the point of this line above?

ShekharGupta1988: one comment here : Should we have a more intutive way for the title of the table or is this is tooltip?
(same thing would need to be changed for php logs too)

app/availability/detector-view/detectors/aspnetcore-detector/aspnetcore-detector.component.ts
```diff
@@ -0,0 +1,33 @@
...
+    styles: ['.row { margin-top: 5px; }']
+})
+
+export class AspNetCoreComponent extends DetectorViewBaseComponent {
+
+    showMetadata: boolean;
+
+    constructor(protected _route: ActivatedRoute, protected _appAnalysisService: AppAnalysisService) {
+        super(_route, _appAnalysisService);
+        this.detectorMetricsTitle = "AspNetCore";
```
ShekharGupta1988: this I believe is graph title. How about AspNetCore Error Timeline

app/availability/detector-view/detectors/aspnetcore-detector/aspnetcore-detector.component.html
stevenernst131: @ShekharGupta1988  we are doing this change as a temporary fix for the demo today, but we want to do something more permanent later. 

If this is a title though, it is not clear to me

hforeste: This is purely a copy of how php log analyzer has been implementing its table view. As I understand it, it looks like it has some kind of description about the table. So I did the aspnetcore equivalent of a description to explain the table.

hforeste:
@ShekharGupta1988 , whats funny about that is that earlier it actually showed data in the production version and then moments later it stopped showing data. I'm not sure where that is coming from. So i need to look into that

Production view:
image

hforeste:
@ShekharGupta1988 , looks like data is showing in the table now. Not sure whats taking time for this. Need to look at this further.

image

app/availability/detector-view/detectors/aspnetcore-detector/aspnetcore-detector.component.html
stevenernst131: I guess this is fine for now. We can make it look better when we move to a common component

hforeste: Yea i agree. I still have my changes for a common component but it isn't complete yet. But for the sake of the demo today, I just get this in. And when we get back, get the refactored version in.

app/availability/detector-view/detector-view.routeconfig.ts
hforeste: Yea, perhaps AspNetCore Log Analyzer would be good

hforeste: Keeping parity with php and calling it "AspNetCore Logs" may be good but that sounds confusing to me. Perhaps they both should have "Log Analyzer" in their name

Issue #28, opened 2017-10-17
Update CPU View


stevenernst131:
Show descriptions of high CPU periods as well as solutions on the detector view page for CPU

Issue #23, opened 2017-10-05
Make sure timeline of downtimes lines up with the graph


stevenernst131:
![image](https://user-images.githubusercontent.com/7739017/31252879-84b91efc-a9d7-11e7-8f72-7080d054b6e4.png)

PR #29, opened 2017-10-17
Cpu view

stevenernst131:
#28 Update the CPU view to show problems statements and solutions
#23 Make sure downtime graph lines up with availability graph

If you come from application or performance analysis:
image

_Note that the instances from the downtime are selected above_

If you go directly into High CPU tool:
image


stevenernst131:
image

Issue #32, opened 2017-11-08
Update services to use new routes


stevenernst131:
Update the services to point to the new ARM routes

PR #33, opened 2017-11-08
Updating Routes and Adding Arm Cache

stevenernst131:
#32 I have updated to the new routes for diagnostic endpoints. 

I also added a cache service for all ARM calls, instead of just having a cache in app analysis service

### DIFF
--- a/app/shared/services/arm.service.ts
+++ b/app/shared/services/arm.service.ts
@@ -39,13 +39,44 @@ export class ArmService {
         return this._cache.get(url, request, invalidateCache);
     }
 
-    postResource<T>(resourceUri: string, body?: any, apiVersion?: string, invalidateCache: boolean = false): Observable<boolean | {} | ResponseMessageEnvelope<T>> {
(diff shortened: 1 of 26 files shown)


app/shared/services/arm.service.ts
```diff
@@ -22,188 +23,53 @@ export class ArmService {
...
-            search: null,
-            headers: this.getHeaders(etag),
-            body: body
-        });
-
-        return this._http.request(request);
-    }
-
-    // From Support Center Project
-    postArmResource(resourceId: string, body: any) {
```
ShekharGupta1988: Include Post and Put calls as - to get site app settings , its actually a POST call and not GET
To update app settings, its a PUT Call

stevenernst131: post is there? does that not work for you? I can add put

ShekharGupta1988: I can try the other POST method :) .

PR #38, opened 2017-11-14
Add Linux/Windows to all telemetry

stevenernst131:
Moving the calls to get this information into logging.service, which
means all the other logging services will enherit it

### DIFF
--- a/app/shared/services/logging/logging.service.ts
+++ b/app/shared/services/logging/logging.service.ts
@@ -1,7 +1,10 @@
 import { Injectable } from '@angular/core';
-import { PortalService, AuthService } from '../';
+import { PortalService, AuthService, ArmService } from '../';
 import { StartupInfo } from '../../models/portal';
 import { CommonLogEventType } from './events.enumerations';
+import { SiteExtensions, OperatingSystem, Site } from '../../models/site';
+import { IDiagnosticProperties } from '../../models/diagnosticproperties';
+import { ResponseMessageEnvelope } from '../../models/responsemessageenvelope';
 
 @Injectable()
 export class LoggingService {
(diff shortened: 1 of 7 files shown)


app/shared/services/logging/logging.service.ts
```diff
@@ -13,44 +16,53 @@ export class LoggingService {
...
 
                 let siteIndex = parts.indexOf('sites');
-                self._resourceName = siteIndex !== -1 ? parts[siteIndex + 1] : '';
+                this._resourceName = siteIndex !== -1 ? parts[siteIndex + 1] : '';
 
                 let providerIndex = parts.indexOf('providers');
-                self._resourceType = providerIndex !== -1 ? parts[providerIndex + 1] + '/' + parts[providerIndex + 2] : '';
+                this._resourceType = providerIndex !== -1 ? parts[providerIndex + 1] + '/' + parts[providerIndex + 2] : '';
+
+                this._armServiceInstance.getResource<IDiagnosticProperties>(this._startUpInfo.resourceId + '/diagnostics/properties').subscribe((properties: ResponseMessageEnvelope<IDiagnosticProperties>) => {
```
ShekharGupta1988: Couple of things:
1. Shouldn't this be **properties.appStack** instead of **properties.properties.appStack**. If you are getting back the envelope, then rename the subscribed result as response for better readability.

2.  Do a **undefined check** here,  if properties returned is null

app/shared/services/logging/logging.service.ts
```diff
@@ -72,6 +85,8 @@ export class LoggingService {
             Object.keys(args).forEach((key: string) => combinedArgs[key] = args[key]);
         }
 
+        console.log(combinedArgs);
```
ShekharGupta1988: remove this

PR #40, opened 2017-11-15
Better solutions

stevenernst131:


app/supportbot/message-flow/cpu-analysis-chat/cpu-analysis-chat-flow.ts
```diff
@@ -0,0 +1,90 @@
...
+    private resourceGroup;
+    private siteName;
+    private slotName;
+
+    constructor(private _appAnalysisService: AppAnalysisService, private _authService: AuthService) {
+        this._authService.getStartupInfo().subscribe(info => {
+            let pieces = info.resourceId.toLowerCase().split('/');
+            this.subscription = pieces[pieces.indexOf('subscriptions') + 1];
+            this.resourceGroup = pieces[pieces.indexOf('resourcegroups') + 1];
+            this.siteName = pieces[pieces.indexOf('sites') + 1];
```
ShekharGupta1988: maybe add this logic somewhere in common utilities class that can be utilized by anyone. I had a same functionality in one of my components (not yet checked in)

app/supportbot/common/graph-message/graph-message.component.ts
```diff
@@ -0,0 +1,77 @@
...
+    detectorMetricsTitle: string;
+    detectorMetricsDescription: string;
+    instanceDetailTitle: string;
+    instanceDetailDescription: string;
+
+    metricsChartType: ChartType = ChartType.lineChart
+
+    constructor(private injector: Injector, private _logger: BotLoggingService, private _route: ActivatedRoute,  private _appAnalysisService: AppAnalysisService) {
+        this.detectorMetricsTitle = "Overall CPU Usage per Instance";
+        this.detectorMetricsDescription = "This graphs shows the total CPU usage on each of the instances where your application is running. " +
```
ShekharGupta1988: for this to be generic, shoudnt these descriptions be coming from an input to this component?

ShekharGupta1988:
@stevenernst131  - Yo, I did noticed that when you do a "no" for cpu analysis, the flow seems to be broken. So that would require a fix.

app/supportbot/message-flow/cpu-analysis-chat/cpu-analysis-chat-flow.ts
stevenernst131: yes I was thinking the same thing, I thought you might be adding that and I was going to use whatever you did after you checked it in

app/supportbot/common/graph-message/graph-message.component.ts
stevenernst131: good catch, found the issue

stevenernst131: Yes I will make these components generic and move the inputs to the cpu detector flow service

PR #50, opened 2017-11-29
Adding Check for detector name for portexhaustion vs portrejections

puneetg1983:


stevenernst131:
This solution is not going to work. You would have to wait for the getDetectors response before making the call to getSummaryViewModel(this.ConnectionsRejections, ...)

puneetg1983:
Interesting, I thought that assigning it to an member variable (this.something) causes the two way binding to take care of it automatically as it worked locally at my end but I may be wrong then.

Since this is going to be throw away code , for now, shall I just move the call to getdetectorresponse in the subscribe block of the first call ???

stevenernst131:
I think if you pass this.something to a member function, then that value is passed to the function as a local variable, it would not just all the sudden change the value of a parameter in a function midway through execution. 

Yes I think that is a better solution.

puneetg1983:
lemme know if this looks ok

stevenernst131:
This should work

PR #54, opened 2017-12-04
TCP Connections Analysis - Constructing message from abnormalTimePeriod's payload

puneetg1983:
This is how it looks, open for any feedback \ suggestions

image

### DIFF
--- a/app/availability/analysis/tcpconnectionsanalysis/tcp-connections-analysis.component.ts
+++ b/app/availability/analysis/tcpconnectionsanalysis/tcp-connections-analysis.component.ts
@@ -8,6 +8,8 @@ import { Observable } from 'rxjs/Observable';
 import 'rxjs/add/operator/map';
 import { ActivatedRoute } from '@angular/router';
 import { AvailabilityLoggingService } from '../../../shared/services/index';
+import { MetaDataHelper } from '../../../shared/utilities/metaDataHelper';
+import  '../../../shared/polyfills/string'
 
 @Component({
     selector: 'tcpconnections-analysis',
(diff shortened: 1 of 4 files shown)


app/availability/analysis/tcpconnectionsanalysis/tcp-connections-analysis.component.ts
```diff
@@ -83,6 +92,49 @@ export class TcpConnectionsAnalysisComponent implements OnInit {
             });
     }
 
+    getAbnormalTimePeriodMessageFromMetadata(detectorName:string, detectorResponse:IDetectorResponse):string
+    {
+        let message = detectorResponse.abnormalTimePeriods[0].message;
```
stevenernst131: should you check that detectorResponse.abnormalTimePeriods.length > 0?

puneetg1983: Line 71 which is the condition when this function is called has that exact check

app/availability/analysis/tcpconnectionsanalysis/tcp-connections-analysis.component.ts
```diff
@@ -83,6 +92,49 @@ export class TcpConnectionsAnalysisComponent implements OnInit {
...
+        if (detectorName == this.OpenSocketCount)
+        {
+            let issueType = MetaDataHelper.getMetaDataValue(detectorResponse.abnormalTimePeriods[0].metaData, "IssueType");
+            let instance = MetaDataHelper.getMetaDataValue(detectorResponse.abnormalTimePeriods[0].metaData, "Instance");
+            let siteName = MetaDataHelper.getMetaDataValue(detectorResponse.abnormalTimePeriods[0].metaData, "SiteName");
+            let processName = MetaDataHelper.getMetaDataValue(detectorResponse.abnormalTimePeriods[0].metaData, "ProcessName");
+            let processId = MetaDataHelper.getMetaDataValue(detectorResponse.abnormalTimePeriods[0].metaData, "ProcessId");            
+            let handleCount = MetaDataHelper.getMetaDataValue(detectorResponse.abnormalTimePeriods[0].metaData, "HandleCount");
+
+            if (issueType == "HigSocketHandleCount" || issueType == "HighSocketHandleCount")
```
stevenernst131: in javascript you should always use '==='

https://stackoverflow.com/questions/359494/which-equals-operator-vs-should-be-used-in-javascript-comparisons

app/availability/analysis/tcpconnectionsanalysis/tcp-connections-analysis.component.ts
```diff
@@ -83,6 +92,49 @@ export class TcpConnectionsAnalysisComponent implements OnInit {
...
+            {
+                message = "<b>High Open Socket handle count</b> detected on instance - " + instance + ". ";
+            }
+            else if (issueType == "SocketHandlesLeaked")
+            {
+                message = "<b>Socket handle leak</b> detected on instance - " + instance + ". It was detected that the TCP Connections were not high on the instance however the open socket handle count on the instance was high.";
```
stevenernst131: comma before 'however'

app/shared/utilities/metaDataHelper.ts
```diff
@@ -5,16 +5,17 @@ export class MetaDataHelper {
 
     static getMetaDataValue(metaData: INameValuePair[][], key: string): string {
         let searchKey = key.toLowerCase();
+        let returnValue = null;
         if (metaData && metaData.length > 0) {
             metaData.forEach(element => {
                 element.forEach(nameValuePair => {
                     if (nameValuePair.name.toLowerCase() === searchKey) {
-                        return nameValuePair.value;
+                        returnValue = nameValuePair.value;
```
stevenernst131: why do you want to keep searching the metadata if you have already found the value. The purpose of this method is that you are only expecting one entry with the specified key.

puneetg1983: The return statement isn't really returning when it finds the element. The whole for each executes and ends up returning NULL even if the element is there. https://stackoverflow.com/questions/34653612/what-does-return-keyword-mean-inside-foreach-function/34653650. 

So we either change it completely or this was the way I found with minimal code change

stevenernst131: oh that is a good catch, I didn't realize that. Thanks.

PR #56, opened 2017-12-07
Profiling Solution Improvement

stevenernst131:
Use Server Farm Service to get the correct site, since if you rely on
the respond from the diag role, it will return the resource group of the
main site, but a neighboring site on the same server farm could have a
different resource group.

### DIFF
--- a/app/solutions/components/specific-solutions/profiling-solution/profiling-solution.component.ts
+++ b/app/solutions/components/specific-solutions/profiling-solution/profiling-solution.component.ts
@@ -2,7 +2,7 @@ import { Component, Input, OnInit, OnDestroy } from '@angular/core';
 import { SolutionBaseComponent } from '../../common/solution-base/solution-base.component';
 import { SolutionData } from '../../../../shared/models/solution';
 import { MetaDataHelper } from '../../../../shared/utilities/metaDataHelper';
-import { SiteService, DaasService, WindowService, AvailabilityLoggingService } from '../../../../shared/services'
+import { SiteService, DaasService, WindowService, AvailabilityLoggingService, ServerFarmDataService } from '../../../../shared/services'
 import { SiteProfilingInfo } from '../../../../shared/models/solution-metadata';
 import { Subscription } from 'rxjs';
 import { TimerObservable } from 'rxjs/observable/TimerObservable';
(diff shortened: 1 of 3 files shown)


stevenernst131:
@puneetg1983 there may be some other improvements we want to make, when I was testing the call to get current sessions timed out, I assume because scm site was not started. Maybe we can send a warm up request earlier?

stevenernst131:
I also removed the profiling solution from the factory so that when it starts being returned from CPU, it will not show up, since that is what you wanted.

app/solutions/components/specific-solutions/profiling-solution/profiling-solution.component.ts
```diff
@@ -46,31 +46,52 @@ export class ProfilingComponent implements SolutionBaseComponent, OnInit, OnDest
     scmPath: string;
     SessionCompleted:boolean;
 
-    constructor(private _siteService: SiteService, private _daasService: DaasService, private _windowService: WindowService, private _logger: AvailabilityLoggingService) {
+    couldNotFindSite:boolean = false;
+
+    constructor(private _siteService: SiteService, private _daasService: DaasService, private _windowService: WindowService, private _logger: AvailabilityLoggingService, private serverFarmService: ServerFarmDataService) {
```
puneetg1983: do you want to keep the same underscore convention? Change to _serverFarmService

app/solutions/components/specific-solutions/profiling-solution/profiling-solution.component.ts
```diff
@@ -46,31 +46,52 @@ export class ProfilingComponent implements SolutionBaseComponent, OnInit, OnDest
...
+            if (targetedSite) {
+                let siteName = targetedSite.name;
+                let slotName = '';
+                if (targetedSite.name.indexOf('(') >= 0) {
+                    let parts = targetedSite.name.split('(');
+                    siteName = parts[0];
+                    slotName = parts[1].replace(')', '');
+                }
+    
+                if (targetedSite) {
```
puneetg1983: repeated if (targetSite) ?

app/solutions/components/specific-solutions/profiling-solution/profiling-solution.component.ts
```diff
@@ -46,31 +46,52 @@ export class ProfilingComponent implements SolutionBaseComponent, OnInit, OnDest
...
+
+    constructor(private _siteService: SiteService, private _daasService: DaasService, private _windowService: WindowService, private _logger: AvailabilityLoggingService, private serverFarmService: ServerFarmDataService) {
     }
 
     ngOnInit(): void {
 
         this._logger.LogSolutionDisplayed('CLR Profiling', this.data.solution.order.toString(), 'bot-sitecpuanalysis');
         
-        this.siteToBeProfiled = MetaDataHelper.getProfilingData(this.data.solution.data);         
+        this.siteToBeProfiled = MetaDataHelper.getProfilingData(this.data.solution.data);
```
puneetg1983: Is this really required now ? you are assigning this.siteToBeProfiled later in this function. The only possibility is when targetedSite is undefined. Could that happen ?

stevenernst131: removed.

app/solutions/components/specific-solutions/profiling-solution/profiling-solution.component.ts
stevenernst131: removed

app/solutions/components/specific-solutions/profiling-solution/profiling-solution.component.ts
stevenernst131: yup

PR #69, opened 2018-01-04
Adding date selection for TCP Connections

puneetg1983:
Adding an option to view TCP Connection information for past days

image

Removing availability from this view as I couldnt find a way to get Availability graphs for older days...

### DIFF
--- a/app/availability/analysis/tcpconnectionsanalysis/tcp-connections-analysis.component.html
+++ b/app/availability/analysis/tcpconnectionsanalysis/tcp-connections-analysis.component.html
@@ -1,10 +1,26 @@
-<availability-graph 
-[(displayGraph)]="displayGraph" >
-</availability-graph>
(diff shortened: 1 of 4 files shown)


stevenernst131:
@puneetg1983 as I mentioned in the past, when there is a widget you are creating that can be used by multiple pages or in different places, in this case the date picker, I think we should create a common component for it. I know it is more work but we also don't want to have to redo this work on every page that needs a date picker. 

Also do we have a cohesive strategy around allowing date pickers on each page? Not that it is bad but it seems confusing to me to have it some places and not others. 

I am working on the observations view update so I will think about the date picker and how that can be included universally.

puneetg1983:
@stevenernst131 - Totally agreed on creating a datetime picker component. I might have gone that route if it hadn't taken a lot of time in understanding the whole ngOnChanges which took quite some time for me. 

By the time I figured all that out, I somehow lost my energy to add another component....

Still getting familiar with Angular....

PR #72, opened 2018-01-17
Profiler error handling , retries and showing errors for failed sessions

puneetg1983:


stevenernst131:
Please go though your code, even stuff that is not changed with this pull request and make sure that you are using '===' instead of '==' when comparing, just so we are consistent in using the correct practice. I know it can be easy to forget when coming from C#

puneetg1983:
done !!!

PR #74, opened 2018-01-17
changing WebApp to web app

puneetg1983:


app/solutions/components/specific-solutions/memorydump-solution/memorydump-solution.component.ts
```diff
@@ -57,15 +56,15 @@ export class MemoryDumpComponent implements SolutionBaseComponent, OnInit, OnDes
...
+    constructor(private _siteService: SiteService, private _daasService: DaasService, private _windowService: WindowService, private _logger: AvailabilityLoggingService, private _serverFarmService: ServerFarmDataService) {
     }
 
     ngOnInit(): void {
 
         this._logger.LogSolutionDisplayed('Memory Dump', this.data.solution.order.toString(), 'bot-sitecpuanalysis');
 
         this.siteToBeDumped = MetaDataHelper.getSiteDaasData(this.data.solution.data);
-        let siteInfo = MetaDataHelper.getSiteDaasData(this.data.solution.data); 
+        let siteInfo = MetaDataHelper.getSiteDaasData(this.data.solution.data);
```
stevenernst131: what are you using siteInfo for?

puneetg1983: Not sure if I understand but that variable is used here and few other places...

` let targetedSite = sites.find(site => site.name.toLowerCase() === siteInfo.siteName.toLowerCase());`

puneetg1983: This was actually added in your commit - https://github.com/Azure/Azure-WebApps-Support-Center/commit/b526cabf2190c362b4b2c9f109b1879e3f8d63d7#diff-aa25b366ee01b76c68bc4f3cf8a74561

stevenernst131: I added it because the site to be dumped may not be the siteInfo that we originally get, since it may be a different one in the server farm. I think the correct thing to do is to remove the line above this when you are assigning a value to siteToBeDumped, since it is being assigned to below.

puneetg1983: If I remove this and if targetedSite is NULL (not sure if it can ever be) , then siteToBeDumped will also be null and rest of the component will fail

stevenernst131: then that scenario should be handled. If App A is using high cpu or has slow responses, but they are in support center for App B, and we suggest doing a dump for App A, then it would also be wrong to just give the solution as App B if we are not able to find it. Does that make sense?

puneetg1983: I think I am not a 100% clear on the whole  this._serverFarmService.sitesInServerFarm.subscribe logic yet. I will connect with you some time to get a better understanding of this and maybe then this question will automatically get answered. For now, I removed this call from the memorydump component and the profiling component...

PR #76, opened 2018-01-24
Updated Observation View

stevenernst131:
image

image

image

image


app/availability/problem-solution/problem-solution.component.ts
```diff
@@ -0,0 +1,138 @@
...
+                break;
+            case 'pagefileoperations':
+            case 'committedmemoryusage':
+                title = 'Memory Issue';
+                break;
+            case 'autoheal':
+                title = 'Auto Heal';
+                break;
+            default:
+                title = 'App Issue'
```
ShekharGupta1988: Do you feel comfortable with this? If we make a detector public, then it will by default say 'App Issue'.
I dont have a good answer to this question. Can this be something we enforce in detector backend. (to choose a category or tag)

stevenernst131: Yeah that would be great, and it is something that we have to do for the generic model for sure. Unfortunately this is sort of a hack to just get this out, but definitely not a final solution.

PR #84, opened 2018-02-13
APIs + Compilation Cache + Local Source Watcher

ShekharGupta1988:
Following Changes:
- List/Get detectors API
- Compilation cache logic
- Local Source watcher for csx scripts
- Tenant Id service to fetch tenant ids and pass it to script
- Unit tests

### DIFF
--- a/src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
@@ -19,10 +19,9 @@ namespace Diagnostics.DataProviders
             _kustoClient = KustoClientFactory.GetKustoClient(configuration);
         }
 
-        public async Task<DataTableResponseObject> ExecuteQuery(string query, string stampName)
+        public async Task<DataTableResponseObject> ExecuteQuery(string query, string stampName, string requestId = null, string operationName = null)
         {
-            return await _kustoClient.ExecuteQueryAsync(query, stampName);
-            
+            return await _kustoClient.ExecuteQueryAsync(query, stampName, requestId, operationName);
         }
     }
 }
--- a/src/Diagnostics.ScriptHost/SourceWatcher/LocalFileSystem/LocalFileSystemSourceWatcherService.cs
+++ b/src/Diagnostics.ScriptHost/SourceWatcher/LocalFileSystem/LocalFileSystemSourceWatcherService.cs
@@ -1,0 +1,82 @@
+﻿using Diagnostics.ScriptHost.Models;
+using Diagnostics.ScriptHost.SourceWatcher.Interfaces;
+using Diagnostics.ScriptHost.Utilities;
+using Diagnostics.Scripts;
+using Diagnostics.Scripts.Models;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Win32;
+using System;
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+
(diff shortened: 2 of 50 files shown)


src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
```diff
@@ -19,10 +19,9 @@ public KustoDataProvider(OperationDataCache cache, KustoDataProviderConfiguratio
             _kustoClient = KustoClientFactory.GetKustoClient(configuration);
         }
 
-        public async Task<DataTableResponseObject> ExecuteQuery(string query, string stampName)
+        public async Task<DataTableResponseObject> ExecuteQuery(string query, string stampName, string requestId = null, string operationName = null)
```
stevenernst131: Should we call this ExecuteQueryAsync since it is an async method? I added this originally, but given we don't know what our auto complete will look like it may make sense to make sure people know it is an async method. I know that would require changing csx also so we don't have to do it right now.

ShekharGupta1988: Ya, We can add async in the function name

src/Diagnostics.ScriptHost/SourceWatcher/LocalFileSystem/LocalFileSystemSourceWatcherService.cs
```diff
@@ -0,0 +1,82 @@
...
+        private Task _completionTask;
+
+        public LocalFileSystemSourceWatcherService(IHostingEnvironment env, IConfiguration configuration, ICacheService<string, Tuple<Definition, EntityInvoker>> cacheService)
+        {
+            _cacheService = cacheService;
+
+            _sourceConfig = new SourceWatcherConfiguration();
+            if (env.IsProduction())
+            {
+                _sourceConfig.LocalSourceDirectory = (string)Registry.GetValue(HostConstants.ScriptSourceConfigRegistryRootPath, HostConstants.LocalSourceDirectoryKey, string.Empty);
```
stevenernst131: Non blocking: not sure how exactly this would work but might be nice in the future to start the process with these values set as environment variables, that way they could be added to the configuration by doing something like this:


public Startup (IHostingEnvironment env) {

            var builder = new ConfigurationBuilder()
                .SetBasePath(env.ContentRootPath)
                .AddJsonFile("appsettings.json", optional: false, reloadOnChange: true)
                .AddJsonFile($"appsettings.{env.EnvironmentName}.json", optional: true)
                .AddEnvironmentVariables(); // <-- This would add environment variables

            Configuration = builder.Build();
        }

This way, we could have an appsettings.json locally, but in georegion the environment variables populate the configuration. 

You add environment variables that look like this: "Section:key"

ShekharGupta1988: Do you mean :
1.) In onstart() Instead of adding configuration to registry, add environment variable
2) Read env variable instead of registry in script host

stevenernst131: yeah exactly, just thinking we don't want to have to write registry values on each of our dev boxes in order to use it locally.

ShekharGupta1988: so for local debugging right now, for dev boxes, you need to update app setting json file with the secrets (no need to write in registry). 
But there is a merit in using environment variables as once set, hopefully they can just continue to persist

PR #85, opened 2018-02-15
Fixing Memory Dump solution to handle errors and do retries

puneetg1983:
Just like the profiler solution

### DIFF
--- a/app/solutions/components/specific-solutions/memorydump-solution/memorydump-solution.component.ts
+++ b/app/solutions/components/specific-solutions/memorydump-solution/memorydump-solution.component.ts
@@ -98,6 +98,7 @@ export class MemoryDumpComponent implements SolutionBaseComponent, OnInit, OnDes
                             this.retrievingInstances = false;
                             this.instances = result;
                             this.checkRunningSessions();
+                            this.populateInstancesToDump();
                         },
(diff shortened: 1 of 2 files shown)


stevenernst131:
Looks good, one question: Do we want to log this failure? Just to know how common it is?

app/solutions/components/specific-solutions/memorydump-solution/memorydump-solution.component.ts
```diff
@@ -237,22 +248,26 @@ export class MemoryDumpComponent implements SolutionBaseComponent, OnInit, OnDes
...
 
     collectMemoryDump() {
         this.instancesToDump = new Array<string>();
 
-        this.InstancesSelected.forEach(x => {
-            if (x.Selected) {
-                this.instancesToDump.push(x.InstanceName);
-            }
-        });
+        if (this.InstancesSelected && this.InstancesSelected != null) {
```
ShekharGupta1988: use !==

puneetg1983:
Hi @stevenernst131 , I can add it to log when it errors, what would be the right function to use ? or can you point me to an example of where we are logging something like this (sorry tied up a bit and havent had a chance to go through the code yet)

PR #92, opened 2018-02-27
Angular cli migration

stevenernst131:


stevenernst131:
The folder structure is changing slightly which is why so many files are affected. Basically all the filed for running the app move to the src folder. Inside the src folder, the folder structure is unchanged. All of the files that are used for building, packaging or testing are contained in the project root. 

I have also added instructions for creating your own self signed cert, which solves the `[WDS] Disconnected` bug that we encountered when running this with ssl enabled. 

The biggest code change was removing the barrel (index.ts) file that was located in the services folder and was used for importing services. You will now have to import them all directly. VS code has really good intellisense for adding imports now.

PR #93, opened 2018-02-28
Adding More DAAS Tools and Improving UI for Profiler to enable collecting raw stack traces

puneetg1983:
1. Adding PHP Logs, PHP Report tools from DAAS 
2. Leveraging DAAS's feature to validate whether HTTP logs and PHP logs have been enabled or not
3.  Adding support for generating raw thread report in profiler
4. Removing caching for all DAAS components as they are reflecting incorrect data due to cache
5. Ensuring that all the DAAS tools show a warning if they site is not running in Dedicated Tiers

### DIFF
--- a/app/shared/components/daas-sessions/daas-sessions.component.html
+++ b/app/shared/components/daas-sessions/daas-sessions.component.html
@@ -1,4 +1,4 @@
-<div class="header2">Last 5 {{DiagnoserHeading}}</div>
+<div *ngIf="supportedTier" class="header2">Last 5 {{DiagnoserHeading}}</div>
     <div>
         <div class="col" *ngIf="checkingExistingSessions">
             <i class="fa fa-circle-o-notch fa-spin spin-icon" aria-hidden="true"></i>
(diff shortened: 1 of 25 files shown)


app/shared/components/daas-sessions/daas-sessions.component.html
```diff
@@ -1,4 +1,4 @@
-<div class="header2">Last 5 {{DiagnoserHeading}}</div>
+<div *ngIf="supportedTier" class="header2">Last 5 {{DiagnoserHeading}}</div>
```
stevenernst131: So it will just show nothing if not supported?

app/shared/components/daas/profiler.component.ts
```diff
@@ -42,12 +42,28 @@ export class ProfilerComponent implements OnInit, OnDestroy {
...
+                }               
+            }
+        }, error => {
+            //TODO: handle error
+        })
     }
 
     ngOnInit(): void {
+        
+        if (!this.supportedTier)
```
stevenernst131: One problem with how you are doing this is that the subscribe above many not finish before ngOnInit() is called. This means that you would have it saying that the tier is not supported, when in fact the server farm call had just not completed. It may never happen just because something else is making that server farm call, but you should be aware of it.

stevenernst131:
@puneetg1983 I am going to merge because I need to merge this with my pull request, which will change the folder structure (moving to angular cli). That makes the merging process a little more complicated so i can do it.

app/shared/components/daas-sessions/daas-sessions.component.html
puneetg1983: This component is a part of another component so it is fine if this one comes as empty as the other component has some text that will display when we are on shared or free SKU.

puneetg1983:
Thanks @stevenernst131 for merging this.

app/shared/components/daas/profiler.component.ts
puneetg1983: Ok, then I will change this part and may be send another pull once your CLI changes are merged.

PR #95, opened 2018-03-01
Menu update

stevenernst131:
Building on top of @puneetg1983 's changes for the menu.


src/app/shared/services/categories.service.ts
```diff
@@ -0,0 +1,145 @@
...
+
+        perf.Subcategories.push({
+            Name: 'Container Initialization',
+            BgColor: '#540d6e',
+            TextColor: 'White',
+            Href: 'availability/detectors/dockercontainerstartstop',
+            OperatingSystem: OperatingSystem.linux
+        });
+
+        let monitor = new Category();
```
ShekharGupta1988: What is the idea behind having this category here? Is it because these 2 tiles were getting hidden when we placed it under availability and perf?
Also one concern is it will be difficult for us in future to decide what goes into Monitor category and not in specific problem category (like  availability and perf or management)

src/app/shared/services/categories.service.ts
```diff
@@ -0,0 +1,145 @@
...
+        tools.Subcategories.push({
+            Name: 'CLR Profiling Tool',
+            BgColor: 'rgb(153, 51, 0)',
+            TextColor: 'White',
+            Href: '',
+            OperatingSystem: OperatingSystem.windows
+        });
+
+        tools.Subcategories.push({
+            Name: 'Database Tester',
```
ShekharGupta1988: Needs a better name :). @puneetg1983  knows about this.

src/app/shared/services/categories.service.ts
```diff
@@ -0,0 +1,145 @@
...
+            BgColor: 'rgb(102, 153, 0)',
+            TextColor: 'White',
+            Href: '',
+            OperatingSystem: OperatingSystem.windows
+        });
+
+        tools.Subcategories.push({
+            Name: 'Client Cert Checker',
```
ShekharGupta1988: same here

puneetg1983: Hi @shekhargupta1988, these are just 'test names' that I came up with to illustrate the menu page display. I am waiting on Jen's feedback on what should the real category names and tools names should be (have also looped you in that email)

src/app/shared/services/categories.service.ts
stevenernst131: Yeah I think Puneet moved them when he was playing around with the menu. We can just move them back, better to be consistent with the categories in case opening flow.

puneetg1983: Before the merge, I will correct them. We are still waiting for the fix for the carousel component...

PR #97, opened 2018-03-02
Adding the Connection Diagnoser Tool to Support Center

puneetg1983:
Here is a quick screenshot

image


stevenernst131:
Do you have access to the name of these configurations? It might make sense to display the name on the expandable box next to the green or red. 

e.g. 'GeoRegionDatabaseConnectionString'

puneetg1983:
@stevenernst131 , not sure if I understood your comment. These are connection strings that will be added by end customer in the Azure Portal so geo-region name isn't applicable

stevenernst131:
Sorry, bad explanation. These names, assuming they are a connection string of some kind:

image

puneetg1983:
This is applicable to connection string only (not app settings)

I think you mean the 'kind' or 'type' of connection string? Thats coming when you expand it. I guess you suggesting we show it in the default heading, yea I think that sounds good.

stevenernst131:
Even app settings are a name value pair correct? I am proposing putting the name there, because they would probably recognize the name

src/app/shared/components/tools/connection-diagnoser-tool/connection-diagnoser-tool.component.html
```diff
@@ -0,0 +1,97 @@
...
+                        <b>{{ connection.Name }}</b>
+                        <span class="pull-right">
+                            <i class="fa fa-plus" *ngIf="!connection.Expanded" aria-hidden="true"></i>
+                            <i class="fa fa-minus" *ngIf="connection.Expanded" aria-hidden="true"></i>
+                        </span>
+                    </div>
+                    <div class="panel-body" *ngIf="connection.Expanded">
+                        <table class="table table-bordered">
+                            <tbody>
+                                <tr>
```
hforeste: I think going along with what @stevenernst131 , suggested. I see that you have a property called 'Name' in the DatabaseTestConnectionResult model class. Does that property point to the name the customer uses to name their connection string?

src/app/shared/models/daas.ts
```diff
@@ -73,4 +67,31 @@ export interface DiagnoserDefinition {
     Name: string;
     Warnings: string[];
     Description: string;
+}
+
+export interface DatabaseTestConnectionResult {
+    Name: string;
```
hforeste: I think going along with what @stevenernst131 , suggested. Is this property the name that the customers uses to name their connection string?

puneetg1983:
The name is what we are putting there already...(I think the names that I chose for my site are confusing you  :) ) 

**This is Azure Portal's connection strings for my site**
image



This is what the UI of this tool will show.

image


The first one ServiceProfilerStorage is not showing up as it is a storage connection string and DAAS doesnt handle those yet (it will handle them next week onwards when I will push the latest bits of DAAS)

The additional ones that you are seeing in the UI of the tool are connection strings defined in my web app's web.config

AppSettings would be just KEY VALUE pairs and we cannot validate anything so this is just for connection strings...

So are you saying that we add the TYPE next to the name in the title ? If yes, thats something that can be done.

src/app/shared/models/daas.ts
puneetg1983: Yes, Hawk

puneetg1983: Please also check my detailed reply below..

hforeste:
Ah I see , I didn't notice that the **title of the box is the connection string name**. At first glance, i didn't recognize it to be the connection string names. 

Maybe that's because I'm not the one who created the connection string so it would make sense that I didn't recognize the name. 

I think this is fine then. I think for the purpose of this PullRequest, it would have been helpful to initially have that azure portal screenshot you took of your web apps connection string. That way as reviewers can correlate the name with the titles of the boxes.

image

puneetg1983:
Yup I realized now that it would have helped...my bad

stevenernst131:
Thanks for clearing that up @puneetg1983, looks like we were on the same page the whole time. 

One last comment is that I think there are too many colors/text formats going on in that table. I would make all the descriptions in the left column the same, say black and bolded. And get rid of the orange and the blue text and make it black. I think the red is good for the status. 

You generally don't want to mix up too many colors and formats like that.

puneetg1983:
Removed too many colors.

Thanks @stevenernst131 and @hforeste

Jen7714:
@puneetg1983 what do you think about changing the name to **Validate Connection Strings**?

PR #100, opened 2018-03-08
Adding Incident Notification

stevenernst131:
Adding ICM incident notifications

Orange banner will pop up when we detect an incident:
image

Then will open new page with details:
image


ShekharGupta1988:
This looks really nice.
Just small feedback - in the Service Incidents page, instead of 03/04, should we say "Mar 4". Few countries use MM/DD and few countries use DD/MM.

PR #102, opened 2018-03-13
Generic Detector Project Code

ShekharGupta1988:
Changes:
- **ModelsAndUtils Project** : All the models and utilities available to scripts
- **CompilerHost** : On-Demand Process which handles compilation of scripts. This is powered by roslyn. This host will be stopped or recycled when it violates memory threshold.
- **RuntimeHost** : Always-On Process which handles main API Calls and Execution of customer scripts/code.
       - **Source Watcher** : Sync source scripts and dlls from github(remote) to local directory.


src/Diagnostics.RuntimeHost/Services/CompilerHostClient.cs
```diff
@@ -0,0 +1,205 @@
...
+                    Arguments = $@"Diagnostics.CompilerHost.dll --urls {_compilerHostUrl}",
+                    UseShellExecute = false,
+                    RedirectStandardOutput = true,
+                    CreateNoWindow = false
+                }
+            };
+
+            proc.Start();
+            // TODO : Remove artificial wait.
+            await Task.Delay(5000);
```
stevenernst131: Are you supposed to remove this?

src/Diagnostics.RuntimeHost/appsettings.json
```diff
@@ -0,0 +1,42 @@
...
+      "LogLevel": {
+        "Default": "Warning"
+      }
+    }
+  },
+  "CompilerHost": {
+    "CompilerHostBinaryLocation": "",
+    "CompilerHostPort": "7000",
+    "PollingIntervalInSeconds": "60",
+    "ProcessMemoryThresholdInMB": "300"
```
stevenernst131: It looks like you have these configurations but in the code you are still using hard coded constants.

src/Diagnostics.RuntimeHost/Services/CompilerHostClient.cs
ShekharGupta1988: Cant remove this. (I'll remove the comment). Cannot wait here for process to exit. Hence artificial wait for process to be ready. We can also create a heartbeat ping on the compiler host and wait for that instead of 5 seconds artificial wait

src/Diagnostics.RuntimeHost/appsettings.json
ShekharGupta1988: Nope. The configurations gets picked up via registry for PROD env. If those settings are not present, I have default values in code.

PR #103, opened 2018-03-14
Making Layout changes and Refactoring Profiler out of DAAS to avoid code duplication

puneetg1983:


src/app/shared/components/daas/daas-validator.component.html
```diff
@@ -0,0 +1,34 @@
...
+    <div>
+      <strong>Error</strong> - This tool is supported only on Dedicated SKU's
+    </div>
+  </div>
+</div>
+
+<div *ngIf="!daasRunnerJobRunning" class="focus-box focus-box-warning" style="margin-top:20px">
+  <div>
+    <strong>Error</strong> - We found that the DAAS Web job is in a stopped state. Please go to WebJobs under this Web App and start
+    the DAAS Web job manually and re-run this tool.
```
ShekharGupta1988: Should we give a button to start the web job? Not very clear from this string that where should customer go to start the web job

puneetg1983: The problem is that the API to start the webjob is broken in support webapi, else we would have just called that API. I am not sure even if we give a button, what will the button really do or call because you cannot start webjobs through ARM API. Because if something exists, we can just call it....

src/app/availability/availability.routeconfig.ts
```diff
@@ -307,15 +307,15 @@ export const AvailabilityAndPerformanceCategoryRouteConfig: Route[] = [
         path: _siteResourceUrl + '/diagnostics/tools/databasetester',
         component: ConnectionDiagnoserToolComponent,
         data: {
-            navigationTitle: 'Connection Diagnoser',
+            navigationTitle: 'Connection Strings',
```
hforeste: I guess this is the name that was decided?

stevenernst131: @hforeste ignore this.

@puneetg1983 you need to pull my menu change, I moved names to a constants file to share between menu and route

puneetg1983: Yes this is in the Heading, the actual name of the tile is Check Connection Strings

puneetg1983: done @stevenernst131

PR #116, opened 2018-03-30
Refreshsessions

stevenernst131:
Hi @puneetg1983,

I already had to deploy this change because when I tried to do a production build I got this error:

ERROR in src\app\shared\components\daas-sessions-detailed\daas-sessions-detailed.component.html(7,109): : Property 'refreshSessions' does not exist on type 'DaasSessionsDetailedComponent'.

Production builds will actually catch any references in the HTML that do not exist in the component, whereas development builds will miss it. 

Please update this fix if it is the wrong fix.


puneetg1983:
The refreshSessions fix looks good to me. That is some left over code that should have been removed.

PR #126, opened 2018-05-08
app insights dependencies fixes

ShekharGupta1988:


stevenernst131:
Screenshot?

stevenernst131:
I thought we had a larger window now in ticket opening flow? An alternative option to doing it this way is using a media query in the css and just have the columns not render under a certain width. 

```(css)
@media screen and (max-width: 800px) {
    .date-holder {
        display: none;
    }
}
```

I have used both methods but it seems like the media queries is a more scalable solution since we don't know how big the window will open when for sure.

PR #132, opened 2018-06-22
Live Agent Chat in App Service Diagnostics

ShekharGupta1988:
Right now, this option is buried deep down in the flow on purpose. We dont want to expose this to many customers. 
Also the live agent chat component is extensible to be embeded in any Page (and not just home page chat bot).


src/app/shared/components/liveagent-chat/liveagent-chat.component.ts
```diff
@@ -0,0 +1,119 @@
...
+})
+export class LiveAgentChatComponent implements OnInit {
+
+    @Input() autoOpen: boolean = false;
+    @Input() source: string = '';
+    private restoreIdTagName: string;
+
+    // As we dont have any Resource Service yet, Right now using Site as resource.
+    // After refactoring, this has to come from a generic Resource service.
+    private currentResource: Site;
```
stevenernst131: Is there any way that we can just pass in the required parts of a resource? For example have the inputs be a resource id and a name. I see you also use tags, i am not sure what is happening there but there might be another way to do it. Basically can this component be generic, and then create a container component around it that handles the site specific logic? That way we don't even have to touch this one in the refactoring.

stevenernst131: I am understanding now a little more about the patching logic to save the id. If you took above approach, this component could have an output function and the containing component would have to handle that correctly for site.

src/app/supportbot/message-flow/talk-to-agent/talktoagentmessageflow.ts
```diff
@@ -0,0 +1,81 @@
...
+
+    public isApplicable: boolean;
+
+    constructor(private siteService: SiteService, private authService: AuthService) {
+        this.isApplicable = false;
+
+        if (this.authService.resourceType === ResourceType.Site) {
+            this.siteService.currentSite.subscribe((site: Site) => {
+
+                this.isApplicable = !(site.sku === 'free' || site.sku === 'shared');
```
stevenernst131: toLower() necessary here?

src/app/shared/components/liveagent-chat/liveagent-chat.component.ts
ShekharGupta1988: I thought of that approach but then there could be lot of inputs to the component. 
Right now, it would need the id, name, tags, subscription. 

Maybe a better approach would be to have ReourceBase as input...

ShekharGupta1988: Tags is common for all resources , so I am guessing it will work for all reosurces

src/app/supportbot/message-flow/talk-to-agent/talktoagentmessageflow.ts
ShekharGupta1988: fixed

PR #142, opened 2018-07-25
Autohealing UI in Diagnostics Portal

puneetg1983:
Including a few screenshots

image

If you click on Custom Action, then you see this.

image

Once you have configured multiple rules, you can see the tiles select and corresponding action selected. The Review shows you all the rules configured.

image

You can modify the startup time and it gets updated in the settings section.

image

At any point, you can edit or delete the rules and the tiles should update accordingly.

Let me know if you guys want me to do a quick demo before going on reviewing the pull request.


src/app/shared/components/autohealing/autohealing-config-summary/autohealing-config-summary.component.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  updateComponent() {
+
+    if (this.autohealSettings == null) {
+      return;
+    }
+
+    this.actualhealSettings = JSON.parse(this.autohealSettings);
+
+    console.log("updateComponent called");
```
stevenernst131: remove all console.log()

src/app/shared/services/formatting.service.ts
```diff
@@ -0,0 +1,36 @@
+import { Injectable } from "../../../../node_modules/@angular/core";
+
+@Injectable()
+export class FormattingService {
```
stevenernst131: This could just be a static class, instead of a service. That way you don't have to inject it everywhere you want to use it.

puneetg1983: done. I was looking for the right way to do myself....

PR #144, opened 2018-07-30
Checking empty error response & changing logic of cache service slightly

puneetg1983:
Hey Fellas,
Can you please closely review this PR to see if there are any breaking changes due to this change ? 

1) For the handleError() I am proposing we check for empty response body and if found one, return the Status and StatusText

2) For the cache service, I found an issue that if make a call to an ARM API which is failing (i.e. not returning HTTP Status OK), **the first call successfully gets the error back but the second call to the same API doesn't return anything**. I noticed this even in DaaS components. The subscriber to the second call never gets fired. The change below seems to have addressed that but I am not too familiar with this codepath so checking if there is a better \ alternate way of doing this.


src/app/shared/services/cache.service.ts
```diff
@@ -32,9 +32,11 @@ export class CacheService {
         if (this.inFlightObservables.has(key)) {
             return this.inFlightObservables.get(key);
         } else if (fallback && fallback instanceof Observable) {
-            this.inFlightObservables.set(key, new Subject());
             this.log(`%c Calling api for ${key}`, 'color: purple');
-            return fallback.do((value) => { this.set(key, value); });
+            return fallback.do((value) => {
+                this.inFlightObservables.set(key, new Subject());
```
stevenernst131: I think the above change could be a race condition, two could enter fallback.do before the inflight observables is set.

puneetg1983: Hey @stevenernst131 , I am not aware of everything in this block of code. I wanted to just raise this problem and hope this gets fixed soon. If you have a better way of fixing this, feel free to send a PR and I will close this one but I would really like to see this one getting fixed soon.

stevenernst131: Let me try to repro this for myself today. I took the cache code from someone else :)

ShekharGupta1988: @stevenernst131  - this is the same bug I fixed in Applens. I pasted a snapshot above too.. See if that helps!!
https://github.com/stevenernst131/Applensv3/pull/13/files

stevenernst131: Nice! @puneetg1983 If you want to copy shekhars change, that is good with me.

puneetg1983:
Thanks @stevenernst131  and @ShekharGupta1988 . I validated that this change also fixes the issue so I updated the PR.

PR #145, opened 2018-07-30
Autohealing UI Changes

puneetg1983:
In summary of this pull request..

1. Added a top menu to check AutoHealing history by calling the autoheal detector.
1. Extended AutoHealingComponent from DetectorViewBaseComponent to be able to call AutoHealing detector
1. Move the startup time to a new tile to avoid breaking any flow
1. Added a check for Always On before configuring diagnostics.
1. Changed icon for Request Duration back to hourglass to avoid repetition of clock icon
1.  Added a Save Button to the Custom Action section (for both the Diagnostics Component and the Custom Component)
1. Changed the autoheal ON\OFF to an ibiza style component.
1.  Introduced a lot of null checks that I ran into with sites that didnt have autohealing enabled.


stevenernst131:
Hey @puneetg1983, if this is not too much to ask, I think it would be beneficial for this work to be its own module, instead of being part of the shared module. I think it will keep the organization a little clearer, as well as when the code is packaged, it can create smaller bundles, instead of having a really big shared bundle. It will also make it easier as I try to refactor some of this project to make it more generic.

puneetg1983:
Hey @stevenernst131 , I created a new Module and moved all the autohealing components to that module. I left two components (timespan and toggle) in shared module as other modules can also leverage those.

Kindly review and help me close this PR soon :)

PR #147, opened 2018-08-03
Implementing Delete session in DaaS UI

puneetg1983:
This will allow users to delete Daas Session and reclaim disk space.

### DIFF
--- a/src/app/shared/components/daas-sessions/daas-sessions.component.ts
+++ b/src/app/shared/components/daas-sessions/daas-sessions.component.ts
@@ -66,8 +66,8 @@ export class DaasSessionsComponent implements OnChanges, OnDestroy {
         }
     }
 
-    reducedSession(obj:Session) {
-        return {SessionId: obj.SessionId, DiagnoserSessions: obj.DiagnoserSessions, Status: obj.Status};
+    reducedSession(obj: Session) {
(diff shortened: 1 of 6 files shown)


src/app/shared/components/daas-sessions/daas-sessions.component.ts
```diff
@@ -165,6 +164,23 @@ export class DaasSessionsComponent implements OnChanges, OnDestroy {
...
+            sessionToDelete.Expanded = true;
+            sessionToDelete.DeletingFailure = "";
+            sessionToDelete.Deleting = true;
+            this._daasService.deleteDaasSession(this.siteToBeDiagnosed, sessionToDelete.SessionId)
+                .subscribe(resp => {
+                    sessionToDelete.Deleting = false;
+                    this.sessions.splice(sessionIndex, 1);
+                },err => {
+                    sessionToDelete.Deleting = false;
+                    sessionToDelete.DeletingFailure = "Failed while deleting the session with an error : " + err;
```
stevenernst131: Are you sure error is a string here?

puneetg1983: I tried various variations of error 401, 500, json based and it is coming properly as I changed handleError method slightly. Since the handleError checks for ***error.json().error || error.json().Message*** and if it doesn't find it, it returns 'Server Error', I think the error will always be string but correct me if I am wrong.

PR #163, opened 2018-08-30
Add Public Holidays logic

ShekharGupta1988:
I am looking to move this config out of code so that it can be updated. Few things needs to be ironed out here.

### DIFF
--- a/src/app/liveChatSettings.ts
+++ b/src/app/liveChatSettings.ts
@@ -21,7 +21,8 @@ export class LiveChatSettings {
         "32581615",     //Configuration and Management/Deployment Slots
         "32542208",     //Configuration and Management/Backup and Restore
         "32589276",     //How Do I/Backup and Restore
-        "32589277"      //How Do I/Configure domains and certificates
+        "32589277",     //How Do I/Configure domains and certificates,
+        "32589281"      //How Do I/IP Configuration
     ];
 
     // This indicates the time after which Live Chat will pop up. 
--- a/src/app/liveChatSettings.ts
+++ b/src/app/liveChatSettings.ts
@@ -32,4 +32,37 @@ export class LiveChatSettings {
     public static BuisnessEndDay: Number = 5;   // Friday
     public static BusinessStartHourPST: Number = 7;   // 7 AM PST
     public static BusinessEndHourPST: Number = 17;     // 5 PM PST
+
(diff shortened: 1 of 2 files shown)


src/app/liveChatSettings.ts
```diff
@@ -32,4 +32,37 @@ export class LiveChatSettings {
     public static BuisnessEndDay: Number = 5;   // Friday
     public static BusinessStartHourPST: Number = 7;   // 7 AM PST
     public static BusinessEndHourPST: Number = 17;     // 5 PM PST
+
+    public static PublicHolidays: any = [
```
stevenernst131: Might be worth looking into this: https://www.npmjs.com/package/moment-holiday

stevenernst131: there is a moment.isHoliday() function it looks like

ShekharGupta1988: oh nice.. We still need to pass a list of holidays. I'll explore that. looks to be a better way

src/app/shared/services/livechat.service.ts
```diff
@@ -206,8 +206,23 @@ export class LiveChatService {
...
         return isApplicable;
     }
+
+    private isPublicHoliday(currentDate): boolean {
+
+        for (var iter = 0; iter < LiveChatSettings.PublicHolidays.length; iter++) {
+
+            var element = LiveChatSettings.PublicHolidays[iter];
+
+            if (element.date == currentDate.date() && ((element.month - 1) == currentDate.month()) && element.year == currentDate.year()) {
```
stevenernst131: There might be some moment utilities to do some of these comparisons

src/app/liveChatSettings.ts
ShekharGupta1988: There is little more exploration needed here. For eg- 'Good Friday' is not a holiday in Microsoft but this moment-holiday returns true. So chat will not show on that day.

I'll explore it little bit more and see how to best use this. If I just use it default, there is an risk of not showing chat on few days.

stevenernst131: Makes sense, did you get the ones you added from the official Microsoft holiday list?

ShekharGupta1988: Right now, I only added for remainder of this year.

PR #169, opened 2018-09-25
Enable Live chat from 9-6

ShekharGupta1988:
Only for Case Submission right now from 9AM to 6PM

### DIFF
--- a/src/app/liveChatSettings.ts
+++ b/src/app/liveChatSettings.ts
@@ -3,11 +3,11 @@ export class LiveChatSettings {
 
     // Global Switch that controls whether chat is shown or not. 
     // If False, it will override, every other setting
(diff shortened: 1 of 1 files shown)


stevenernst131:
Just internal?

ShekharGupta1988:
Internal Just for Home Page . 
I found that chat is not there on Home Page?

stevenernst131:
Chat should only show up for case submission, I haven't integrated it into chat yet. What do you mean for 'Home Page'

ShekharGupta1988:
Diagnose and Solve blade

stevenernst131:
Way ahead of you :)
image
Now that flag just has to actually be used

ShekharGupta1988:
lol.. That's cool.

Then whenever you have time, just add the live chat back on the category. :)

PR #172, opened 2018-09-26
Fix Case Submission For Availability and Perf

ShekharGupta1988:


src/app/home/components/support-topic-redirect/support-topic-redirect.component.ts
```diff
@@ -1,6 +1,6 @@
 import { Component, OnInit } from '@angular/core';
 import { ActivatedRoute, Router } from '@angular/router';
-import { SupportTopicService } from '../../../shared-v2/services/support-topic.service';
+import { SiteSupportTopicService } from '../../../resources/web-sites/services/site-support-topic.service';
```
stevenernst131: What wasn't working here? There shouldn't be any resource specific services in the home module

PR #179, opened 2018-10-31
Outage Comms in Case Submission

ShekharGupta1988:
image


stevenernst131:
Nit: Just a suggestion I think we were talking and I mentioned it might make more sense to name the folder something other than 'AngularApp' just because it will be confusing if you have two vscode windows open, to know where you are.

ShekharGupta1988:
Sure.. will rename the folder .
BTW, I had to deploy this to PROD as in last couple of days, there are 2 comms - one with SN1 and other in West US and we suspect cases got leaked through

PR #186, opened 2018-11-16
Hide Legend when there are numerous sites in one server farm

stevenernst131:
This is what the customers were seeing:

image

I increased the height but also hid the legend when there are too many sites so that this doesn't happen.

### DIFF
--- a/AngularApp/src/app/shared/components/instance-view-graph/instance-view-graph.component.html
+++ b/AngularApp/src/app/shared/components/instance-view-graph/instance-view-graph.component.html
@@ -17,7 +17,7 @@
             <span class="tool-tip-text">{{description}}</span>
         </div>
     </div>
-    <div class="row" style="padding-right: 10px; height: 200px">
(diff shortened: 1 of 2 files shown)


cindylovescoding:
Looks good!

AngularApp/src/app/shared/components/instance-view-graph/instance-view-graph.component.html
```diff
@@ -17,7 +17,7 @@
             <span class="tool-tip-text">{{description}}</span>
         </div>
     </div>
-    <div class="row" style="padding-right: 10px; height: 200px">
+    <div class="row" style="padding-right: 10px; height: 300px">
```
hforeste: Steve what do you think about setting the height to `auto` and dynamically decide the chart height in the component?

stevenernst131: I can try that, not sure exactly what the behavior for auto will be. CSS can do surprising things.

PR #193, opened 2018-12-11
Enabling all the DAAS tools on Function Apps running on Dedicated SKU

puneetg1983:
All the DAAS tools should run on Function Apps that are not on Consumption plan so with this PR, I am enabling those...


stevenernst131:
I included this in my other Pull Request so I am going to close this one.

PR #195, opened 2018-12-14
Set up CI with Azure Pipelines

stevenernst131:


stevenernst131:
Adding PR build checking

PR #200, opened 2019-01-09
fixing dropdowns

puneetg1983:


AngularApp/projects/diagnostic-data/src/lib/components/dropdown/dropdown.component.ts
```diff
@@ -37,7 +37,7 @@ export class DropdownComponent extends DataRenderBaseComponent {
       const row = table.rows[i];
       this.label = row[labelColumn];
       const key: string = row[keyColumn];
-      const selected: boolean = row[selectedColumn].toLowerCase() === 'true';
+      const selected: boolean = String(row[selectedColumn]).toLowerCase() === 'true';
```
stevenernst131: isn't row[selectedColumn] already a boolean? may just be able to get rid of the translating it to string and doing the comparison?

puneetg1983: It is currently a boolean but ideally the backend should send a string and I was thinking that will get fixed at some point ? So if it changes in future, we dont have to change this part....

stevenernst131: No the back end will not change, I don't think.

ShekharGupta1988: If the changes are deployed everywhere (meaning all georegions), then just using the boolean value const selected: boolean = row[selectedCloumn] should be fine..

However, the current change looks more guarded against any future changes that might happen.

stevenernst131:
I checked in fix for this

PR #205, opened 2019-01-11
Make dropdown work correctly in app service diagnsotics

stevenernst131:


stevenernst131:
This is needed because the dropdown component is using functionality that is given by bootstrap, and bootstrap is dependent on jquery.

PR #206, opened 2019-01-12
Better error messaging for incorrect time range input at AppLens

michimune:
This is a PR for the work item 1393: Better error messaging for incorrect time range input.

This is only for AppLens, not for App Service Diagnostics.

### DIFF
--- a/AngularApp/projects/diagnostic-data/src/lib/components/detector-view/detector-view.component.html
+++ b/AngularApp/projects/diagnostic-data/src/lib/components/detector-view/detector-view.component.html
@@ -45,5 +45,5 @@
 <loader-view style="margin-top:300px" *ngIf="!detectorDataLocalCopy && !errorState" message="Loading Detector..."></loader-view>
 
 <ng-container *ngIf="!detectorDataLocalCopy && errorState">
-  Error loading detector: <span class="critical-color" *ngIf="errorState.error">{{errorState.error}}</span>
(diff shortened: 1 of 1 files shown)


stevenernst131:
This code is shared between both applens and app service diagnostics. I think we should come up with a solution that works for both. Right now this will also show up for App Service Diagnostics. 

Praveen had also asked me to remove the word 'detector' for external, since we don't call them that to customers anywhere else.

AngularApp/projects/diagnostic-data/src/lib/components/detector-view/detector-view.component.html
```diff
@@ -45,5 +45,5 @@ <h1 *ngIf="!isSystemInvoker">
 <loader-view style="margin-top:300px" *ngIf="!detectorDataLocalCopy && !errorState" message="Loading Detector..."></loader-view>
 
 <ng-container *ngIf="!detectorDataLocalCopy && errorState">
-  Error loading detector
+  Error loading detector: <span class="critical-color">{{errorState.error}}</span>
```
stevenernst131: You may want to do something like this, just to make sure the error property exists:

```(html)
<span class="critical-color" *ngIf="errorState.error">{{errorState.error}}</span>
```

michimune: Thanks! So do we also want to change the message "Error loading detector" to remove the word 'detector'?

michimune:
Thank you!

PR #209, opened 2019-01-18
Enable Chat for Memory + Update holidays

ShekharGupta1988:


stevenernst131:
What app setting change? We will need to add it to app setting of the app.

ShekharGupta1988:
Updated the public holidays. I noticed you moved them to app settings from the angular app

PR #210, opened 2019-01-22
Enable .net Profiling for ASP.NET Core

puneetg1983:
Following are the changes in this PR

1. Enabling profiler for ASP.NET Core and getting rid of the ASP.NET Core button from the stack as both ASP.NET and ASP.NET core show the same set of tools 
image
 
2. When we detect ASP.NET Core stack less than 2.2.3 version, the below message is shown
 
image

3. Also adding a mode for end customers if they need only data to be collected and not anlaysis (on CSS request)

image

4. Changing the output section of diagnostic tools to show both logs and reports instead of just reports (on CSS requests) and enabling the cancel button

 image

5. Displaying Session Size for each DaaS session 

image

6. Adding all the DAAS API for the CPU monitoring feature to DaaS Service. Will add the UI for that later.


PraveenTB:
The changes look pretty cool - very nice improvements. In the .net core requirements, is there any version of .net core that doesn't require the change to program.cs? If so may be we can give that as a second option to update the version

puneetg1983:
@stevenernst131  and @PraveenTB -  Based on your feedback, I brought back the ASP.NET Core button so now no changes if someone switches between ASP.NET or ASP.NET core but the right set of tools are displayed.

@PraveenTB - **All versions of .net core > 2.2.3 do not require changes to Program.cs** and in those cases, we are not showing this warning.

@stevenernst131 - I had to modify **site-filter.pipe.ts** to ensure that duplicate tiles dont get added. Please review this code if you get a chance and let me know if there is any better to deal with this. Without this change, we were getting the .net profiler tile twice if someone chooses ASP.NET or All

stevenernst131:
@puneetg1983 I think the correct method is to allow for the items to have an array of stacks that are enabled for that tool or resource. Then you just see if any of those stacks provided match the site stack or the override stack.

PR #211, opened 2019-01-22
Adding Cards Rendering Type

stevenernst131:
The purpose of this rendering type is to be used for certain problem areas where we want to drill down further, and give the user a better idea of the type of errors that would make them want to check on each card. The primary purpose of this detector is to use it in case submission only, however if it is a public detector, there is nothing stopping it from showing up in the search results. That may be a TODO item to tackle.

image

If you notice in the screenshot, I have hidden the time picker and the heading for any detector that has a cards rendering. I think this solution is a bit hacky, and we should probably add an ability for a detector to hide the heading, or other options related to how the detector is displayed.


AngularApp/projects/app-service-diagnostics/src/app/app.module.ts
```diff
@@ -46,7 +47,9 @@ import { PortalKustoTelemetryService } from './shared/services/portal-kusto-tele
     CustomReuseStrategy,
     { provide: KustoTelemetryService, useExisting: PortalKustoTelemetryService },
     { provide: RouteReuseStrategy, useExisting: CustomReuseStrategy },
-    { provide: DiagnosticService, useExisting: GenericApiService },
+    { provide: DiagnosticService, 
+      useFactory: (_localBackendService: LocalBackendService, _genericApiService: GenericApiService) => environment.useApplensBackend ? _localBackendService : _genericApiService,
```
cindylovescoding: Are we able to run against local Applens backend in the local Azure portal environment? The extension portal environment is still counted as production env right? So the DiagnosticService of that env will always use the _genericApiService?

AngularApp/projects/app-service-diagnostics/src/app/shared/services/local-backend.service.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  resourceId: string;
+
+  detectorList: DetectorMetaData[];
+
+  useLocal: boolean = false;
+
+  constructor(private _http: HttpClient, private _armService: ArmService, private _authService: AuthService) {
+    this._authService.getStartupInfo().subscribe(info => {
+      this.resourceId = info.resourceId.replace('microsoft.web', 'Microsoft.Web');
```
cindylovescoding: Is this the only place that's need to be replaced?

AngularApp/projects/app-service-diagnostics/src/app/shared/services/local-backend.service.ts
```diff
@@ -0,0 +1,77 @@
...
+    let headers = new HttpHeaders({
+      'Content-Type': 'application/json',
+      'Accept': 'application/json'
+    });
+
+    if (path) {
+      headers = headers.append('x-ms-path-query', path);
+    }
+
+    if (method) {
```
cindylovescoding: this is always true right?

AngularApp/projects/app-service-diagnostics/src/styles/controls/_cards.scss
```diff
@@ -0,0 +1,17 @@
...
+    min-height: 64px;
+    box-shadow: $card-box-shadow;
+    margin-left: 0px;
+    margin-right: 15px;
+    margin-bottom: 10px;
+    border-radius: $card-border-radius
+}
+
+.card {
+    @extend %button
```
cindylovescoding: Does scss allow extend from a different file without import?

AngularApp/projects/app-service-diagnostics/src/styles/controls/_cards.scss
```diff
@@ -0,0 +1,17 @@
...
+    font-size: 14px;
+    min-height: 64px;
+    box-shadow: $card-box-shadow;
+    margin-left: 0px;
+    margin-right: 15px;
+    margin-bottom: 10px;
+    border-radius: $card-border-radius
+}
+
+.card {
```
cindylovescoding: Didn't find this class being used.. Do we have related style defined in Applens project?

AngularApp/projects/diagnostic-data/src/lib/components/card-selection/card-selection.component.ts
```diff
@@ -0,0 +1,56 @@
...
+        linkType: parseInt(row[3]),
+        linkValue: row[4]
+      });
+    })
+  }
+
+  public selectCard(card: CardSelection) {
+    if (card && card.linkType === CardActionType.Detector) {
+      this._navigator.NavigateToDetector(this._activatedRoute.snapshot.params['detector'], card.linkValue);
+    }
```
cindylovescoding: What will happen if CardActionType equals to Tool?

AngularApp/projects/diagnostic-data/src/lib/services/diagnostic.service.ts
```diff
@@ -10,6 +10,10 @@ export class DiagnosticService {
     // In applens we provide this withValue: applens-diagnostics.service
     // In Support Center we provide this withValue: generic-api.service
 
+    public getDetectorById(detectorId: string) {
```
cindylovescoding: Why do we always return null?

AngularApp/projects/app-service-diagnostics/src/app/app.module.ts
stevenernst131: What do you mean by "extension portal environment"? If you are running the App Service Diagnostics Portal project locally, you can now point that to hit your applens backend. My purpose for doing this was testing out a new rendering type in the portal by using this link https://ms.portal.azure.com/?websitesextension_ext=appsvc.env%3Dlocal

GenericApiService will always be used in production.

AngularApp/projects/app-service-diagnostics/src/app/shared/services/local-backend.service.ts
stevenernst131: Yes this is an issue related to case sensitivity in our routes. May be a better way to fix it but for now this works.

AngularApp/projects/app-service-diagnostics/src/app/shared/services/local-backend.service.ts
stevenernst131: Right now it is, yes. But method is not a required parameter to the method.

AngularApp/projects/app-service-diagnostics/src/styles/controls/_cards.scss
stevenernst131: I think this may not be used right now.

AngularApp/projects/diagnostic-data/src/lib/components/card-selection/card-selection.component.ts
stevenernst131: Right now nothing.. still need to figure that out. Right now only detectors are supported.

AngularApp/projects/diagnostic-data/src/lib/services/diagnostic.service.ts
stevenernst131: This is basically a base class. There may be another way to do this, but for now it is overridden in applens and app service diagnostics, so this base class is never used.

stevenernst131: I am basically using it as a generic injection token.

AngularApp/projects/app-service-diagnostics/src/styles/controls/_cards.scss
stevenernst131: Removed this file

PR #213, opened 2019-01-24
Fixing broken links in profiler and updating session state properly

puneetg1983:


AngularApp/projects/app-service-diagnostics/src/app/shared/components/daas-sessions/daas-sessions.component.ts
```diff
@@ -77,8 +76,8 @@ export class DaasSessionsComponent implements OnChanges, OnDestroy {
             .subscribe(sessions => {
                 const newSessions = sessions.map(this.reducedSession);
                 const existingSessions = this.sessions.map(this.reducedSession);
-
-                if (newSessions.length === existingSessions.length && newSessions.filter(newSession => existingSessions.findIndex(session => session === newSession) === -1).length > 0) {
+                let anySessionUpdated = newSessions.filter(newSession => existingSessions.findIndex(session => JSON.stringify(session) === JSON.stringify(newSession)) === -1).length > 0;
```
stevenernst131: Is the only way to compare a session to do a JSON.stringify? If it works it works but seems strange, no id or anything?

puneetg1983: Let me try to explain. The Session has an Id but it has a property called DiagnoserSessions which gets updated as an active session is progressing. So this DiagnoserSessions will have current information about logs collected and reports generated and this changes as a session progresses for the same sessionId, hence the Json.Stringify. 

Would love to know If there is a better way to handle this without writing a custom function to compare object's properties.

stevenernst131: You could just always update the sessions list, instead of comparing and seeing if anything changed, change happens so fast that it wouldn't even be detectable by the user. Other than that I think this solution is fine if you want to keep it.

puneetg1983: Unfortunately changes are not that fast always... At times we can take from a few minutes to even 30 to 40 minutes... That's why updating the session status becomes important... I can explain more if required. I'll go ahead and merge this now.

PR #216, opened 2019-01-30
Ran Style Tools On Various Files

dagrooms52:
- Ran style tools while surveying the codebase
- Tried out `namespace` feature on a group of interfaces (DetectorResponse)
- No big deal, can remove any of these if they're a problem


AngularApp/projects/app-service-diagnostics/src/app/shared/models/detectorresponse.ts
```diff
@@ -1,53 +1,55 @@
...
-    startTime: string;
-    endTime: string;
-    issueDetected: boolean;
-    detectorDefinition: IDetectorDefinition;
-    metrics: IMetricSet[];
-    abnormalTimePeriods: IDetectorAbnormalTimePeriod[];
-    data: INameValuePair[][];
-    responseMetaData: IDetectorMetaData;
-}
+namespace DetectorResponse {
```
stevenernst131: Why is it trying to add this namespace? Not sure what the ramifications are

AngularApp/projects/app-service-diagnostics/src/app/supportbot/message-flow/detector-summary/detector-summary.component.ts
```diff
@@ -140,9 +142,10 @@ export class DetectorSummaryComponent implements OnInit, AfterViewInit, IChatMes
 
     for (let i: number = 0; i < data.rows.length; i++) {
       const row = data.rows[i];
-      let insight: DetectorSummaryViewModel;
+      let insight: DetectorSummaryViewModel = insights.find(insight => insight.name === insightName);
```
stevenernst131: insight design is not defined until the line after so I would think this would fail

dagrooms52: Right. Weird that this doesn't show up in static analysis or compilation

AngularApp/projects/app-service-diagnostics/src/app/shared/models/detectorresponse.ts
dagrooms52: I found namespace as a suggestion to reduce importing a ton of interfaces, but it looks like I need to take a better look at it later

PR #218, opened 2019-02-01
Case Cleansing Improvements

craigmaher:
Updates to the case Cleansing tool:
* Full width table
* Filtering options
* Improoved UI
* Bug Fixes


stevenernst131:
@craigmaher Can you add a screenshot so we can get an idea as to what the changes are?

craigmaher:
annotation 2019-02-01 133628
annotation 2019-02-01 133925

craigmaher:
@ShekharGupta1988 yes I can, that will just be a change in the underlying database and not in this repository.

PR #221, opened 2019-02-06
UI change to support for removal of GuageSize.Inherit. Also fixed a b…

nmallick1:
…ug where the label wasn't being displayed if a non markdown string was passed. Changed the name of the class from label to guage-label as the import was conflicting.


stevenernst131:
Yes, styles from the component style file will only apply to the component, but be careful with really simple names, as that can cause unintended styles from the common style files to be applied. Another good example is 'container'

PR #227, opened 2019-02-13
Removing Icon and Adding color to Card Tiles

puneetg1983:
minor changes to card rendering based on feedback...

image

### DIFF
--- a/AngularApp/projects/diagnostic-data/src/lib/components/card-selection/card-selection.component.ts
+++ b/AngularApp/projects/diagnostic-data/src/lib/components/card-selection/card-selection.component.ts
@@ -55,10 +55,6 @@ export class CardSelectionComponent extends DataRenderBaseComponent {
   }
 
   public getColor(index: number): string {
-    let color = "";
-    if (index < this.colors.length) {
-      color = this.colors[index];
-    }
-    return color;
+    return this.colors[index % this.colors.length];
   }
 }
--- a/AngularApp/projects/diagnostic-data/src/lib/components/card-selection/card-selection.component.ts
+++ b/AngularApp/projects/diagnostic-data/src/lib/components/card-selection/card-selection.component.ts
@@ -14,10 +14,9 @@ export class CardSelection {
   linkValue: string;
 }
 
-export enum CardActionType
-{
-    Detector,
-    Tool
+export enum CardActionType {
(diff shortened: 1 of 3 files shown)


stevenernst131:
Did you run  a Keros accessibility check on this, specifically color contrast?

AngularApp/projects/diagnostic-data/src/lib/components/card-selection/card-selection.component.ts
```diff
@@ -53,4 +53,12 @@ export class CardSelectionComponent extends DataRenderBaseComponent {
       this._navigator.NavigateToDetector(this._activatedRoute.snapshot.params['detector'], card.linkValue);
     }
   }
+
+  public getColor(index: number): string {
+    let color = "";
+    if (index < this.colors.length) {
```
stevenernst131: shouldn't you do:

```
return this.colors[index % this.colors.length];
```

otherwise if you have more than 8 there are no colors on the additional cards

stevenernst131:
I personally think the colors are a little bit of stimulation overload, but you can go of course forward if you and Jen agree. I will say though that I think we should try to be more consistent with colors, for example use the same set of colors in this as we use for the cards in the home screen as we do here or vice versa.

ShekharGupta1988:
Agreed that we use the existing colors we have on home page. 
I somehow like this more than the original view to be honest as its more vibrant and playful as compared to grey background titles earlier. 
@puneetg1983  - its probably a good idea to take everyone's opinion. Maybe show this in tomorrow's meeting to take feedback.

puneetg1983:
Thanks guys...will show both versions in meeting tomorrow and will take everyone's feedback

puneetg1983:
All the colors I used above are exactly same as what we use in the ASD homepage so I guess that's fine.

AngularApp/projects/diagnostic-data/src/lib/components/card-selection/card-selection.component.ts
puneetg1983: I was somehow thinking that we will never have more than 8 cards on a page but this suggestion is way better :)

puneetg1983:
The common consensus on this one in our today's meeting is get the colors so merging this PR

PR #230, opened 2019-02-14
Ignore case while fetching headers

ShekharGupta1988:


ShekharGupta1988:
@stevenernst131  - can you quickly review this. I think AFD makes all the headers as lower case. Anyway the change to ignore case is more solid. 
Thinking of pushing this data to appinsights for some sort of monitoring.

PR #232, opened 2019-02-20
Update README with Diagnostics Portal local environment setup

dagrooms52:
- Added instructions for setting up a local Diagnostics Portal environment to test Diagnostics changes in the portal

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Azure App Services Diagnostics Portal And Applens
 
-This is the repository for Azure App Service diagnostics experience. 
+This is the repository for Azure App Service Diagnostics experience.
 
 ## Prerequisites
 - [Node 6.*](https://nodejs.org/en/download/)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,6 +1,6 @@
 # Azure App Services Diagnostics Portal And Applens
 
-This is the repository for Azure App Service diagnostics experience. 
+This is the repository for Azure App Service Diagnostics experience.
```
stevenernst131: Lol this was lower case on purpose, ask Jen about it. I don't care either way.

stevenernst131:
This is a really nice write up, I am fine with you checking it in, I just think we should make this process a little cleaner if possible. Removing the Authorize tags was kind of a hack, and having to hardcode the resource group is a pretty bad bug that we should fix, instead of documenting a workaround, or maybe something that belongs in OneNote, but not on the readme of the public repo.

dagrooms52:
@stevenernst131 I've gotten all but the Authorize tags worked out using the environment.ts settings. Working on figuring out that auth piece now.

I'll update the docs and finish the PR once I have those sharp edges cleared out

PR #233, opened 2019-02-21
Enable Applens To Run In Local Environment

dagrooms52:
- Added environment.local.ts to set up Angular project for local environment
- Local server setup is described in README.md


ApplensBackend/Startup.cs
```diff
@@ -89,4 +98,30 @@ public class Startup {
...
     }
+
+    // Use this to skip Auth on local server
+    public class LocalFilterProvider: IFilterProvider
+    {
+        public int Order
+        {
+            get
+            {
+                return -1500;
```
cindylovescoding: What's this used for?

dagrooms52: This bypasses Auth if the server is running locally (user has to add `"ServerMode": "internal"` to `appsettings.Development.json`)
Otherwise requests from the local ASD portal will fail Auth

cindylovescoding: I am just wondering what's this Order property for:)

dagrooms52: Determines the Order that the filter will run, so it's set arbitrarily low to ensure this filter runs first. Could be `Int.MinValue` if that makes things clearer

cindylovescoding: Got it. Thanks!

PR #237, opened 2019-02-26
Forms ui

rekhaswaminathan:
- Form components to render UI

- Compilation optimization using script-etag from Reponse header 

- Local development - Check for runtimeSucceeded before publishing
 Sample UI screenshots:

image

image

image

image

image


AngularApp/projects/app-service-diagnostics/src/app/shared/services/generic-api.service.ts
```diff
@@ -42,13 +42,16 @@ export class GenericApiService {
         }
     }
 
-    public getDetector(detectorName: string, startTime: string, endTime: string, refresh?: boolean, internalView?: boolean) {
+    public getDetector(detectorName: string, startTime: string, endTime: string, refresh?: boolean, internalView?: boolean, formQueryParams?: string) {
```
ShekharGupta1988: should we just call them additionalQueryParams as far as client is concerned?

AngularApp/projects/applens/src/app/modules/dashboard/onboarding-flow/onboarding-flow.component.ts
```diff
@@ -230,17 +233,28 @@ export class OnboardingFlowComponent implements OnInit, OnDestroy {
...
         this.runButtonDisabled = false;
         this.runButtonText = "Run";
         this.runButtonIcon = "fa fa-play";
-        this.queryResponse.compilationOutput.compilationOutput.forEach(element => {
+        this.queryResponse.compilationOutput.compilationTraces.forEach(element => {
           this.buildOutput.push(element);
         });
-
+        // If the script has been compiled at the server, store those values in memory.
+        if(this.queryResponse.compilationOutput.isCompiled) {
```
ShekharGupta1988: Once you do the backend change, this clause will have to be updated to match the original request Etag with the response Etag

AngularApp/projects/diagnostic-data/src/lib/components/detector-container/detector-container.component.ts
```diff
@@ -16,7 +16,7 @@ export class DetectorContainerComponent implements OnInit {
   error: any;
   hideDetectorControl: boolean = false;
 
-  private _detector: string;
+   _detector: string;
```
ShekharGupta1988: remove the _ at the beginning. (Typescript code guidelines)
If you see old places where we have used that _ , its most likely that we werent aware of the typescript stuff that time :D

rekhaswaminathan:
> @rekhaswaminathan - Maybe from screenshot, its not very apparent but one small improvement you can do in the UI is to align the label and textbox close to each other
 
Made changes to adjust the label and text 
image

AngularApp/projects/diagnostic-data/src/lib/components/form/form.component.ts
```diff
@@ -0,0 +1,167 @@
...
+      if(totalForms > 0) {
+       for(let i=0; i<totalForms; i++) {
+         this.detectorForms[i] = new Form();
+         this.detectorForms[i].formId = data.rows[i][0];
+         this.detectorForms[i].formTitle = data.rows[i][1];
+         let formInputs = data.rows[i][2];
+         for(let ip =0; ip<formInputs.length; ip++) {
+           if(formInputs[ip]["inputType"] === InputType.Button) {
+              this.detectorForms[i].formButtons.push(new FormButton(
+                this.detectorForms[i].formId+"."+formInputs[ip]["inputId"],
```
ShekharGupta1988: instead of using + , you can do 

`${this.detectorForms[i].formId}.${formInputs[ip]["inputId"]}`

AngularApp/projects/diagnostic-data/src/lib/components/form/form.component.ts
```diff
@@ -0,0 +1,167 @@
...
+               formInputs[ip]["inputType"],
+               formInputs[ip]["label"],
+               formInputs[ip]["isRequired"]));
+           }       
+         }
+       }
+      }
+    }
+
+    executeForm(formId: any, buttonId: any) {
```
ShekharGupta1988: should we call it OnSubmitFormAction

ApplensBackend/Controllers/DiagnosticController.cs
```diff
@@ -97,20 +98,41 @@ public async Task<IActionResult> Invoke([FromBody]JToken body)
...
+                scriptETag = Request.Headers["diag-script-etag"];
+            }
+
+            string assemblyName = "";
+
+            if (Request.Headers.ContainsKey("diag-assembly-name"))
+            {
+                assemblyName = Request.Headers["diag-assembly-name"];
+            }
+            var response = await this._diagnosticClient.Execute(method, path, body?.ToString(), internalView, new CompilationParameters
```
ShekharGupta1988: I think this approach will bite us in future. Imagine if you have add more headers in some other contexts.
A more generic way would be to add an extra parameter to Execute Method called 'additionalHeaders' of Type  Http Headers and just add those in the request.

ShekharGupta1988: That means you dont even need an extra class for Compilation Parameters

ApplensBackend/Startup.cs
```diff
@@ -78,6 +78,7 @@ public class Startup {
                 .AllowAnyHeader()
                 .AllowAnyMethod()
                 .AllowAnyOrigin()
+                .WithExposedHeaders(new string[] {"diag-script-etag"})
```
ShekharGupta1988: Is it possible to say diag-*

ApplensBackend/Controllers/DiagnosticController.cs
rekhaswaminathan: Agreed, i'll change it.

ApplensBackend/Startup.cs
rekhaswaminathan: .NET core documentation says it is for specific custom headers.
https://docs.microsoft.com/en-us/aspnet/core/security/cors?view=aspnetcore-2.2#set-the-exposed-response-headers

https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.cors.infrastructure.corspolicybuilder.withexposedheaders?view=aspnetcore-2.2

AngularApp/projects/app-service-diagnostics/src/app/shared/services/local-backend.service.ts
```diff
@@ -39,8 +39,11 @@ export class LocalBackendService {
     }));
   }
 
-  public getDetector(detectorName: string, startTime: string, endTime: string, refresh?: boolean, internalView?: boolean) {
-    const path = `v4${this.resourceId}/detectors/${detectorName}?startTime=${startTime}&endTime=${endTime}`;
+  public getDetector(detectorName: string, startTime: string, endTime: string, refresh?: boolean, internalView?: boolean, formQueryParams?: string) {
+    let path = `v4${this.resourceId}/detectors/${detectorName}?startTime=${startTime}&endTime=${endTime}`;
+    if(formQueryParams != undefined) {
```
puneetg1983: >formQueryParams != undefined [](start = 7, length = 28)

Better to do if (!formsQueryParams) or if (formsQueryParams != null) in typescript

AngularApp/projects/diagnostic-data/src/lib/components/form/form.component.html
```diff
@@ -0,0 +1,47 @@
+<ng-template ngFor let-element [ngForOf]="detectorForms">
```
puneetg1983: >ngForOf [](start = 32, length = 7)

More of a personal choice but across our project we are using *ngFor and this made me read what is ngForOf :-) I think using that keeps it consistent though no harm in this approach too

You can achieve the same thing with ng-container and *ngFor (if thats what made you use ngForOf instead of ngFor)

AngularApp/projects/diagnostic-data/src/lib/models/form.ts
```diff
@@ -0,0 +1,55 @@
...
+}
+export enum InputType {
+    TextBox,
+    Checkbox,
+    RadioButton,
+    DropDown,
+    Button
+}
+
+export enum ButtonStyles {
```
puneetg1983: >ButtonStyles [](start = 12, length = 12)

Just a thought :- Do you want the ability to specific "class" from the backend so detectors can just set a class attribute ? That way we have more flexibility but not sure if that gives too much variance on the UI

ApplensBackend/Controllers/DiagnosticController.cs
```diff
@@ -92,20 +94,46 @@ public async Task<IActionResult> Invoke([FromBody]JToken body)
                     }
                 }
             }
+            string scriptETag = "";
+
+            if (Request.Headers.ContainsKey("diag-script-etag"))
+            {
+                scriptETag = Request.Headers["diag-script-etag"];
+            }
```
puneetg1983: The code will reduce drastically in size with use of ternary operators here...

ApplensBackend/Services/DiagnosticRoleClientService/DiagnosticRoleClient.cs
```diff
@@ -101,7 +101,7 @@ public async Task<HttpResponseMessage> Execute(string method, string path, strin
                         {
                             requestMessage.Content = new StringContent(body ?? string.Empty, Encoding.UTF8, "application/json");
                         }
-
+                        if (additionalHeaders != null) AddAdditionalHeaders(additionalHeaders, ref requestMessage);
```
puneetg1983: >AddAdditionalHeaders [](start = 55, length = 20)

I think I read in some of our docs, we discourage single line if blocks without any parenthesis {}

AngularApp/projects/diagnostic-data/src/lib/models/form.ts
rekhaswaminathan: Nice idea. I will try to do something like this in my future iterations.

AngularApp/projects/diagnostic-data/src/lib/components/form/form.component.html
rekhaswaminathan: I see. At the time i was doing it i didnt know about ng-container and *ngFor together. Will fix it in my next iteration to maintain consistency.

ApplensBackend/Controllers/DiagnosticController.cs
rekhaswaminathan: I tried to the ternary operator way initially, but there was C# type conversion error that I got and there were if conditions inside same method used this way so i followed that.

PR #239, opened 2019-02-28
Making ARM Resource URI route case insensitive

nmallick1:
Trimmed space from resource name and made ARM Resource URI routing case insensitive.


AngularApp/projects/applens/src/app/modules/main/main/main.component.ts
```diff
@@ -93,7 +94,26 @@ export class MainComponent implements OnInit {
     }
   }
 
+  private normalizeArmUriForRoute( resourceURI: string) : string {
```
stevenernst131: This would potentially be useful other places as well, you can leave here for now, but we have an issue with case sensitivity with routes, Daniel could tell you about what he encountered.

AngularApp/projects/applens/src/app/modules/main/main/main.component.ts
```diff
@@ -93,7 +94,26 @@ export class MainComponent implements OnInit {
...
+    resourceURI = resourceURI.trim();    
+    var resourceUriPattern = /subscriptions\/(.*)\/resourceGroups\/(.*)\/providers\/(.*)/i;
+    var result = resourceURI.match(resourceUriPattern);
+    if(result && result.length === 4){
+      return "subscriptions/" + result[1] + "/resourceGroups/" + result[2] + "/providers/" + result[3];
```
stevenernst131: Nit: 
1. For strings, always use single quotes in typescript, either ` .. ` or ' .. ', the former is used for string replacement (point 2)
2. You can create the string like below in typescript:

return \`subscriptions/${result[1]}/resourceGroups/${result[2]}/providers/${result[2]}\`

AngularApp/projects/applens/src/app/modules/main/main/main.component.ts
```diff
@@ -93,7 +94,26 @@ export class MainComponent implements OnInit {
...
+      return resourceURI;
+    }    
+  }
+
   onSubmit(form: any) {
+    
+    form.resourceName = form.resourceName.trim();    
+    
+    if(this.selectedResourceType.displayName === "ARM Resource ID"){
+      form.resourceName = this.normalizeArmUriForRoute(form.resourceName);
```
stevenernst131: You should actually display an error in the case of it being invalid, otherwise nothing will happen and people will be confused.

nmallick1: Is there an existing DIV to show error messages or shall I create it ?

stevenernst131: You can create.

nmallick1: Made the change, the error message will now appear as below

image

AngularApp/projects/applens/src/app/modules/main/main/main.component.ts
nmallick1: Thanks !! Made this change

PR #240, opened 2019-03-01
Had a typo in the condition to check the error message due to which t…

nmallick1:
Had a typo in the condition to check the error message due to which the app wouldn't have loaded for any resource other than ARM resource ID.

Clearing out the error message now for other resource types and also clearing the error message if the normalization for ARM uri succeeds.


stevenernst131:
run ctrl + alt + f to format some of your brackets

PR #242, opened 2019-03-01
Allow only enabled resource types in AppLens and make those routes case insensitive 

nmallick1:
Allow only enabled resource types in AppLens and make those routes case insensitive. If some random resource is entered, the following error message will show up.

image

### DIFF
--- a/AngularApp/projects/applens/src/app/modules/main/main/main.component.ts
+++ b/AngularApp/projects/applens/src/app/modules/main/main/main.component.ts
@@ -1,13 +1,14 @@
 import { AdalService } from 'adal-angular4';
 import * as momentNs from 'moment';
-import { map } from 'rxjs/operators';
+import { map, catchError } from 'rxjs/operators';
 import { Component, OnInit } from '@angular/core';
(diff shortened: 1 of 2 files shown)


AngularApp/projects/applens/src/app/modules/main/main/main.component.ts
```diff
@@ -117,7 +138,7 @@ e.g..
     form.resourceName = form.resourceName.trim();
     
     if(this.selectedResourceType.displayName === "ARM Resource ID"){
-      form.resourceName = this.normalizeArmUriForRoute(form.resourceName);
+      form.resourceName = this.normalizeArmUriForRoute(form.resourceName, this.enabledResourceTypes);
```
stevenernst131: you don't necessarily have to pass an object that already belongs to the component

nmallick1: I did that intentionally, eventually I intend to migrate this function into common resources where other parts of the code can make use of it. Can remove it if required.

PR #246, opened 2019-03-04
Adding UI for Proactive CPU Monitoring to Portal code

puneetg1983:
Including Screenshots...

image

When you hit configure...

image

image

When you click Observe

image

When you click Analyze...

image


AngularApp/projects/app-service-diagnostics/src/app/shared/services/categories.service.ts
```diff
@@ -219,6 +219,17 @@ export class CategoriesService {
             AppStack : ''
         });
 
+        tools.Subcategories.push({
+            Name: ToolNames.CpuMonitoring,
+            BgColor: 'rgb(0, 102, 153)',
+            TextColor: 'White',
+            Href: 'tools/cpumonitoring',
+            OperatingSystem: OperatingSystem.windows,
+            AppType: AppType.WebApp | AppType.FunctionApp,
```
ShekharGupta1988: Will it work for Function App also? What about Dynamic SKU

puneetg1983: It will work for Function Apps that are in dedicated mode but not on DYNAMIC Sku. That's why I defined **Sku: Sku.NotDynamic**

puneetg1983:
@stevenernst131 - There is a lot of _English_ text in here. Can you please do a quick review of the text and confirm that it is not coming out weird  :) ?

AngularApp/projects/app-service-diagnostics/src/app/shared/components/daas/cpu-monitoring/cpu-monitoring-activity/cpu-monitoring-activity.component.html
```diff
@@ -0,0 +1,45 @@
...
+    <div class="col-sm-10">
+      <div
+        *ngIf="activeMonitoringSession.MonitoringLogs == null || activeMonitoringSession.MonitoringLogs.length === 0">
+        <i class="fa fa-circle-o-notch fa-spin spin-icon" aria-hidden="true"></i>
+        Monitoring session in progress, logs from all instances will populate shortly...
+      </div>
+
+      <div *ngIf="activeMonitoringSession.MonitoringLogs !=null && activeMonitoringSession.MonitoringLogs.length > 0"
+        class="monitoring-logs">
+        <div style="margin-bottom: 15px;">
```
dagrooms52: I think we should add something like the (Bootstrap 4 spacing helpers](https://getbootstrap.com/docs/4.0/utilities/spacing/) to prevent inline styling. I know we have Bootstrap 3 but from what I've seen its spacing utilities are incomplete. I'll work on adding those spacing helpers to our main scss

PR #247, opened 2019-03-05
Style Fixes

dagrooms52:
- Fixed style issues I noticed from reviewing a different PR
- Found global variables in C# constants file, set them to `readonly`


rekhaswaminathan:
> * Fixed style issues I noticed from reviewing a different PR

Is there change global setting that we can apply for styling?

PR #251, opened 2019-03-06
Enable TsLint

dagrooms52:
- Regenerated tsconfig.json to fix error preventing tslint from running
- Added sonar-ts for improved static analysis on typescript code
- Install TSLint extension in VS Code to use the linter


AngularApp/tsconfig.json
```diff
@@ -1,30 +1,75 @@
 {
   "compileOnSave": false,
   "compilerOptions": {
-    "baseUrl": "./",
```
stevenernst131: I think you need this baseUrl, there are errors in your build related to this.

PR #253, opened 2019-03-07
Decommission ZRay feature

ShekharGupta1988:
### Description
Azure app services is decommissioning Z-Ray for web apps. Starting Monday (3/11), no new web apps will be able to install this add-on. The existing sites which has this add-on will continue to use it till some time. These sites have been proactively notified.

### Changes on our Side
- For sites which doesn't use Zray, the option to enable it (shown as `PHP Debugging` in `Diagnostic Tools`) will not show up
- If the site has Zray premier add-on enabled, the option will show up and instead of opening Ibiza blade(which is also getting removed), customers will be redirected to ZRay Site Extension in Kudu.

### DIFF
--- a/AngularApp/projects/app-service-diagnostics/src/app/resources/web-sites/services/site-feature.service.ts
+++ b/AngularApp/projects/app-service-diagnostics/src/app/resources/web-sites/services/site-feature.service.ts
@@ -117,7 +117,7 @@ export class SiteFeatureService extends FeatureService {
       
       if (data && data.value) {
         let premierAddOns: any[] = data.value;
-        let zRayAddOn = premierAddOns.find(x => (x.plan && (x.plan.product == "z-ray")));
+        let zRayAddOn = premierAddOns.find(x => (x.plan && (x.plan.product === "z-ray")));
         if (zRayAddOn) {
           this.premiumTools.push({
             appType: AppType.WebApp,
--- a/AngularApp/projects/app-service-diagnostics/src/app/resources/web-sites/services/site-feature.service.ts
+++ b/AngularApp/projects/app-service-diagnostics/src/app/resources/web-sites/services/site-feature.service.ts
@@ -33,7 +33,7 @@ export class SiteFeatureService extends FeatureService {
       }
       this.addDiagnosticTools(startupInfo.resourceId);
       this.addProactiveTools(startupInfo.resourceId);
-      this.addPremiumTools();
+      this.addPremiumTools(startupInfo.resourceId);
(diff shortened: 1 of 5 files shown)


AngularApp/projects/app-service-diagnostics/src/app/resources/web-sites/services/site-feature.service.ts
```diff
@@ -93,24 +93,8 @@ export class SiteFeatureService extends FeatureService {
...
-        sku: Sku.NotDynamic,
-        stack: '',
-        item: {
-          id: 'zray',
-          name: 'PHP Debugging',
-          category: 'Premium Tools',
-          description: '',
-          featureType: FeatureTypes.Tool,
-          clickAction: this._createFeatureAction('zray', 'Premium Tools', () => {
-            this._portalActionService.openPHPDebuggingBlade();
```
cindylovescoding: Since this is not used anymore, can we delete this function then?

AngularApp/projects/app-service-diagnostics/src/app/resources/web-sites/services/site-feature.service.ts
```diff
@@ -128,6 +112,36 @@ export class SiteFeatureService extends FeatureService {
...
+            stack: '',
+            item: {
+              id: 'zray',
+              name: 'PHP Debugging',
+              category: 'Premium Tools',
+              description: '',
+              featureType: FeatureTypes.Tool,
+              clickAction: this._createFeatureAction('zray', 'Premium Tools', () => {
+                const site: Site = <Site>this._resourceService.resource.properties;
+                let scmHostName = site.enabledHostNames.find(h => h.indexOf('.scm.') > 0);
```
cindylovescoding: What will happen if we can not find Kudu site for ZRay?

AngularApp/projects/app-service-diagnostics/src/app/resources/web-sites/services/site-feature.service.ts
ShekharGupta1988: Good point. I'll remove it.

AngularApp/projects/app-service-diagnostics/src/app/resources/web-sites/services/site-feature.service.ts
ShekharGupta1988: so this option will only show up for sites that has Zray enabled
so the kudu will have the site extension installed.

PR #258, opened 2019-03-12
AB Testing - Toggling Availability and Performance view for Case Submission after checking subscriptionId

puneetg1983:
When we decide to turn this ON, we will have to set enableV2  to true...

### DIFF
--- a/AngularApp/projects/app-service-diagnostics/src/app/resources/web-sites/services/web-sites.service.ts
+++ b/AngularApp/projects/app-service-diagnostics/src/app/resources/web-sites/services/web-sites.service.ts
@@ -29,7 +29,7 @@ export class WebSitesService extends ResourceService {
     }
 
     public get azureServiceName(): string {
-        return this.appType === AppType.WebApp ? this.platform === OperatingSystem.windows ? 'Azure Web App(Windows)' : 'Azure Web App(Linux)' : 'Azure Function App';
+        return this.appType === AppType.WebApp ? this.platform === OperatingSystem.windows ? 'Web App (Windows)' : 'Web App (Linux)' : 'Azure Function App';
     }
 
     public get isApplicableForLiveChat(): boolean {
--- a/AngularApp/projects/app-service-diagnostics/src/app/resources/web-sites/services/web-sites.service.ts
+++ b/AngularApp/projects/app-service-diagnostics/src/app/resources/web-sites/services/web-sites.service.ts
@@ -29,7 +29,7 @@ export class WebSitesService extends ResourceService {
     }
 
     public get azureServiceName(): string {
-        return this.appType === AppType.WebApp ? this.platform === OperatingSystem.windows ? 'Web App (Windows)' : 'Web App (Linux)' : 'Azure Function App';
+        return this.appType === AppType.WebApp ? this.platform === OperatingSystem.windows ? 'Web App (Windows)' : 'Web App (Linux)' : 'Function App';
     }
 
     public get isApplicableForLiveChat(): boolean {
--- a/AngularApp/projects/app-service-diagnostics/src/app/shared/services/logging/logging.service.ts
+++ b/AngularApp/projects/app-service-diagnostics/src/app/shared/services/logging/logging.service.ts
@@ -23,7 +23,7 @@ export class LoggingService {
     private _supportTopicId: string = '';
     private _appType: string = '';
     private _source: string = '';
-    private _ticketSubmissionFlowVersion: string = '';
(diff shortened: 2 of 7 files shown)


AngularApp/projects/app-service-diagnostics/src/app/shared/utilities/versioningHelper.ts
```diff
@@ -0,0 +1,12 @@
+export class VersioningHelper{
+
+static isV2Subscription(subscriptionId:string):boolean {
+      
+    // When we decide to enable the A/B testing for v2, change this to true
+    let enableV2 = false;
+
+    let firstDigit = "0x" + subscriptionId.substr(0, 1);
+    return (parseInt(firstDigit, 16) % 2 == 1) && enableV2;
```
cindylovescoding: Do we know what percentage of subscriptions will return true for this?

puneetg1983: It's almost a 50-50 breakdown if we go with this logic...

Here is the data for last 7 days...

v1v2Comparison

Also added you to an email thread....

cindylovescoding: Got it!

hforeste: Hey Puneet, do you wan't to start with 50-50 or should we start small with something like 30-70?

puneetg1983: Its all in our control. If everyone thinks we should go 70-30, thats totally possible

AngularApp/projects/app-service-diagnostics/src/app/shared/services/logging/logging.service.ts
```diff
@@ -50,6 +52,7 @@ export class LoggingService {
 
                 if (this._startUpInfo.workflowId) {
                     this._ticketBladeWorkflowId = this._startUpInfo.workflowId;
+                    this._ticketSubmissionFlowVersion =  VersioningHelper.isV2Subscription(this._subscriptionId) ? "v2" : "v1" ;
```
ShekharGupta1988: The name is confusing with the case Submission UI (v1/v2)
Should this be called DiagnosticsVersion or DetectorVersion

ShekharGupta1988:
@puneetg1983 - If you make the v3 detector public, in Diagnose and Solve blade, Will it show two tiles for Web App down ?

puneetg1983:
If we make the v3 detectors public, we will still see the old tiles (v2 Tiles) in Diagnose and Solve provided we do not add the category as AvailabilityAndPerformance (which we wont). The only place where the new detectors will show up will be in the search but no where else.

ShekharGupta1988:
> If we make the v3 detectors public, we will still see the old tiles (v2 Tiles) in Diagnose and Solve provided we do not add the category as AvailabilityAndPerformance (which we wont). The only place where the new detectors will show up will be in the search but no where else.

Hmm.. I see. I think we should enable this versioning in diagnose and solve also then. Having two entries even in search for same scenario is not a good experience.

AngularApp/projects/app-service-diagnostics/src/app/shared/utilities/versioningHelper.ts
puneetg1983: Changed to a 70:30 logic...

puneetg1983:
Discussed with @ShekharGupta1988 , the new entries will search will be labeled either v2 (like Web App Down v2) or something like HTTP Server Errors so the search experience wouldn't show up as duplicate entries....

AngularApp/projects/app-service-diagnostics/src/app/resources/web-sites/services/web-sites.service.ts
```diff
@@ -29,7 +29,7 @@ export class WebSitesService extends ResourceService {
     }
 
     public get azureServiceName(): string {
-        return this.appType === AppType.WebApp ? this.platform === OperatingSystem.windows ? 'Azure Web App(Windows)' : 'Azure Web App(Linux)' : 'Azure Function App';
+        return this.appType === AppType.WebApp ? this.platform === OperatingSystem.windows ? 'Web App (Windows)' : 'Web App (Linux)' : 'Azure Function App';
```
ShekharGupta1988: make it "Function App"

PR #261, opened 2019-03-13
Local development supporting gist

crazyxy:
You can do:
- List all gists
image

- List all versions of a gist and browse the gist in applens
image

- Install gist
image

- Run detector
- Publish detector
- Publish gist from local dev

You cannot do (todo):
- <del>Create a gist from local dev</del> (have a separate pr for this feature)


ApplensBackend/Controllers/LocalDevelopmentController.cs
```diff
@@ -37,7 +61,36 @@ public async Task<IActionResult> PrepareLocalDevEnvironementstring (string detec
...
+
+            var config = string.Empty;
+            var sourceReference = new Dictionary<string, string>();
+            if (body["configuration"] != null)
+            {
+                config = body["configuration"].ToString();
+                var dependencies = body["configuration"]["dependencies"].ToObject<Dictionary<string, string>>();
+                foreach (var p in dependencies)
+                {
+                    sourceReference.Add(p.Key.ToLower(), await GithubClient.GetCommitContent(p.Key.ToLower(), p.Value));
```
cindylovescoding: Do we want to make all the sync calls here?

cindylovescoding:
When we run the "List gists" , can you give a more descriptive console output?
Eg.
version: sha   Check the gist content here: applensURL

ApplensBackend/Controllers/LocalDevelopmentController.cs
crazyxy: It's async call.

ApplensBackend/Controllers/LocalDevelopmentController.cs
```diff
@@ -37,7 +61,36 @@ public async Task<IActionResult> PrepareLocalDevEnvironementstring (string detec
...
+            var gists = new List<string>();
+            if (body["gists"] != null)
+            {
+                gists = body["gists"].ToObject<List<string>>();
+            }
+
+            var gistDef = new Dictionary<string, IEnumerable<string>>();
+            foreach (var gist in gists)
+            {
+                var commits = await GithubClient.GetAllCommits(gist);
```
hforeste: Because you are awaiting a task each time in a foreach loop, these tasks will get finished one at a time. Consider, adding the task return from `GithubClient.GetAllCommits(gists)` into a task list and use `Task.WhenAll()` so that each task can be run in separate threads.

ApplensBackend/Controllers/LocalDevelopmentController.cs
hforeste: I think what @cindylovescoding is saying here is the same thing I'm saying in https://github.com/Azure/Azure-AppServices-Diagnostics-Portal/pull/261/files#r265351974

ApplensBackend/LocalDevelopmentTemplate/Framework/Tools/StartUp.ps1
```diff
@@ -44,34 +39,33 @@ function System-Check
...
+function Copy-FrameworkDependencies {
     [CmdletBinding()]
     param
     (
-    [Parameter(Mandatory=$false)]
-    [string]
-    $buildPath = "\\reddog\builds\branches\rd_websites_stable_release\",
+        [Parameter(Mandatory = $false)]
+        [string]
+        $buildPath = "\\reddog\builds\branches\rd_websites_stable_release\",
```
hforeste: @ShekharGupta1988 @cindylovescoding we need to change this build path to \\reddog\builds\branches\git_aapt_antares_websites_master

hforeste: @crazyxy feel free to make that change but it will need to be tested so you can leave it for @cindylovescoding if you'd like

crazyxy:
> When we run the "List gists" , can you give a more descriptive console output?
> Eg.
> version: sha Check the gist content here: applensURL

Use `Format-Table` to format the table but no color highlight.

PR #266, opened 2019-03-14
Log Unhandled Exceptions to Kusto

dagrooms52:
- Implement unhandled exception handler which logs fatal exceptions to Kusto for monitoring and alerting
- Fully implement kusto logging to match ITelemetryProvider interface

### DIFF
--- a/AngularApp/projects/app-service-diagnostics/src/app/shared/services/portal-kusto-telemetry.service.ts
+++ b/AngularApp/projects/app-service-diagnostics/src/app/shared/services/portal-kusto-telemetry.service.ts
@@ -9,25 +9,49 @@ export class PortalKustoTelemetryService implements ITelemetryProvider {
   }
 
   logEvent(eventMessage: string, properties: { [name: string]: string }, measurements?: any) {
-    // measurements is ignored
-    this._portalService.logAction('diagnostic-data', eventMessage, properties);
+    this._portalService.logAction('diagnostic-data', eventMessage, {
+      ...properties,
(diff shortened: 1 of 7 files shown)


AngularApp/projects/app-service-diagnostics/src/app/shared/services/portal-kusto-telemetry.service.ts
```diff
@@ -9,25 +9,49 @@ export class PortalKustoTelemetryService implements ITelemetryProvider {
   }
 
   logEvent(eventMessage: string, properties: { [name: string]: string }, measurements?: any) {
-    // measurements is ignored
-    this._portalService.logAction('diagnostic-data', eventMessage, properties);
+    this._portalService.logAction('diagnostic-data', eventMessage, {
+      ...properties,
```
stevenernst131: does the '...' notation work with AOT build? I have never used it in this project, only in a little work I did with react. I remember seeing something about needing to have a special build handling for it, could be wrong though

dagrooms52: Not sure, is there an easy way to test AOT in this project? It seems like Typescript handles the ES6 features anyway

dagrooms52: It did work with `npm run applens` which is `"applens": "ng serve applens --aot --optimization=false"` so the feature is enabled in AOT

PR #278, opened 2019-03-21
Reduce the height of gist management

crazyxy:
image


AngularApp/projects/applens/src/app/modules/dashboard/onboarding-flow/onboarding-flow.component.html
```diff
@@ -31,7 +31,7 @@ <h4>Gist Version Management</h4>
         </div>
       </div>
 
-      <div style="margin-bottom: 10px">
+      <div style="margin-top: 10px;margin-bottom: 10px">
```
dagrooms52: You can use `class="my-2"` for similar behavior. Putting styles directly in html tags is generally avoided.

PR #279, opened 2019-03-21
Fix dynamic insight always expanded

cindylovescoding:


AngularApp/projects/diagnostic-data/src/lib/components/dynamic-insight/dynamic-insight.component.ts
```diff
@@ -38,7 +38,7 @@ export class DynamicInsightComponent extends DataRenderBaseComponent {
       title: this.renderingProperties.title,
       description: this._markdownService.compile(this.renderingProperties.description),
       status: this.renderingProperties.status,
-      expanded: this.renderingProperties.expanded ? this.renderingProperties.expanded : true,
+      expanded: this.renderingProperties.expanded !== undefined? this.renderingProperties.expanded : true,
```
dagrooms52: Should `expanded == null` be caught in this default case as well?

cindylovescoding: Yes I think it works for both null and undefined

dagrooms52: I'm seeing that `null !== undefined` evaluates to `true` which would cause `expanded = null` in this case. I think we would want the other case to apply, making `expanded = true`.

cindylovescoding: Using null and comparing with null are not recommended in typescript. https://medium.com/@basarat/null-vs-undefined-in-typescript-land-dc0c7a5f240a
https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#null-and-undefined
But that could also happen, since they are not technically equal to each other. Let me add that check also.

dagrooms52: I agree that we shouldn't explicitly set `null`. If a null value were to come through from using a library though, the double equals (`==`) check should cover both cases. It looks like `null == undefined` is `true` whereas `null === undefined` is `false`.

cindylovescoding: To handle this tricky part, we decided to change that into: this.renderingProperties.expanded != undefined, that should cover both scenarios

puneetg1983: My understanding is that the below line will check for both null and undefined.

`if ( x == null )`

Me and @stevenernst131 once discussed this and I learnt that checking for !=null is better than checking for undefined 

I would be interested to know if this works?

`expanded: this.renderingProperties.expanded != null ? this.renderingProperties.expanded : true`

https://stackoverflow.com/questions/28975896/is-there-a-dedicated-function-to-check-null-and-undefined-in-typescript

cindylovescoding: Yes Puneet that will work, similarly this will also work: 
`expanded: this.renderingProperties.expanded != undefined ? this.renderingProperties.expanded : true.  
`
Double quotes will do the type conversion first thus "undefined == null" returns true. 

But since it's best practice to use tripple equals (===) and inequality comparison operators (!==), maybe it's better to check in the following way ?
` this.renderingProperties.expanded !== undefined && this.renderingProperties.expanded !== null`
https://bytearcher.com/articles/equality-comparison-operator-javascript/

PR #281, opened 2019-03-22
Change to new share path with the production backend dlls 

cindylovescoding:
Change to new share path with the production backend dlls

### DIFF
--- a/ApplensBackend/LocalDevelopmentTemplate/StartUp.cmd
+++ b/ApplensBackend/LocalDevelopmentTemplate/StartUp.cmd
@@ -3,3 +3,4 @@ Powershell.exe Unblock-File .\Framework\Tools\*
 Powershell.exe Unblock-File .\Framework\Tools\References\Microsoft.IdentityModel.Clients.ActiveDirectory\*
 Powershell.exe Unblock-File .\Framework\Tools\References\Newtonsoft.Json\*
 powershell.exe -ExecutionPolicy Unrestricted -File .\Framework\Tools\StartUp.ps1
+PAUSE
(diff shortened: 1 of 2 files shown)


ApplensBackend/LocalDevelopmentTemplate/StartUp.cmd
```diff
@@ -2,4 +2,5 @@ Powershell.exe Unblock-File .\Detector\*
 Powershell.exe Unblock-File .\Framework\Tools\*
 Powershell.exe Unblock-File .\Framework\Tools\References\Microsoft.IdentityModel.Clients.ActiveDirectory\*
 Powershell.exe Unblock-File .\Framework\Tools\References\Newtonsoft.Json\*
-powershell.exe -ExecutionPolicy Unrestricted -File .\Framework\Tools\StartUp.ps1
\ No newline at end of file
+powershell.exe -ExecutionPolicy Unrestricted -File .\Framework\Tools\StartUp.ps1
+PAUSE
```
ShekharGupta1988: wanted to remove the PAUSE ?

cindylovescoding: Shekhar I have noticed that even when there is a terminate error, the window will dismiss if we launch the startup.cmd by double click it. Do you think we should keep the window open after the execution finished until user press any key?

ShekharGupta1988: I understand your point. Maybe Pause it only when there is an error.
Having them to press Enter for success case also seems like an extra step.

cindylovescoding: Yeah I was not able to find a decent good way to do that previously.. Let's remove that for now, and add the check back once I have a solution

PR #282, opened 2019-03-26
Style fix in modal

cindylovescoding:
Before:
image

After:
image

### DIFF
--- a/AngularApp/projects/applens/src/styles.scss
+++ b/AngularApp/projects/applens/src/styles.scss
@@ -74,23 +74,13 @@ body {
 }
 
 .wide-modal {
-  max-width: 100%;
-  width: 600px;
+  max-width: 560px!important;
+  padding: 0px!important;
 }
 
 .package-modal {
   max-width: 100% !important;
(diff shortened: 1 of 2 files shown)


ShekharGupta1988:
Small Request @cindylovescoding  - Can you also fix the paddings between the "Publish" and "Cancel" button?

AngularApp/projects/applens/src/styles.scss
```diff
@@ -74,25 +74,15 @@ body {
 }
 
 .wide-modal {
-  max-width: 100%;
-  width: 600px;
+  max-width: 560px!important;
```
dagrooms52: Can we increase specificity instead of using `!important`? This may disrupt other elements in the future.

cindylovescoding:
@ShekharGupta1988  Yes I have added the padding between those buttons. @dagrooms52 I was trying to avoid using !important, but in this case the ngx modal are always using reflection class styles. That's why I created a specific class which is only used by that specific element, and overwrite the reflected class style for that element.

PR #283, opened 2019-03-26
Solutions Using Gists and Vertical Selection Panel

dagrooms52:
- Solutions can be written and imported to a Detector via Gists
- Integrated vertical tab selector for multiple Solutions

Customer (external) view:
solution-external

Support (internal) view:
solution-internal


AngularApp/projects/app-service-diagnostics/src/app/shared/services/arm.service.ts
```diff
@@ -19,12 +19,18 @@ export class ArmService {
     public storageApiVersion = '2015-05-01-preview';
     public websiteApiVersion = '2015-08-01';
 
-    constructor(private _http: HttpClient, private _authService: AuthService, private _cache: CacheService) {
+    constructor(private _http: HttpClient, private _authService: AuthService, private _cache: CacheService) { }
+
+    createUrl(resourceUri: string, apiVersion?: string) {
+        return `${this.armUrl}${resourceUri}${resourceUri.indexOf('?') >= 0 ? '&' : '?'}` +
+            `api-version=${!!apiVersion ? apiVersion : this.websiteApiVersion}`;
```
cindylovescoding: Just curious , is this !!apiVersion different from using apiVersion ? apiVersion : this.websiteApiVersion?

AngularApp/projects/app-service-diagnostics/src/app/shared/services/generic-solution.service.ts
```diff
@@ -0,0 +1,113 @@
...
+    if (verb === 'get') {
+      return this.armService.getResourceFullResponse(route, true, apiVersion);
+    }
+
+    let actionMethod = `${verb}ResourceFullResponse`;
+    return this.armService[actionMethod](route, body, true, apiVersion);
+  }
+
+  OpenTab(resourceUri: string, actionOptions: {tabUrl: string}): Observable<any> {
+    this.logService.logEvent('SolutionOpenTab', {'resourceUri': resourceUri, ...actionOptions});
```
cindylovescoding: Should this be logged only after assertPropertyExist and tab was opened?

AngularApp/projects/app-service-diagnostics/src/app/shared/services/generic-solution.service.ts
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  getBladeInfo(options: {}) {
+    this.assertPropertyExists(options, 'detailBlade');
+
+    return options as OpenBladeInfo;
+  }
+
+  GoToBlade(resourceUri: string, actionOptions: {detailBlade: string}): Observable<any> {
+    this.logService.logEvent('SolutionGoToBlade', {'resourceUri': resourceUri, ...actionOptions});
```
cindylovescoding: Same here

AngularApp/projects/diagnostic-data/src/lib/components/solution/solution.component.ts
```diff
@@ -53,75 +33,74 @@ export class SolutionComponent extends DataRenderBaseComponent {
...
+      case ActionType.GoToBlade: {
+        return this._siteService.GoToBlade(this.solution.ResourceUri, options);
+      }
+      case ActionType.OpenTab: {
+        return this._siteService.OpenTab(this.solution.ResourceUri, options);
+      }
     }
 
-    return resultText
+    throw new Error(`Not Implemented: Solution Service does not have an implementation for the action.`)
```
cindylovescoding: Move this inside the switch block as a default case?

AngularApp/projects/app-service-diagnostics/src/app/shared/services/generic-solution.service.ts
cindylovescoding: Remove the spaces before actionOptions?

AngularApp/projects/app-service-diagnostics/src/app/shared/services/arm.service.ts
dagrooms52: I think this is used because the original type of apiVersion can be a truthy false-positive. I'm not sure if it's necessary.

AngularApp/projects/diagnostic-data/src/lib/components/solution/solution.component.ts
dagrooms52: Weird Typescript functionality doesn't allow the default case since the type is "Never" - so it's actually a compiler error to make a default case. It really upset me to have to do it this way 😐

AngularApp/projects/app-service-diagnostics/src/app/shared/services/generic-solution.service.ts
dagrooms52: That's the expansion operator from ES6, not spaces:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax

dagrooms52: I do want to make sure that we log incorrect usage in case a detector author messed up the API. Do you think it should be a different event if validation failed?

cindylovescoding: Yeah if we are evaluate the user engagement data for this action, we may want to ensure customer is successfully engaged. A different event for failures will be helpful.

AngularApp/projects/diagnostic-data/src/lib/components/solution/solution.component.ts
cindylovescoding: I see..

AngularApp/projects/app-service-diagnostics/src/app/shared/services/generic-solution.service.ts
cindylovescoding: Interesting hh

dagrooms52: Actually the unhandled exception handler will catch incorrect usage, so I'll just move the success event logging to the end of the method after the validation and action

puneetg1983:
Hey @dagrooms52  - Looks like there are some conflicts. Can you resolve them first ?

AngularApp/projects/app-service-diagnostics/src/app/shared/services/site.service.ts
```diff
@@ -148,12 +148,12 @@ export class SiteService {
     }
 
     getSiteAppSettings(subscriptionId: string, resourceGroup: string, siteName: string, slot: string = ''): Observable<any> {
-        const url: string = this._uriElementsService.getListAppSettingsUrl(subscriptionId, resourceGroup, siteName, slot);
```
puneetg1983: >: string  [](start = 17, length = 9)

Just curious - are all these changes result of running some automated tool on our project ?

AngularApp/projects/app-service-diagnostics/src/app/shared/services/generic-solution.service.ts
```diff
@@ -0,0 +1,111 @@
...
+import { OpenBladeInfo } from '../models/portal';
+import { ArmService } from './arm.service';
+import { PortalActionService } from './portal-action.service';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class GenericSolutionService {
+
+  whitelistedRoutes = {
```
ShekharGupta1988: lets rename this to "allowedRoutes" :)

AngularApp/projects/app-service-diagnostics/src/app/shared/services/generic-solution.service.ts
```diff
@@ -0,0 +1,111 @@
...
+    this.assertPropertyExists(options, 'detailBlade');
+
+    return options as OpenBladeInfo;
+  }
+
+  GoToBlade(resourceUri: string, actionOptions: {detailBlade: string}): Observable<any> {
+    const bladeInfo = this.getBladeInfo(actionOptions);
+    this.logService.logEvent('SolutionGoToBlade', {'resourceUri': resourceUri, ...actionOptions});
+
+    switch (bladeInfo.detailBlade) {
```
ShekharGupta1988: You want to possibly ignore case sensitive match here

AngularApp/projects/diagnostic-data/src/lib/utilities/uri-utilities.ts
```diff
@@ -0,0 +1,7 @@
+export class UriUtilities {
+  static BuildDetectorLink(resourceUri: string, detectorId: string): string {
+    return "https://ms.portal.azure.com" +
```
puneetg1983: >ms.portal.azure.com [](start = 20, length = 19)

Should this be portal.azure.com? I thought the ms. is added for MS subscriptions but not really sure.

AngularApp/projects/diagnostic-data/src/lib/components/vertical-display-list/vertical-display-list.component.html
```diff
@@ -0,0 +1,35 @@
...
+  <div class="outer-container">
+
+    <div class="column-container left-column-container" *ngIf="!smallMenu">
+      <div class="list-item" [ngClass]="{'list-item-selected':listItem.metaData.isSelected}"
+          *ngFor="let listItem of listItems" (click)="selectItem(listItem)">
+        <span class="list-item-text">
+          <b>{{listItem.metaData.title}}</b>
+        </span>
+        <div class="tag-holder">
+          <solution-type-tag *ngFor="let tag of listItem.metaData.tags" [tagType]="tag"></solution-type-tag>
```
puneetg1983: >solution-type-tag [](start = 11, length = 17)

Does this mean one solution can have multiple tags like Investigation and Mitigation?

AngularApp/projects/app-service-diagnostics/src/app/shared/services/site.service.ts
dagrooms52: No, I found duplicate code and took the opportunity to refactor it out

AngularApp/projects/app-service-diagnostics/src/app/shared/services/generic-solution.service.ts
dagrooms52: I wondered if that would be the case - I haven't tested the portal API that receives this data, but is it also case insensitive? It seemed to be sensitive based on the code in the underlying API wrapper.

AngularApp/projects/app-service-diagnostics/src/app/shared/services/site.service.ts
dagrooms52: No, but I'd like to incorporate a style tool that stops type declaration and inference on the same line. Just like using `var` in C#, the code looks cleaner when we just infer type

AngularApp/projects/diagnostic-data/src/lib/utilities/uri-utilities.ts
dagrooms52: I didn't know that, good catch

AngularApp/projects/diagnostic-data/src/lib/components/vertical-display-list/vertical-display-list.component.html
dagrooms52: I copied that from the way old solutions handled tags. Should we restrict it to a single tag instead?

PR #284, opened 2019-03-26
Time range check on detectors

nmallick1:
Added a check in the date range selection control to check for invalid dates. Also defaulting to past 24 hours in case someone bypasses the UX and hits the applens URL directly. Error messages are displayed as follows.

I am aesthetically challenged, please suggest a better highlighting color for the error message :)

image

### DIFF
--- a/AngularApp/projects/diagnostic-data/src/lib/components/detector-control/detector-control.component.scss
+++ b/AngularApp/projects/diagnostic-data/src/lib/components/detector-control/detector-control.component.scss
@@ -1,3 +1,5 @@
+@import "../../../../../app-service-diagnostics/src/styles/main.scss";
+
 .control-container {
     display: flex;
 }
@@ -52,6 +54,5 @@
 }
 
 .errorMessage {
-    background-color: darkred;
-    color: white;
+    color: $error-color;
 }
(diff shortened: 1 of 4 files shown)


AngularApp/projects/diagnostic-data/src/lib/components/detector-control/detector-control.component.scss
```diff
@@ -49,4 +49,9 @@
 
 .date-control > input:focus {
     outline: none;
+}
+
+.errorMessage {
+    background-color: darkred;
```
dagrooms52: If you import `main.scss` it will include `_variables.scss` which defines some nice error colors. The first one I saw was `$error-color: #e81123;`, you might want to try that.

nmallick1:
Post the suggested changes, error message appears as below.

image

AngularApp/projects/diagnostic-data/src/lib/services/detector-control.service.ts
```diff
@@ -64,7 +67,79 @@ export class DetectorControlService {
...
+                returnValue = 'Start and End date time cannot be equal.';
+              }
+              else {
+                returnValue = '';
+              }
+            }
+          }
+        }
+        else {
+          console.log(start);
```
ShekharGupta1988: lets get rid of console messages

nmallick1: Thanks for catching that. Removed it.

PR #285, opened 2019-03-27
Enabling CPU detector in case submission

puneetg1983:
Enabling CPU detector instead of web app down in case submission
And enabling this for beta subscriptions so we can test it before we start A/B testing.

### DIFF
--- a/AngularApp/projects/app-service-diagnostics/src/app/shared/utilities/versioningHelper.ts
+++ b/AngularApp/projects/app-service-diagnostics/src/app/shared/utilities/versioningHelper.ts
@@ -8,9 +8,13 @@ export class VersioningHelper {
         let enableV2 = false;
 
         let isBetaSubscription = DemoSubscriptions.betaSubscriptions.findIndex(item => subscriptionId.toLowerCase() === item.toLowerCase()) > -1;
+        if (isBetaSubscription) {
+            return true;
+        }
+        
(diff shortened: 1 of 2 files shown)


AngularApp/projects/app-service-diagnostics/src/app/shared/utilities/versioningHelper.ts
```diff
@@ -8,9 +8,13 @@ export class VersioningHelper {
         let enableV2 = false;
 
         let isBetaSubscription = DemoSubscriptions.betaSubscriptions.findIndex(item => subscriptionId.toLowerCase() === item.toLowerCase()) > -1;
```
cindylovescoding: Do we still need this?

puneetg1983: Yea , I am enabling it for just a few internal subs to make sure things are working and for testing. I will send another pr to set enableV2 to true which will enable it for external subs.

cindylovescoding: Got it!

puneetg1983: ping @cindylovescoding , @ShekharGupta1988 , @hforeste

PR #289, opened 2019-04-01
Nmallick1/time range check on detectors

nmallick1:
If an empty end date was supplied, we defaulted to startdate + 1 day however the start date might be < 15 minutes from now. Handled this condition and accordingly calculating the start and end dates.

### DIFF
--- a/AngularApp/projects/diagnostic-data/src/lib/services/detector-control.service.ts
+++ b/AngularApp/projects/diagnostic-data/src/lib/services/detector-control.service.ts
@@ -179,7 +179,7 @@ export class DetectorControlService {
       if (this.timeRangeErrorString === 'Selected time duration must be at least 15 minutes.') {
         this.timeRangeErrorString = 'Defaulting to a 15 minutes duration. Selected time duration was less than 15 minutes.';
         this._endTime = endTime;
-        this._startTime = this._endTime.clone().subtract(30, 'minutes');
+        this._startTime = this._endTime.clone().subtract(15, 'minutes');
       }
       else {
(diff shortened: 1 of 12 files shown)


AngularApp/projects/diagnostic-data/src/lib/services/detector-control.service.ts
```diff
@@ -182,9 +190,23 @@ export class DetectorControlService {
         this._startTime = this._endTime.clone().subtract(15, 'minutes');
       }
       else {
-        this.timeRangeErrorString = `Defaulting to last 24 hrs. Start and End date time must not be more than ${(this.allowedDurationInDays * 24).toString()} hrs apart and Start date must be within the past 30 days.`;
-        this._endTime = moment.utc(moment());
-        this._startTime = this._endTime.clone().subtract(1, 'days');
+        if (this.timeRangeErrorString === 'Empty End date time supplied.') {
+          this._startTime = moment.utc(start);
+          if (moment.duration(moment.utc(moment()).diff(this._startTime)).asMinutes() < 15) {
```
hforeste: Do you have to do `moment.utc(moment())`. Not sure what version of moment.js this is but according to the docs if you want to get current UTC time you can just do `moment.utc()` https://momentjs.com/docs/#/parsing/utc/

AngularApp/projects/diagnostic-data/src/lib/services/detector-control.service.ts
```diff
@@ -182,9 +190,23 @@ export class DetectorControlService {
...
+            this._startTime = moment.utc(moment()).subtract(15, 'minutes');
```
hforeste: I think the same here as well

nmallick1: Thank you for the suggestion Hawk, you are correct. I removed the unnecessary calls and re-submitted the PR

PR #290, opened 2019-04-02
Remove markdown border if no title is given to markdown view

cindylovescoding:
image

### DIFF
--- a/AngularApp/projects/diagnostic-data/src/lib/components/markdown/markdown.component.html
+++ b/AngularApp/projects/diagnostic-data/src/lib/components/markdown/markdown.component.html
@@ -8,7 +8,6 @@
         <span class="fa fa-envelope"></span>
         Open in Outlook
     </button>
-
 </ng-template>
 
 <data-container *ngIf="renderingProperties.isContainerNeeded else noContainerMarkdown" [headerTemplate]="header" [title]="renderingProperties.title">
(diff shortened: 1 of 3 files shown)


hforeste:
Hey Cindy, by removing the border does it also remove the open in outlook and copy to email
image

cindylovescoding:
Yes @hforeste , it will also remove the header templates and buttons.

ShekharGupta1988:
Not sure how many people are actually using that feature to be honest. Cindy - you can pull that info from kusto just to check the clicks on those buttons.

It might be better to provide a way for authors to specify whether they need the container or not from authoring experience.

hforeste:
> Yes @hforeste , it will also remove the header templates and buttons.

Why are we removing this?

hforeste:
> Not sure how many people are actually using that feature to be honest. Cindy - you can pull that info from kusto just to check the clicks on those buttons.
> 
> It might be better to provide a way for authors to specify whether they need the container or not from authoring experience.

I am also for having this behavior where the author can decide

ShekharGupta1988:
> > Yes @hforeste , it will also remove the header templates and buttons.
> 
> Why are we removing this?

There are scenarios where people want to just add a Title to the section of data  or some description (in markdown). (Xin wanted that for Restart analysis, Marie wanted that for few function detectors..)

The problem with Markdown rendering is anytime you use res.AddMarkdownView(..), it add those borders with buttons, even to just put a H3 tag on the page.

cindylovescoding:
@ShekharGupta1988 @hforeste I have sent out another PR in backend to add the flag while AddMarkdownview
https://github.com/Azure/Azure-AppServices-Diagnostics/pull/135a

 Will update this PR .

PR #296, opened 2019-04-04
Unhandled Exception Logging

dagrooms52:
- Log unhandled exceptions in Kusto


dagrooms52:
@puneetg1983 Since this caused so many issues last time, can you please check out the branch and run any smoke tests you can think of to make sure we can deploy this safely?

hforeste:
Hey @dagrooms52, do we know what the root cause was the last time we had issues when deploying the exception handling logic?

dagrooms52:
> Hey @dagrooms52, do we know what the root cause was the last time we had issues when deploying the exception handling logic?

The exception was being re-thrown, but this time I realized we can't re-throw it because we're already in the exception processing pipeline, so the only thing to catch it would be that pipeline. That's why there is a try/catch around the logging call that just says `// squash logging exception`

PR #297, opened 2019-04-05
Few string updates for Chat

ShekharGupta1988:
Changes:
- CSAT string update for Chat
- Changing the doc search endpoint


Backend/Controllers/AppSettingsController.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        [HttpGet("{name}")]
+        public IActionResult GetAppSettingValue(string name)
+        {
+            if (string.IsNullOrWhiteSpace(name))
+            {
+                return BadRequest("App setting name is empty");
+            }
+
+            return Ok(config[name]);
```
cindylovescoding: Will there be any exception if the setting key is not found?

PR #298, opened 2019-04-08
Disable eventviewer and freb on ILB ASE

puneetg1983:


AngularApp/projects/app-service-diagnostics/src/app/resources/web-sites/models/site-filter.ts
```diff
@@ -8,5 +8,6 @@ export interface SiteFilteredItem<T> {
     appType: AppType;
     stack: string;
     sku: Sku;
+    hostingEnvironmentKind: HostingEnvironmentKind;
```
hforeste: Its your call but I notice tons of `HostingEnvironmentKind.All` in the code. You can make this interface into a abstract class and give a default value to hostingEnvironmentKind.

puneetg1983: Hey @hforeste - this is a good suggestion. While trying to implement this, I realize that most of the places where SiteFilteredItem is used are inline typescript array. For e.g. this


```
this.diagnosticTools = [
      {
        appType: AppType.WebApp | AppType.FunctionApp,
        platform: OperatingSystem.windows,
        sku: Sku.NotDynamic,
        hostingEnvironmentKind: HostingEnvironmentKind.All,
        stack: 'ASP.NET',
        item: {
          id: ToolNames.Profiler,
          name: ToolNames.Profiler,
          category: 'Diagnostic Tools',
          description: '',
          featureType: FeatureTypes.Tool,
          clickAction: this._createFeatureAction(ToolNames.Profiler, 'Diagnostic Tools', () => {
            this._router.navigateByUrl(`resource${resourceId}/tools/profiler`);
          })
        }
      },

````

this.diagnosticTools is defined as

`public diagnosticTools: SiteFilteredItem<Feature>[];`

Even if I make it an abstract class, I still need to pass hostingEnvironmentKind here else Typescript is giving an error that it is not finding the property hostingEnvironmentKind even if I have a default value for it.

Any better way to handle this? Would want to avoid creating a constructor and change a whole lot of code to achieve this....

hforeste: I can't think of any other options that wouldn't require more coding than necessary. I think we can just merge what you have as is.

puneetg1983: cool - just wanted to double check. merging this now!

PR #299, opened 2019-04-08
Fix commit history bug

crazyxy:


ApplensBackend/Services/GithubClientService/GithubClientService.cs
```diff
@@ -142,8 +142,16 @@ public async Task<string> GetConfiguration(string id)
         /// <returns>Task for getting commit content.</returns>
         public async Task<string> GetCommitContent(string filePath, string sha)
         {
-            var commitContent = await OctokitClient.Repository.Content.GetAllContentsByRef(UserName, RepoName, filePath, sha);
-            return commitContent?[0].Content;
+            try
+            {
+                var commitContent = await OctokitClient.Repository.Content.GetAllContentsByRef(UserName, RepoName, filePath, sha);
+                return commitContent?[0].Content;
```
hforeste: Instead of doing this and possibly catching NullReferenceException to return empty string just check to see if its null and return empty string. Its kind of a nitpick but exceptions are more costly

PR #304, opened 2019-04-12
Configure Detector Link in Internal Solution Text

dagrooms52:
- Detector link can be placed anywhere in internal solution text
- If the format string {detectorlink} is not present, end the internal text with the default detector link markdown;
- Fixed detector link builder

### DIFF
--- a/AngularApp/projects/diagnostic-data/src/lib/components/solution/solution.component.ts
+++ b/AngularApp/projects/diagnostic-data/src/lib/components/solution/solution.component.ts
@@ -20,6 +20,7 @@ export class SolutionComponent extends DataRenderBaseComponent {
   defaultCopyText = 'Copy to Email';
   copyText = this.defaultCopyText;
   appName: string;
+  renderedInternalMarkdown = '';
 
   constructor(telemetryService: TelemetryService, private _siteService: SolutionService) {
     super(telemetryService);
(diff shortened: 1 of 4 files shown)


AngularApp/projects/diagnostic-data/src/lib/components/solution/solution.component.ts
```diff
@@ -29,16 +30,26 @@ export class SolutionComponent extends DataRenderBaseComponent {
...
+  buildInternalText() {
+    let markdownBuilder = this.solution.InternalMarkdown;
 
-    if (!this.solution.InternalMarkdown.includes(detectorLinkMarkdown)) {
-      this.solution.InternalMarkdown = this.solution.InternalMarkdown + "\n\n" + detectorLinkMarkdown;
+    let detectorLink = UriUtilities.BuildDetectorLink(this.solution.ResourceUri, this.solution.DetectorId);
+    let detectorLinkMarkdown = `Go to [App Service Diagnostics](${detectorLink})`;
+
+    if (markdownBuilder.toLowerCase().includes("{detectorlink}")) {
+        console.log(markdownBuilder);
```
cindylovescoding: Remove the console logging?

AngularApp/projects/diagnostic-data/src/lib/utilities/string-utilities.ts
```diff
@@ -0,0 +1,29 @@
+export class StringUtilities {
+    static TrimStart(target: string, trimSubstring: string): string {
+        let result = target;
```
cindylovescoding: Will *.length throw any exception if target or trimSubstring is undefined?

AngularApp/projects/diagnostic-data/src/lib/components/solution/solution.component.ts
```diff
@@ -29,16 +30,26 @@ export class SolutionComponent extends DataRenderBaseComponent {
...
+        markdownBuilder = markdownBuilder.replace(/{detectorlink}/gi, detectorLink);
```
cindylovescoding: Just curious, what does /gi stands for?

dagrooms52: It's the `g`lobal and case `i`nsensitive options for the regex string

cindylovescoding: I see. Thanks Daniel!

AngularApp/projects/diagnostic-data/src/lib/utilities/string-utilities.ts
dagrooms52: Yes, I guess I ignored that because I didn't make `trimSubstring` optional like in `TrimBoth`, but this made me realize I still need the check

PR #305, opened 2019-04-12
More Unhandled Exception Information

dagrooms52:
- Add stack trace to unhandled exception logs
- Add site's current route to unhandled exception logs

### DIFF
--- a/AngularApp/projects/diagnostic-data/src/lib/services/unhandled-exception-handler.service.ts
+++ b/AngularApp/projects/diagnostic-data/src/lib/services/unhandled-exception-handler.service.ts
@@ -11,7 +11,7 @@ export class UnhandledExceptionHandlerService {
 
   handleError(error: Error) {
     try {
-      this.logService.logException(error, null, null, null, SeverityLevel.Critical);
+      this.logService.logException(error, "unhandled", null, null, SeverityLevel.Critical);
     }
     catch(err) {
       // Squash logging error
--- a/AngularApp/projects/diagnostic-data/src/lib/services/unhandled-exception-handler.service.ts
+++ b/AngularApp/projects/diagnostic-data/src/lib/services/unhandled-exception-handler.service.ts
@@ -22,6 +22,10 @@ export class UnhandledExceptionHandlerService {
                 'route': this.router.url
             }
 
+            if (error.stack != undefined) {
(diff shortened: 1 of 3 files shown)


AngularApp/projects/diagnostic-data/src/lib/services/unhandled-exception-handler.service.ts
```diff
@@ -7,14 +8,28 @@ import { KustoTelemetryService } from './telemetry/kusto-telemetry.service';
...
+
+            const props = {
+                'route': this.router.url
+            }
+
+            if (error.stack != undefined) {
+                props['stack'] = error.stack;
+            }
+
+            this.logService.logException(error, "unhandled", props, null, SeverityLevel.Critical);
```
ShekharGupta1988: Can you put an example how the event looks like now?

AngularApp/projects/diagnostic-data/src/lib/services/unhandled-exception-handler.service.ts
```diff
@@ -7,14 +8,28 @@ import { KustoTelemetryService } from './telemetry/kusto-telemetry.service';
...
-      // Squash logging error
+    constructor(private logService: KustoTelemetryService, private injector: Injector) { }
+
+    handleError(error: Error) {
+        try {
+            if (this.router == undefined) {
+                this.router = this.injector.get(Router);
+            }
+
+            const props = {
```
cindylovescoding: Should we define this into const if you are going to add more key value pair below?

dagrooms52: Yes, I'm only adding keys not repointing `props` to a different object

AngularApp/projects/diagnostic-data/src/lib/services/unhandled-exception-handler.service.ts
dagrooms52: Sure, here's the exception object. You can see I added `route` and `stack` to the logs. I cut it off in the screenshot but the route goes to "{resourceUri}/categories/WindowsAvailabilityAndPerformance", and the stack is verbose although it only details internal Angular code in this case. I've tested that it does show stack traces for our code as well.

exception_logging

PR #311, opened 2019-04-17
 Fix for Issue : Premium Tools are not showing up

ShekharGupta1988:
It looks like PR #298 broke Premium tools feature.  Tinfoil Security tool is not showing up in portal anymore.

The fix is to have the new ASE filter added in Premium tools as well.


ShekharGupta1988:
> sorry, my bad

oh no worries at all. Its difficult to check each and every thing.

PR #313, opened 2019-04-18
Nmallick1/time range check bug

nmallick1:
Time range validation check was not being invoked when time duration was clicked. Also allowing 3 days for internal and 1 day for external clients.


nmallick1:
> This fix the error message not refreshed bug, but there is still something wrong with the internalView flag right?

Found the problem. It should be fixed in the latest update. Thanks for finding the bug @cindylovescoding

PR #322, opened 2019-05-09
Use diagnostic role for basic site stamp information

hforeste:
Today, when the end user requests for a site in applens, the backend makes a call to SupportBayApi to get basic site information. Instead of making that call to SupportBayApi we will make the call to diagnostic backend instead to get this site info. The same will go as for stamp/ase.

### DIFF
--- a/AngularApp/projects/applens/src/app/shared/services/ase.service.ts
+++ b/AngularApp/projects/applens/src/app/shared/services/ase.service.ts
@@ -17,15 +17,13 @@ export class AseService extends ResourceService {
   }
 
   public startInitializationObservable() {
-    this._initialized = this._observerApiService.getAse(this._armResource.resourceName).pipe(
-      mergeMap((observerResponse: Observer.ObserverAseResponse) => {
+    this._initialized = this._observerApiService.getAse(this._armResource.resourceName)
+      .pipe(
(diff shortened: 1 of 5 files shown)


AngularApp/projects/applens/src/app/shared/services/ase.service.ts
```diff
@@ -17,15 +17,13 @@ export class AseService extends ResourceService {
...
-    this._initialized = this._observerApiService.getAse(this._armResource.resourceName).pipe(
-      mergeMap((observerResponse: Observer.ObserverAseResponse) => {
+    this._initialized = this._observerApiService.getAse(this._armResource.resourceName)
+      .pipe(
+        map((observerResponse: Observer.ObserverAseResponse) => {
         this._hostingEnvironmentResource = observerResponse.details;
         this._currentResource.next(observerResponse.details);
-        return this._observerApiService.getAseRequestBody(this._hostingEnvironmentResource.Name);
-      }),map((requestBody: any) => {
-        this._requestBody = requestBody.details;
```
cindylovescoding: Hey Hawk, this requestbody is required by DiagRole backend to filter the right detectors we want for that type of resource, right? If we delete this requestbody, how are we going to return detectors say for Functions or Linux

PR #325, opened 2019-05-13
Rekha/enablement fixes

rekhaswaminathan:
- In enablement blade show in progress message when Resource Provider is registering. This fixes the issue where user has to do several clicks to save to enable Change Analysis Service
image

- When there is no change sets and no scans performed, show message to handle that
image
- Initiated by not populated
image

- Description is N/A in latest Change set.

### DIFF
--- a/AngularApp/projects/diagnostic-data/src/lib/utilities/changeanalysis-utilities.ts
+++ b/AngularApp/projects/diagnostic-data/src/lib/utilities/changeanalysis-utilities.ts
@@ -51,7 +51,8 @@ export class ChangeAnalysisUtilities {
         }
     }
 
-    public static findGroupBySource(source: any): number {
+    public static findGroupBySource(source: string): number {
+        source = source.toUpperCase();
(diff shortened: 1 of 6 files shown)


AngularApp/projects/diagnostic-data/src/lib/utilities/changeanalysis-utilities.ts
```diff
@@ -0,0 +1,66 @@
...
+        } catch(ex) {
+            return {
+                "code": diffvalue,
+                "language": 'text/plain'
+            }
+        }
+    }
+
+    public static findGroupBySource(source: any): number {
+        switch(source){
```
ShekharGupta1988: do you want to ignore case here as well by using source.ToUpperCase() just in case it changes in response

rekhaswaminathan: Sure, I can change that

PR #326, opened 2019-05-13
Fixing the time selector error for Edge

nmallick1:
Fixing the time selector for Edge. momentJS was falling back on creating date via js Date which wasn't working on Edge and hence was creating an invalid moment object causing the issue. Replaced .toString() with .format() which converts the date to one of the ISO date formats and conversion back from string to moment object always works in all browsers. Tested with Edge, Chrome and IE 11. Assuming it works with the rest of the browsers :-)

### DIFF
--- a/AngularApp/projects/diagnostic-data/src/lib/services/detector-control.service.ts
+++ b/AngularApp/projects/diagnostic-data/src/lib/services/detector-control.service.ts
@@ -212,23 +212,23 @@ export class DetectorControlService {
     }
   }
 
-  public selectDuration(duration: DurationSelector) {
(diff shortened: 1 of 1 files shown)


nmallick1:
> When does momentJs defer to using JS Dates?
momentJS will attempt to create a valid Date from the passed in string based on pattern matching. If none of of its mechanism works then it falls back to the use of javascript's Date function. This wasn't working on Edge based on the passed input but worked n Chrome. Moment library clearly states that this final fall back mechanism is unreliable on different browsers.

https://github.com/moment/moment/blob/96d0d6791ab495859d09a868803d31a55c917de1/moment.js#L2388

PR #335, opened 2019-05-18
Fix accessibility bugs 3723132, 3731213

michimune:
https://msazure.visualstudio.com/Antares/_workitems/edit/3723132
https://msazure.visualstudio.com/Antares/_workitems/edit/3731213


AngularApp/projects/app-service-diagnostics/src/app/availability/problem-solution/problem-solution.component.html
```diff
@@ -19,7 +19,7 @@
                         <tr>
                             <th>Observation</th>
                             <th>Description</th>
-                            <th></th>
+                            <th>Links</th>
```
rekhaswaminathan: nitpik: The `detector-list-analysis.component` has "Link" while this has "Links"

michimune: Ah yes, but this page will be replaced by Puneet's new analytics page. So it should be ok.

PR #340, opened 2019-05-20
Retry 3 times register Change Analysis

rekhaswaminathan:
- When registering resource provider if 404 is received, retry 3 times.
- Added more logging if error occurs

### DIFF
--- a/AngularApp/projects/app-service-diagnostics/src/app/home/components/diagnostics-settings/diagnostics-settings.component.ts
+++ b/AngularApp/projects/app-service-diagnostics/src/app/home/components/diagnostics-settings/diagnostics-settings.component.ts
@@ -7,6 +7,7 @@ import { Subscription, interval } from 'rxjs';
 import { ArmResource } from '../../../shared-v2/models/arm';
 import { ResourceService } from '../../../shared-v2/services/resource.service';
 import { PortalKustoTelemetryService} from '../../../shared/services/portal-kusto-telemetry.service';
+import { HttpResponse } from '@angular/common/http';
 
 @Component({
   selector: 'diagnostics-settings',
@@ -42,6 +43,7 @@ export class DiagnosticsSettingsComponent implements OnInit, OnDestroy {
   codeScanOption: any = {};
   featureRegOption: any = {};
   alwaysOnOption: any = {};
+  retryCount: number = 1;
   constructor(private armService: ArmService, private authService: AuthService,
      private activatedRoute: ActivatedRoute, private resourceService: ResourceService,
      private loggingService: PortalKustoTelemetryService) { }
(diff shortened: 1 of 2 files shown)


AngularApp/projects/app-service-diagnostics/src/app/home/components/diagnostics-settings/diagnostics-settings.component.ts
```diff
@@ -244,10 +256,24 @@ export class DiagnosticsSettingsComponent implements OnInit, OnDestroy {
             }
             this.updatingProvider = false;
         }, (error: any) => {
-            this.showGeneralError = true;
-            this.generalErrorMsg = 'Unable to register/unregister Change Analysis Resource Provider. Please try again later.';
-            this.updatingProvider = false;
-            this.featureRegOption = this.EnablementOptions[1];
+           this.logHTTPError(error, 'updateProviderRegister');
+            // When 404 error occurs when registering Resource Provider, retry
+            if(error.status === 404 && this.retryCount <= 3 && isRegister) {
```
ShekharGupta1988: In case of error, should we retry for all status codes and then eventually fail? 
I am thinking of a condition like http 429 (Too many requests). In that case also, you want to retry.

Now for server errors, (5xx class), we can decide whether to just fail without even retrying.

AngularApp/projects/app-service-diagnostics/src/app/home/components/diagnostics-settings/diagnostics-settings.component.ts
```diff
@@ -244,10 +256,24 @@ export class DiagnosticsSettingsComponent implements OnInit, OnDestroy {
...
+                this.updatingProvider = false;
+                this.showInProgress = true;
+                this.regState = 'Registering';
+                setTimeout( () => {
+                    this.loggingService.logTrace('Attempting retry to register resource provider, retry attempt '+this.retryCount);
```
ShekharGupta1988: Log the error code also

rekhaswaminathan: `this.logHTTPError(error, 'updateProviderRegister');` logs errorcode

AngularApp/projects/app-service-diagnostics/src/app/home/components/diagnostics-settings/diagnostics-settings.component.ts
rekhaswaminathan: I see. So Status code >=400 and <500 we do retry?

ShekharGupta1988: ya, thats better. I am thinking whether we should retry even for a 5xx like 503 (Service Unavailable) or no

rekhaswaminathan: let me remove the check for 404 status code so that we retry for any errors.

PR #353, opened 2019-05-23
Update Aria-Label With Search Result Count

dagrooms52:
Updated aria-label notifies user of search result count


dagrooms52:
This actually needed to have `<div role="status">` to announce the result count - noting for future implementations

PR #360, opened 2019-05-27
Update logging for ChildDetectors in Analysis and fixing accessibility bug - 3737201

puneetg1983:


AngularApp/projects/app-service-diagnostics/src/app/shared/components/collapsible-list/collapsible-list.component.scss
```diff
@@ -38,4 +37,7 @@
 }
 .list-item.disabled:hover{
     text-decoration: none;
+}
+.list-item-title{
+    padding-left:10px;
```
dagrooms52: I added bootstrap 4 spacing helpers, so `class="pl-3"` would apply similar padding without needing to create a class

AngularApp/projects/diagnostic-data/src/lib/components/detector-list-analysis/detector-list-analysis.component.ts
```diff
@@ -85,6 +85,15 @@ export class DetectorListAnalysisComponent extends DataRenderBaseComponent imple
           if (this.detectorId !== "") {
             let currentDetector = detectorList.find(detector => detector.id == this.detectorId)
             this.detectorName = currentDetector.name;
+            return;
```
dagrooms52: The original version let this method continue, is this supposed to exit without processing the `detectorList`?

puneetg1983: Yes, when we have a detector in the URL, we need not process all the child detectors

AngularApp/projects/diagnostic-data/src/lib/components/detector-list-analysis/detector-list-analysis.component.ts
```diff
@@ -85,6 +85,15 @@ export class DetectorListAnalysisComponent extends DataRenderBaseComponent imple
...
+          } else {
+            this.detectorEventProperties = {
+              'StartTime': String(this._detectorControl.startTime),
+              'EndTime': String(this._detectorControl.endTime),
+              'DetectorId': this.analysisId,
+              'ParentDetectorId': "",
+              'Url': window.location.href
```
cindylovescoding: Can we also add the detectorType to tell if it's analysis or detector?

puneetg1983: We can add it. Does it have any benefit if we add or don't add it?

puneetg1983: not opposing the ask, just trying to understand how it helps :-)

cindylovescoding: Just want to make sure we don't mix the usage analysis for some of the detector type and analysis type with the same name. But it is fine for now.

PR #361, opened 2019-05-28
 Fix bug 3723714

rekhaswaminathan:
- Fix issue where screen reader is reading it as not group.

- Better aria-label for detector heading div


AngularApp/projects/diagnostic-data/src/lib/components/changesets-view/changesets-view.component.ts
```diff
@@ -159,7 +159,7 @@ export class ChangesetsViewComponent extends DataRenderBaseComponent implements
             this.loadingChangesTable = false;
             this.changesTableError = '';
         } else {
-            let queryParams = `&changeSetId=${encodeURI(changeSetId)}`;
+            let queryParams = `&changeSetId=${encodeURIComponent(changeSetId)}`;
```
cindylovescoding: What's the difference between these two?

rekhaswaminathan: Actually this change was originally part of a different branch, I made that change to encode the query string parameter. The one used before was supposed to be used to encode the full URI

PR #364, opened 2019-05-29
Width settings for publish modal only

SharmaAjay19:


AngularApp/projects/applens/src/styles.scss
```diff
@@ -151,7 +151,7 @@ body {
     background-color: #f0f0f0
 }
 
-.nsm-dialog-open {
-  min-width: 70vw;
-  min-height: 70vh;
+.publish-dialog{
```
cindylovescoding: Could you help change all  the modal styles to be previous styles?
We are seeing all the modals are wide modals now

SharmaAjay19: That's exactly what this change is for! :P

PR #365, opened 2019-05-30
Rekha/settings ui changes

rekhaswaminathan:
Made improvements to current enablement UI:

- Change Analysis Option Off by Default
- Tool tip to show what each setting means. 
- Toggle on/off button to make it more obvious
- Renamed variables to make more sense
- Removed banner above Always on option as it can be missed.

Current UI - 
image

New UI - 

image

Toggle button disabled state while feature reg in progress:

image


yunjchoi:
@rekhaswaminathan 
The UI looks more organized now. I think using the tool tip over the banner was the right call. However, I do prefer the original radio button over the new toggle. I think the original radio button does a better job at showing which setting is selected. Can we change the color of the radio button when it's selected like the below image so it would help highlight the selected setting?
  
image

rekhaswaminathan:
Chatted with Yun Jung offline and we agreed that toggle switch would be a better approach here as this setting option is mutually exclusive. 

https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/radio-button

https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/toggles

Open to feedback from others as well.

pharring:
@CawaMS @mrbullwinkle to update screenshots in https://docs.microsoft.com/en-us/azure/azure-monitor/app/change-analysis when this goes live

ShekharGupta1988:
@rekhaswaminathan  - quick clarification : what is the first view that the customer is presented with? will it have all three options as OFF , or just the first Change Analysis ON/OFF (Default to OFF). Once I toggle that ON, the other two options appear, as sub-options  (by default both ON)

rekhaswaminathan:
@ShekharGupta1988  In the first view after the Feature reg is completed, Change Analysis & Always on is defaulted to OFF. Once they toggle Change Analysis to ON, code scan will be defaulted to ON. Always on is not coded to be a sub-option. Only code scan is a sub-option

rekhaswaminathan:
Updated text:
image

ShekharGupta1988:
> nce they toggle Change Analysis to ON

Got it. Should we even show those 2 options if Change Analysis is OFF.  We cant really do anything with the

AngularApp/projects/app-service-diagnostics/src/app/home/components/diagnostics-settings/diagnostics-settings.component.html
```diff
@@ -16,48 +16,37 @@
...
-                  <fieldset [disabled]="!isFeatureRegistered">
-                    <ng-container *ngFor="let option of EnablementOptions">
-                      <label class="radio-inline control-label">
-                        <input type="radio" name="changAnalysis" [checked]="featureRegOption.option == option.option" value="option.option" (change)="setChangeAnalysisEnabled(option)"> {{option.option}}
-                      </label>
-                    </ng-container>
-                  </fieldset>
+                <label class="control-label col-sm-2 settings-label">Change Analysis</label>
+                <div class="col-sm-2">
+                  <toggle-button [isDisabled]="!isFeatureRegistered" [selected]="providerRegOption.option == 'On'" (selectedChange)="$event ? providerRegOption = EnablementOptions[0] : providerRegOption = EnablementOptions[1]"></toggle-button>
```
cindylovescoding: We try to avoid using double equal but more tripple equals  providerRegOption.option use === 'On'  for strict comparison :https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness

AngularApp/projects/app-service-diagnostics/src/app/home/components/diagnostics-settings/diagnostics-settings.component.ts
```diff
@@ -276,9 +266,9 @@ export class DiagnosticsSettingsComponent implements OnInit, OnDestroy {
             let providerRegistrationStateResponse = <ProviderRegistration>response.body;
             let state = providerRegistrationStateResponse.registrationState;
             if (state.toLowerCase() == 'registered') {
```
cindylovescoding: Same for these places:)

AngularApp/projects/app-service-diagnostics/src/app/home/components/diagnostics-settings/diagnostics-settings.component.html
rekhaswaminathan: Thanks Cindy, I will make the change

PR #368, opened 2019-06-03
Fix ARM resource uri login

cindylovescoding:
1. Arm resource id login fix
2. Add assets for app service cert and domain resource type
3. Some style fix


rekhaswaminathan:
Does this change UI? If so can you add some screenshots?

AngularApp/projects/applens/src/styles.scss
```diff
@@ -76,6 +76,7 @@ body {
 .wide-modal {
   max-width: 560px!important;
   padding: 0px!important;
+  min-height: 160px!important;
```
dagrooms52: `!important` is a separate command so you can have space between the value and `!important`.
However, we should also try to avoid it if we can increase specificity per use-case of the class instead. Will this work without `!important` so that we can configure it differently per usage?

cindylovescoding: Yeah unfortunately this won't work without !important in this case. Will avoid using that in our project.

cindylovescoding:
@rekhaswaminathan  Hey Rekha, just small changes for the detector name border inside of those tiles:
image

Also added image for app service cert and domain:
image
image

PR #374, opened 2019-06-04
Fixing time calculation for Daas Sessions table

puneetg1983:


AngularApp/projects/app-service-diagnostics/src/app/shared/components/daas-sessions/daas-sessions.component.html
```diff
@@ -71,7 +71,7 @@
                         <i class="fa health-icon fa-hourglass-half warning-icon-color" aria-hidden="true"></i>
                     </span>
                     <b>{{ session.StartTime }} UTC</b>
-                    <span>- {{ session.StartTime | datetimediff }} ago</span>
+                    <span>- {{ session.StartTime + 'Z' | datetimediff }} ago</span>
```
hforeste: nitpik; Adding this in the html seems a little hacky, is it more appropriate to append the string in the business logic? Or even better should DaaS be updated to respond with timestamp in UTC with the appended 'Z'

puneetg1983: yup, will fix this from backend shortly. I haven't changed this in ages, not sure how this was working earlier :-D

PR #376, opened 2019-06-05
Wrap word formatting for resource name

cindylovescoding:
Wrap the resource name and limit the max width, to make sure it doesn't overlap with the informational icon.
image

### DIFF
--- a/AngularApp/projects/applens/src/app/modules/dashboard/dashboard/dashboard.component.scss
+++ b/AngularApp/projects/applens/src/app/modules/dashboard/dashboard/dashboard.component.scss
@@ -85,6 +85,7 @@ side-nav {
     padding: 2px 5px 5px 15px;
     display: inline-block;
     max-width: 180px;
+    word-wrap: break-word;
 }
 
 .user-panel {
--- a/AngularApp/projects/applens/src/app/modules/dashboard/dashboard/dashboard.component.scss
+++ b/AngularApp/projects/applens/src/app/modules/dashboard/dashboard/dashboard.component.scss
@@ -86,6 +86,7 @@ side-nav {
     display: inline-block;
     max-width: 180px;
     word-wrap: break-word;
+    top: 9px;
 }
 
 .user-panel {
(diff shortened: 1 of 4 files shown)


AngularApp/projects/applens/src/app/modules/dashboard/dashboard/dashboard.component.scss
```diff
@@ -85,6 +85,7 @@ side-nav {
     padding: 2px 5px 5px 15px;
     display: inline-block;
     max-width: 180px;
+    word-wrap: break-word;
```
ShekharGupta1988: what happens if the name of the resource is very big? will it result in 3 lines leading to increased height if the top blue bar?

cindylovescoding: No Blue bar will not increased height. The resource name will get hided a little though. 
Shekhar do you think we should truncate the resource name if it is too long?
The full resource info will still be there if people click the info icon to get resource info summary.
image
image

ShekharGupta1988: ya.. truncation with three dots would be a nice experience in this case

cindylovescoding: Sounds good! Added the truncate logic and now this is how it looks like:
image

AngularApp/projects/applens/src/app/modules/dashboard/dashboard/dashboard.component.ts
```diff
@@ -29,6 +29,7 @@ export class DashboardComponent implements OnDestroy {
 
   currentRoutePath: string[];
   resource: any;
+  resourceName: string="";
```
ShekharGupta1988: Alternatively, you can create a angular Pipe for truncation and use it in the Html directly.

If you are going with this, lets call it 'displayedResourceName' instead of 'resourceName' in order to avoid confusion in future. If someone tries to use this name to make any api calls, it might fail for long names.

PR #386, opened 2019-06-14
Accessibility Fixes

hforeste:
This fixes the focus order bug in https://msazure.visualstudio.com/Antares/_workitems/edit/3723691


hforeste:
Add fixes for https://msazure.visualstudio.com/Antares/_workitems/edit/3717257

PR #389, opened 2019-06-16
Telemetry enhancements for applens UI

SharmaAjay19:


ShekharGupta1988:
@SharmaAjay19 - This is great. Can you add one more column to the  HomePageLoaded event which is the logged-in user id.

AngularApp/projects/applens/src/app/modules/dashboard/resource-home/resource-home.component.ts
```diff
@@ -112,6 +115,9 @@ export class ResourceHomeComponent implements OnInit {
 
 
             });
+            let alias = this._adalService.userInfo.profile ? this._adalService.userInfo.profile.upn : '';
+            let userId = alias.replace('@microsoft.com', '').toLowerCase();
+            this._telemetryService.logPageView(TelemetryEventNames.HomePageLoaded, {"numCategories": this.categories.length.toString(), "userId": userId});
```
rekhaswaminathan: It looks like userid is getting logged, Isn't that PII? Can you give more context on why its needed?

SharmaAjay19: This is for applens users only which are all internal users (mostly CSS people). So we will not be logging any customer PII. It is needed to do user behavior analysis (Across sessions).

PR #390, opened 2019-06-17
Reduce the 3 buttons of change analysis onboarding page to 1 button

yantang-msft:


AngularApp/projects/app-service-diagnostics/src/app/home/components/diagnostics-settings/diagnostics-settings.component.ts
```diff
@@ -1,362 +1,207 @@
...
-   ngOnDestroy(): void {
-        if (this.subscription) {
-            this.subscription.unsubscribe();
+    private _unregisterSubscription(subscription: Subscription): void {
+        if (subscription) {
+            subscription.unsubscribe();
+        }
+    }
+
+    private _getGeneralErrorMsg(baseMsg: string, errorStatus: number) {
```
rekhaswaminathan: Nit: We dont generally use _ starting for method names.

AngularApp/projects/app-service-diagnostics/src/app/home/components/diagnostics-settings/diagnostics-settings.component.ts
```diff
@@ -1,362 +1,207 @@
...
-                    this.generalErrorMsg = 'Unable to register/unregister Change Analysis Resource Provider. Please try again later.';
-                }
-                this.updatingProvider = false;
-                this.showInProgress = false;
-                this.providerRegOption = this.EnablementOptions[1];
-            }
+            this.logHTTPError(error, 'updateScanTag');
+            this.updatingTag = false;
+            this.showGeneralError = true;
+            this.generalErrorMsg = this._getGeneralErrorMsg('Unable to add scan tag. ', error.status);
```
rekhaswaminathan: This message gets displayed in the UI, right? So we should change the wording to something like "Error occurred when trying to enable  Change Analysis"

AngularApp/projects/app-service-diagnostics/src/app/home/components/diagnostics-settings/diagnostics-settings.component.ts
```diff
@@ -1,362 +1,207 @@
...
-            this.generalErrorMsg = 'Unable to check Change Anaylsis feature status. ' +this.tokenErrorMsg;
-        } else {
-            this.generalErrorMsg = 'Unable to check Change Anaylsis feature status. Please try again later.';
-        }
-        if(this.subscription) {
-            this.subscription.unsubscribe();
-        }
-    });
-   }
+        let url = `/subscriptions/${this._subscriptionId}/providers/Microsoft.ChangeAnalysis/register`;
```
rekhaswaminathan: can you use the provider reg url component property here?

AngularApp/projects/app-service-diagnostics/src/app/home/components/diagnostics-settings/diagnostics-settings.component.ts
```diff
@@ -1,362 +1,207 @@
...
-       }, (error: any) => {
-            this.logHTTPError(error, 'checkIfProviderRegistered');
-            this.showGeneralError = true;
-            if(error.status === 403) {
-                this.generalErrorMsg = 'Unable to check Change Analysis Resource Provider status. ' + this.permissionErrorMsg;
-            } else if (error.status === 401) {
-                this.generalErrorMsg = 'Unable to check Change Anaylsis Resource Provider status. ' + this.tokenErrorMsg;
+    private _pollResourceProviderRegStatus(): void {
+        this.pollingResourceProviderRegProgress = true;
+        this._pollResourceProviderStatusSubscription = timer(0, 5000).subscribe(_ => {
```
rekhaswaminathan: what is the difference between `timer(0,5000)` and `interval(5000).subscribe` ?

AngularApp/projects/app-service-diagnostics/src/app/home/components/diagnostics-settings/diagnostics-settings.component.ts
```diff
@@ -1,362 +1,207 @@
...
-    this.subscriptionId = this.activatedRoute.snapshot.params['subscriptionid'];
-    this.authService.getStartupInfo().subscribe(data => {
-        this.resourceId = data.resourceId;
-    });
-    this.featureRegUrl = `/subscriptions/${this.subscriptionId}/providers/Microsoft.Features/providers/Microsoft.ChangeAnalysis/features/PreviewAccess`;
-    this.providerRegUrl = `/subscriptions/${this.subscriptionId}/providers/Microsoft.ChangeAnalysis`;
-    this.currentResource = this.resourceService.resource;
-    this.checkIfFeatureRegister();
-  }
+        this._isRPRegistered.subscribe(_ => this._updateChangeAnalysisEnableStatus());
```
rekhaswaminathan: Can you remove the starting _ ?

rekhaswaminathan:
Have you tested how this new UI experience looks like with the current detector flow ? i.e Currently we are not consuming the API which tells us if web app is on-boarded. We are checking if the subscription is on boarded. So we should test to make sure experience is not broken as UI deployment and backend deployment are not always in sync

yantang-msft:
@rekhaswaminathan Yes, I tested it and I believe I found a minor bug regarding how detector check if subscription is registered with RP. Basically, while the subscription is still in "Registering" status, the detector will treat the subscription as onboarded and show the ARG result.
Other than that, the workflow is doing what's expected (and I really feel the need of the backend change so we promote the onboarding page when the hidden tag is not added).

AngularApp/projects/app-service-diagnostics/src/app/home/components/diagnostics-settings/diagnostics-settings.component.ts
yantang-msft: Sure

rekhaswaminathan:
@yantang-msft When the subscription is still in "Registering" state, what response does the /onboarding endpoint in the dataplane return? The detector is basically consuming that. 

> and I really feel the need of the backend change so we promote the onboarding page when the hidden tag is not added

- Yes we are going to make this change as soon as our deployment completes in all 4 public georegions. If we make it right now, it will work partially and fail in other regions.

yantang-msft:
> what response does the /onboarding endpoint in the dataplane return

Oh right, the detector is consuming the response from ChangeAnalysis dataplane but not ARM, and the dataplane returns "Registered" once we have a record in the storage account. I think this is just a minor problem.
While FYI there is a real problem on our end and we are investigating is even if ARM says the RP is registered, it still takes 20mins before our RP can actually access user's data (and sometimes takes hours).
So if the user want to try ChangeAnalysis right after they are onboarded, they will get an 403 forbidden error (and we saw couple cases already).

rekhaswaminathan:
@yantang-msft Got it. Thanks

PR #392, opened 2019-06-18
Change the home page health check component to point to AppLens v3

puneetg1983:
With this PR, adding a new component HealthCheckV3Component which we can enable whenever we want to point to the AppLens v3 detectors. 

I am not currently enabling it, just getting the code reviewed and merged.

This is how it will look with v3

image


puneetg1983:
Hey Shekhar, do you think we should just enable it? I was talking to Praveen and he was saying we can go with it... what do you think?

puneetg1983:
@ShekharGupta1988 , @cindylovescoding  - Can you review the updates to the Insight component and just validate that I am not messing up something. The insights are used in a lot of places so would love to get a quick validation from you both :)

PR #393, opened 2019-06-18
Change Analysis onboarding UI tweaks

yantang-msft:
1. Tweak the enablement button group so it works better on narrow sized screen
2. Update the last scan message so we don't confuse user with the concept of code scan
3. Update the last scan message so if code scan is enabled and there is no last scan yet, we don't ask user to enable Change Analysis

### DIFF
--- a/AngularApp/projects/diagnostic-data/src/lib/components/changesets-view/changesets-view.component.ts
+++ b/AngularApp/projects/diagnostic-data/src/lib/components/changesets-view/changesets-view.component.ts
@@ -65,12 +65,12 @@ export class ChangesetsViewComponent extends DataRenderBaseComponent implements 
                 this.initializeChangesView(data);
             }
             // Convert UTC timestamp to user readable date
-            this.scanDate = rows[0][6] != '' ? 'Changes were last scanned on ' + moment(rows[0][6]).format("ddd, MMM D YYYY, h:mm:ss a") : 'No recent scans were performed on this web app. Make sure scan for code changes has been enabled in settings.';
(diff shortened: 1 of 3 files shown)


AngularApp/projects/diagnostic-data/src/lib/components/changesets-view/changesets-view.component.ts
```diff
@@ -221,48 +224,52 @@ export class ChangesetsViewComponent extends DataRenderBaseComponent implements
...
-                        this.setScanState(submissionState);
-                        let completedTime = rows[0][3];
-                        let currentMoment = moment();
-                        let completedMoment = moment(completedTime);
-                        let diff = currentMoment.diff(completedMoment, 'seconds');
-                        // If scan has been completed more than a minute ago, display default message
-                        if (diff >= 60) {
+            if (isEnabled) {
+                if (this.scanDate.startsWith(this.noScanMsg)) {
+                    this.scanDate = this.noScanMsg;
```
rekhaswaminathan: why do we need to reset `this.scanDate = this.noScanMsg`?

yantang-msft: For a newly enabled resource without last scan, we original set it to something like "No recent scan, please enable change analysis".
Then portal is able to detect the hidden tag is actually set, in this case, update the message so we don't ask user to "enable change analysis" since it's already enabled.
This is  just a minor tweak and we won't have this problem after the detector checks the hidden tag.

PR #398, opened 2019-06-21
Fixed case where initiatedByList is empty

rekhaswaminathan:


AngularApp/projects/diagnostic-data/src/lib/components/changesets-view/changesets-view.component.ts
```diff
@@ -462,7 +462,7 @@ export class ChangesetsViewComponent extends DataRenderBaseComponent implements
         let initiatedByListIndex = this.getColumnIndexByName('initiatedByList');
         let initiatedByIndex = this.getColumnIndexByName('initiatedBy');
         if(initiatedByListIndex >= 0) {
-            return changeSet[initiatedByListIndex];
+            return changeSet[initiatedByListIndex].length > 0 ? changeSet[initiatedByListIndex] : ['Unable to Determine'];
```
ShekharGupta1988: should we also check for NULL on `changeSet[initiatedByListIndex]`

rekhaswaminathan: Good point, Let me add that also

rekhaswaminathan: Actually null check wont be required here because its coming as array and we are checking if Index is also valid.

PR #402, opened 2019-06-25
Allowing to order tiles for categories and accessibility fixes for Analysis

puneetg1983:
With this change, any resource type can choose the order of the tiles for various categories by modifying _featureDisplayOrder array inside FeatureService


AngularApp/projects/app-service-diagnostics/src/app/shared-v2/services/feature.service.ts
```diff
@@ -64,6 +76,49 @@ export class FeatureService {
...
+          }
+        });
+
+        // Remove all the features for the sorted category
+        this._features = this._features.filter(x => {
+          return x.category !== feature.category;
+        });
+
+        // Sort all the features for this category
+        categoryFeatures.sort(
```
ShekharGupta1988: we can prob. achieve this by creating a Angular Pipe?

PR #404, opened 2019-06-25
Add url in event logs and page views

SharmaAjay19:


AngularApp/projects/diagnostic-data/src/lib/services/telemetry/telemetry.service.ts
```diff
@@ -43,6 +43,9 @@ export class TelemetryService {
                 }
             }
         }
+        if (!(properties["url"] || properties["Url"])){
```
nmallick1: Do you think it can be properties["URL" ] too ?

PR #407, opened 2019-06-26
Fix Diag & Solve loading for ASE

rekhaswaminathan:
- Removed injecting Websites service in feature service.
- Moved display order logic to just site-feature service.


puneetg1983:
Thanks @rekhaswaminathan  for taking care of this.

PR #411, opened 2019-06-28
Display search only if enabled for product Id

SharmaAjay19:


AngularApp/projects/applens/src/app/modules/dashboard/services/search.service.ts
```diff
@@ -16,11 +18,15 @@ export class SearchService {
...
         let userId = alias.replace('@microsoft.com', '').toLowerCase();
-        this._applensDiagnosticService.getHasTestersAccess(userId).subscribe(res => {
-            this.searchIsEnabled = res;
-        },
-        (err) => {
-            this.searchIsEnabled = false;
+        let hasTestersAccess = this._applensDiagnosticService.getHasTestersAccess(userId).pipe(map((res) => res), catchError(e => of(false)));
+        let isEnabledForProductId = this._applensDiagnosticService.getSearchEnabledForProductId().pipe(map((res) => res), catchError(e => of(false)));
+        forkJoin([hasTestersAccess, isEnabledForProductId]).subscribe(enabledFlags => {
+            if (enabledFlags[0] && enabledFlags[1]){
```
rekhaswaminathan: what about error case here? Also should we validating `enabledFlags` before accessing by index?

SharmaAjay19: Error case is handled as feature being disabled. enabledFlags is always an array of two values since we catch any errors and default to false.

rekhaswaminathan: Where is it handled? Applens backend project?

AngularApp/projects/applens/src/app/modules/dashboard/services/search.service.ts
```diff
@@ -16,11 +18,15 @@ export class SearchService {
...
+                this.searchIsEnabled = true;
+            }
+            else {
+                this.searchIsEnabled = false;
+            }
```
cindylovescoding: This if else clause can be replaced by this single line right?
this.searchIsEnabled = enabledFlags[0] && enabledFlags[1];

AngularApp/projects/applens/src/app/modules/dashboard/services/search.service.ts
SharmaAjay19: In the catchError function above.

ApplensBackend/Controllers/GithubController.cs
```diff
@@ -45,6 +45,31 @@ public async Task<IActionResult> GetTemplate(string name)
...
+                var resConf = await GithubService.GetResourceConfigFile();
+                if (resConf.Contains(productId)){
+                    return Ok(true);
+                }
+                else{
+                    return Ok(false);
+                }
+            }
+            catch {
+                return Ok(false);
```
cindylovescoding: How can we tell whether there is an exception or search is simply not enabled for the product id?

AngularApp/projects/applens/src/app/modules/dashboard/services/search.service.ts
rekhaswaminathan: Oh ok got it. Thanks!

PR #416, opened 2019-07-05
Adding SupportTopic information in Chat Backend

puneetg1983:
Hi @ShekharGupta1988 , filing this PR to get a quick sense on whether I am on the right track or not. Can you please review and let me know if this direction makes sense? Also, I wasn't able to figure out if chat is enabled currently in Diagnose and Solve or not and if any of these changes will make a difference to that flow. Based on your feedback, will test it fully on Monday and update the PR accordingly

### DIFF
--- a/Backend/Controllers/ChatController.cs
+++ b/Backend/Controllers/ChatController.cs
@@ -23,7 +23,7 @@ namespace Backend.Controllers
         [HttpGet("status")]
         public IActionResult GetChatStatus(string product, string supportTopic)
         {
-            if (string.IsNullOrWhiteSpace(product) || string.IsNullOrWhiteSpace(product))
+            if (string.IsNullOrWhiteSpace(product) || string.IsNullOrWhiteSpace(supportTopic))
             {
                 return BadRequest("product and supportTopic cannot be empty");
             }
--- a/Backend/Controllers/ChatController.cs
+++ b/Backend/Controllers/ChatController.cs
@@ -10,7 +10,7 @@ using Backend.Services;
 namespace Backend.Controllers
 {
     [Produces("application/json")]
-    [Route("api/chat/{product}")]
+    [Route("api/chat/{product}/{supportTopic}")]
     [ResponseCache(CacheProfileName = "Default")]
     public class ChatController : Controller
     {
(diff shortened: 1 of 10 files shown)


Backend/Controllers/ChatController.cs
```diff
@@ -21,14 +21,14 @@ public ChatController(IChatService chatService)
         }
 
         [HttpGet("status")]
-        public IActionResult GetChatStatus(string product)
+        public IActionResult GetChatStatus(string product, [FromHeader(Name = "supportTopic")] string supportTopic = "")
```
ShekharGupta1988: Should the supportTopic be a query string instead of a header? Just helps testing it a bit easier

Backend/Controllers/ChatController.cs
```diff
@@ -21,14 +21,14 @@ public ChatController(IChatService chatService)
...
         {
             if (string.IsNullOrWhiteSpace(product))
             {
                 return BadRequest("product cannot be empty");
             }
 
-            return Ok(ChatService.GetChatStatus(product));
```
ShekharGupta1988: What is the behavior if support topic is not passed or empty? Should it be a bad Request?

puneetg1983: Yes, Changed it.

Backend/Controllers/ChatController.cs
puneetg1983: done.

Backend/Controllers/ChatController.cs
```diff
@@ -21,14 +21,14 @@ public ChatController(IChatService chatService)
         }
 
         [HttpGet("status")]
-        public IActionResult GetChatStatus(string product)
+        public IActionResult GetChatStatus(string product, string supportTopic)
         {
-            if (string.IsNullOrWhiteSpace(product))
+            if (string.IsNullOrWhiteSpace(product) || string.IsNullOrWhiteSpace(product))
```
ShekharGupta1988: this should be supportTopic :)

Backend/Controllers/ChatController.cs
```diff
@@ -10,7 +10,7 @@
 namespace Backend.Controllers
 {
     [Produces("application/json")]
-    [Route("api/chat/{product}")]
+    [Route("api/chat/{product}/{supportTopic}")]
```
ShekharGupta1988: this approach might backfire when we plan to enable chat at diagnose and solve where there is no support topic. 
You can revisit this later if its blocking the chat enablement in India

Backend/Controllers/ChatController.cs
puneetg1983: ouch - thanks for catching this...Fixed

Backend/Controllers/ChatController.cs
puneetg1983: Let's unblock Khaled for now and I will try to sync with you once to understand the right thing to do.

PR #429, opened 2019-07-12
Securing applens using Azure AD Security groups

SharmaAjay19:
- Securing applens backend (all controllers except Fresh chat controller)
- Securing applens frontend with a ping call to check user access in auth guard.
- Display unauthorized message to user and steps to get access
image


AngularApp/projects/applens/src/app/shared/auth/aad-auth-guard.service.ts
```diff
@@ -26,7 +30,27 @@ export class AadAuthGuard implements CanActivate {
...
+                    if (err.status == 401)
+                    {
+                        this.isAuthorized = false;
+                        this._router.navigate(['unauthorized']);
+                        return Observable.throw(false);
+                    }
+                    else {
+                        this.isAuthorized = false;
+                        return Observable.throw(true);
+                    }
```
rekhaswaminathan: does this mean that for other error codes such as 403 we activate it? shouldn't we return false for any error condition?

ApplensBackend/Services/Authorization/AuthorizationHandler.cs
```diff
@@ -0,0 +1,239 @@
...
+                loggedInUsersCache.Add(securityGroup.GroupId, new Dictionary<string, long>());
+            }
+
+            StartTokenRefresh();
+            ClearLoggedInUserCache();
+            _httpContextAccessor = httpContextAccessor;
+        }
+
+        private async Task StartTokenRefresh()
+        {
```
rekhaswaminathan: I believe this logic is repeated in `GraphTokenService.cs` also. We can consider abstracting the logic like how RuntimeHost does.

ApplensBackend/appsettings.json
```diff
@@ -75,5 +75,18 @@
     "FreshChatBaseUri": "https://api.freshchat.com/v2",
     "UserAgentToFreshChat": "applensv3"
   },
-  "TestSecurityGroup": "ajsharm,xipeng,cmaher,dagrooms,hawfor,khzayed,mikono,micjohn,nmallick,pbabut,puneetg,reswamin,shgup"
+  "SecurityGroups": [
+    {
+      "GroupName": "ApplensAccess",
+      "GroupId": ""
```
rekhaswaminathan: Is `GroupId` value a secret?

AngularApp/projects/applens/src/app/shared/auth/aad-auth-guard.service.ts
ShekharGupta1988: Good point.. Looking at the code, the authorization would still be false.. However, I am not sure if the unauth page would be shown.

ApplensBackend/Controllers/TestersAccessController.cs
```diff
@@ -0,0 +1,24 @@
...
+    [Route("api/hasTestersAccess")]
+    [Authorize(Policy = "ApplensAccess")]
+    [Authorize(Policy = "ApplensTesters")]
+    public class TestersAccessController : Controller
+    {
+        /// <summary>
+        /// Method to check testers access.
+        /// </summary>
+        /// <returns>True result.</returns>
+        [HttpGet("")]
```
cindylovescoding: We can remove the empty string right? like [HttpGet] or [HttpPost] here?

AngularApp/projects/applens/src/app/shared/auth/aad-auth-guard.service.ts
SharmaAjay19: For unauthorized we will always return 401 from backend. For other errors we will build a common error alerting piece that can be reused across the app. Will work on that.

ApplensBackend/Controllers/TestersAccessController.cs
SharmaAjay19: My bad. I was putting a route there earlier and later forgot to remove the empty string.

ApplensBackend/Services/Authorization/AuthorizationHandler.cs
SharmaAjay19: I started with that thinking and concluded that wouldn't be a healthy approach since we are using two different Azure Ad apps for this, and hence two different authentication contexts. Making it a common method would mean passing the authentication context and other locally global variables to the method (which are large in number just to complicate the method signature) and use of out parameters to get the context updated. Since this is a async task, use of out/ref params isn't supported.

ApplensBackend/Services/Authorization/AuthorizationHandler.cs
```diff
@@ -0,0 +1,239 @@
...
+                var token = new JwtSecurityToken(accessToken);
+                object upn;
+                if (token.Payload.TryGetValue("upn", out upn)){
+                    userId = upn.ToString();
+                    if (userId != null){
+                        if (IsUserInCache(requirement.SecurityGroupObjectId, userId)){
+                            isMember = true;
+                        }
+                        else{
+                            isMember = CheckSecurityGroupMembership(userId, requirement.SecurityGroupObjectId).GetAwaiter().GetResult();
```
hforeste: Instead of `GetAwaiter().GetResult` use `await CheckSecurityGroupMembership(..,..,,)`

ApplensBackend/Services/Authorization/AuthorizationHandler.cs
rekhaswaminathan: You can have the method that does the token acquisition in the base class and override the authentication contexts in the derived classes. We can discuss offline if you want more context.

ApplensBackend/Services/Authorization/AuthorizationHandler.cs
```diff
@@ -0,0 +1,239 @@
...
+        }
+
+        protected override Task HandleRequirementAsync(AuthorizationHandlerContext context, SecurityGroupRequirement requirement) {
+            HttpContext httpContext = _httpContextAccessor.HttpContext;
+            Boolean isMember = false;
+            string userId = null;
+            try
+            {
+                string authorization = httpContext.Request.Headers["Authorization"].ToString();
+                string accessToken = authorization.Split(" ")[1];
```
hforeste: Check if `authorization` is null. In .NET Core even though the authorization header could be missing the authentication middleware will not shortcircuit the request, so that means the request can still make it to the authorization middleware that you created thus the authorization header missing could result in null.

ApplensBackend/appsettings.json
SharmaAjay19: Yeah it is the object id of a security group.

PR #431, opened 2019-07-12
Accessibility - Media query to fix issues for high contrast in graphs

SharmaAjay19:
Bug link - https://msazure.visualstudio.com/Antares/_workitems/edit/3716845
Normal Mode
image

High contrast mode - white on black
image

High contrast mode - black on white
image

### DIFF
--- a/AngularApp/projects/app-service-diagnostics/src/styles/global.scss
+++ b/AngularApp/projects/app-service-diagnostics/src/styles/global.scss
@@ -120,6 +120,23 @@ body{
     }
 }
 
+@media screen and (-ms-high-contrast: black-on-white) {
+    .nvd3 text {
+        fill: #000 !important;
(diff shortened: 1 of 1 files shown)


AngularApp/projects/app-service-diagnostics/src/styles/global.scss
```diff
@@ -120,6 +120,23 @@ body{
     }
 }
 
+@media screen and (-ms-high-contrast: black-on-white) {
+    .nvd3 text {
+        fill: #000 !important;
```
cindylovescoding: Does these have to use !important?

SharmaAjay19: Yes, because we are also trying to override styling from any other sources in this mode.

PR #436, opened 2019-07-16
Support regional geomaster changes in applens

SharmaAjay19:
--Point applens backend from support bay api to wawsobserver
--Deprecate /hostnames api in applens backend as wawsobserver does not support it and the data does not seem to be being used anywhere in frontend.
--Display correct error on request failure.


AngularApp/projects/applens/src/app/modules/site/site-finder/site-finder.component.html
```diff
@@ -1,6 +1,6 @@
 <div class="container-fluid main-container">
   <div class="big-text" *ngIf="loading">Searching for site ...</div>
-  <div class="big-text" *ngIf="!loading && error">Site {{site}} not found</div>
+  <div class="big-text" *ngIf="!loading && error">{{error}}</div>
```
dagrooms52: Not sure what the error is but if the message isn't standard, we probably don't want to show it (even to CSS users).

PR #439, opened 2019-07-19
Rely only on stack from diagnosticProperties call

puneetg1983:
Even though I am sending this PR, this is basically **dead code.**None of the below components are currently used. They were used at some point but not any more. 

So we should be totally ok with making the backend change for properties call.


hforeste:
@puneetg1983  are you also making the diagnostic role change?

cindylovescoding:
I am going to make that change for Diag Role @hforeste

PR #440, opened 2019-07-19
Applens auth flow to RuntimeHost

rekhaswaminathan:
- Created a flag in the appsettings called `UseAppService`. This flag determines if applens needs to talk with Diagrole endpoint or RuntimeHost AFD endpoint. In this way we can merge this without breaking any changes. Once we have RuntimeHost AFD endpoint set up, we can remove this flag completely.

- Created TokenServiceBase class to abstract token acquisition logic and derived classes pass auth context to it.


ApplensBackend/appsettings.json
```diff
@@ -15,10 +15,16 @@
   "ApplicationInsights": {
     "InstrumentationKey": ""
   },
-  "DiagnosticRole": {
+  "DiagnosticService": {
     "endpoint": "",
     "authCertThumbprint": "",
-    "isLocalDevelopment": ""
+    "isLocalDevelopment": "",
+    "UseRuntimeHost": "",
```
hforeste: Maybe change name to `UseAppService` instead?

ApplensBackend/appsettings.json
```diff
@@ -15,10 +15,16 @@
...
+    "RuntimeHostEndpoint":  "",
```
hforeste: Same here `AppServiceEndpoint`

ApplensBackend/appsettings.json
```diff
@@ -15,10 +15,16 @@
   "ApplicationInsights": {
     "InstrumentationKey": ""
   },
-  "DiagnosticRole": {
+  "DiagnosticService": {
```
hforeste: Although this makes sense, I don't think it is necessary since you will have to go to portal and change this value and also notify everyone on the team that you change the app setting name so that locally it doesn't break. But take your call.

PR #442, opened 2019-07-22
Applens china cloud changes

hforeste:
Set up a national cloud angular build and serve configuration for the angular UI project and startup configuration for the applens backend.


ApplensBackend/Program.cs
```diff
@@ -17,10 +18,12 @@ public static IWebHost BuildWebHost(string[] args)
             .AddCommandLine(args)
             .Build();
 
+            var assemblyName = typeof(Startup).GetTypeInfo().Assembly.FullName;
+
             return
             WebHost.CreateDefaultBuilder(args)
                 .UseConfiguration(config)
-                .UseStartup<Startup>()
+                .UseStartup(assemblyName)
```
ShekharGupta1988: Just curious - Any specific reason for this change?

ApplensBackend/Services/ObserverClientService/DiagnosticObserverClientService.cs
```diff
@@ -0,0 +1,91 @@
...
+        /// </summary>
+        /// <param name="diagnosticService">Injected DiagnosticService instance.</param>
+        public DiagnosticObserverClientService(IDiagnosticClientService diagnosticService)
+        {
+            _diagnosticClientService = diagnosticService;
+        }
+
+        public async Task<ObserverResponse> GetHostingEnvironmentDetails(string hostingEnvironmentName)
+        {
+            var hostingEnvironmentDetails = await ExecuteDiagCall($"hostingEnvironments/{hostingEnvironmentName}");
```
ShekharGupta1988: Does this call goes to Diag Role only for mooncake or for public cloud as well?
If later, dont you think its little inefficient in doing all the hops to get to wawsobserver?

hforeste: Only in mooncake

ApplensBackend/Program.cs
hforeste: This allows us to take advantage of .NET Core's ability to use multiple startup classes for different environments. This pretty much tells .NET Core to follow the conventional approach of looking at the `ASPNETCORE_ENVIRONMENT` name and match the StartUp class with that environment. So any cs files with a StartUp<ASPNETCORE_ENVIRONMENT>.cs name will be matched with the appropriate environment.

Since china and public azure have different auth stories and china cloud requires only a subset of the services that we provide for public azure applensv3, the cleanest way to achieve this was by separating the StartUp classes.

ApplensBackend/Services/ObserverClientService/DiagnosticObserverClientService.cs
hforeste: This call will to the localhost observer in diag role in mooncake.

AngularApp/projects/applens/src/app/modules/dashboard/services/applens-diagnostic.service.ts
```diff
@@ -19,7 +19,7 @@ export class ApplensDiagnosticService {
       detector,
       startTime,
       endTime,
-      this._resourceService.getRequestBody(),
+      null,
       refresh,
```
SharmaAjay19: Wow! This means we can eliminate the postBody call from applens load. Nice!

AngularApp/projects/applens/src/app/shared/services/site.service.ts
```diff
@@ -49,12 +46,12 @@ export class SiteService extends ResourceService {
...
 
     public updatePesIdAndImgSrc() {
-        if (this._requestBody.Kind && this._requestBody.Kind.toString().toLowerCase().indexOf("functionapp") !== -1) {
+        if (this._siteObject.Kind && this._siteObject.Kind.toString().toLowerCase().indexOf("functionapp") !== -1) {
             this.pesId = '16072';
             this.imgSrc = 'assets/img/Azure-Functions-Logo.png';
             this.staticSelfHelpContent = 'microsoft.function';
         }
-        else if (this._requestBody.IsLinux) {
+        else if (this._siteObject.IsLinux != undefined && this._siteObject.IsLinux) {
```
cindylovescoding: this is same as : else if (this._siteObject.IsLinux) right?

AngularApp/projects/applens/src/app/modules/dashboard/dashboard/dashboard.component.ts
```diff
@@ -73,16 +74,18 @@ export class DashboardComponent implements OnDestroy {
...
-    this.userId = alias.replace('@microsoft.com', '');
-    this._diagnosticService.getUserPhoto(this.userId).subscribe(image => {
-      this.userPhotoSource = image;
-    });
-
-    this._diagnosticService.getUserInfo(this.userId).subscribe((userInfo: UserInfo) => {
-      this.userName = userInfo.givenName;
-      this.displayName = userInfo.displayName;
-    });
+    if(environment.adal.enabled){
```
cindylovescoding: So for now we will see a spinning circle for the user pic right?

ApplensBackend/appsettings.json
```diff
@@ -25,7 +25,8 @@
     "resourceId": "",
     "clientSecret": "",
     "authority": "",
-    "targetSupportApiTestSlot": "false"
+    "targetSupportApiTestSlot": "false",
+    "diagnosticObserverEnabled": false
```
cindylovescoding: nitpick: Do we need to make the values in configuration json [IConfiguration Object] to be string type? or it will be good to keep it as boolean?

AngularApp/projects/applens/src/app/modules/dashboard/dashboard/dashboard.component.ts
hforeste: Yep, I have a follow up plan to get the user information. So consider this as a first iteration.

AngularApp/projects/applens/src/app/shared/services/site.service.ts
hforeste: I don't think that can work here right? Otherwise it would be kind of like a NullReferenceException if I just do `else if(this._siteObject.IsLinux)`

ApplensBackend/appsettings.json
hforeste: I like to keep it as boolean.

PR #443, opened 2019-07-22
Small fix for ASE on UI

hforeste:


hforeste:
The angular routing module was not properly routing because the expected `subscription` JSON property was missing from the backend response. Instead use the already populated `customerSubscriptionId` property from response.

ShekharGupta1988:
Short and sweet :)

PR #444, opened 2019-07-22
Geomaster hostname header for data provider

SharmaAjay19:
- Pass geomaster hostname header from applens frontend to applens backend correctly
- Pass all x-ms- & diag- headers from applens backend to runtime host

### DIFF
--- a/ApplensBackend/Controllers/DiagnosticController.cs
+++ b/ApplensBackend/Controllers/DiagnosticController.cs
@@ -145,6 +145,13 @@ namespace AppLensV3.Controllers
                 headers.Add("diag-assembly-name", assemblyName);
             }
 
+            // Add all remaining headers with x-ms- or diag- prefix to request
+            var allRequestHeaders = Request.Headers.ToDictionary(header => header.Key, header => header.Value, StringComparer.OrdinalIgnoreCase);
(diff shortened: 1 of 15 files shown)


ApplensBackend/Controllers/DiagnosticController.cs
```diff
@@ -145,6 +145,13 @@ public async Task<IActionResult> Invoke([FromBody]JToken body)
                 headers.Add("diag-assembly-name", assemblyName);
             }
 
+            // Add all remaining headers with x-ms- or diag- prefix to request
+            var allRequestHeaders = Request.Headers.ToDictionary(header => header.Key, header => header.Value, StringComparer.OrdinalIgnoreCase);
+            foreach(var item in allRequestHeaders){
+                if (item.Key.StartsWith("x-ms-") || item.Key.StartsWith("diag-")){
+                    headers.Add(item.Key, item.Value.ToString());
```
ShekharGupta1988: Will this add duplicate headers in the dictionary for the ones which are already added on the top..

ShekharGupta1988: actually nvmind, I see you are handling that while adding these to request.

PR #447, opened 2019-07-26
Infer Solution ActionType From options

dagrooms52:
- Infer Solution ActionType based on the options, ignoring the enum from the backend
- Avoids backwards-compatibility issues with enum
- Avoids backend changes for new Solution types


dagrooms52:
The entire file diffed because of whitespace (changed 2 space indent to 4 automatically); the only addition here is the `inferActionType` method and its call on `ngOnInit`

puneetg1983:
The build seems to be failing. Can you double check the change or queue another build to confirm that this is not breaking anything.

PR #448, opened 2019-07-29
Navigator fixes

rekhaswaminathan:
- Getting changes for dependent resources by passing form query string parameters to the detector.
- Disabling zoom and drag in the graph 
- allow scan for dependent resources

### DIFF
--- a/AngularApp/projects/diagnostic-data/src/lib/components/app-dependencies/app-dependencies.component.html
+++ b/AngularApp/projects/diagnostic-data/src/lib/components/app-dependencies/app-dependencies.component.html
@@ -3,12 +3,7 @@
   <input type="hidden" id="resourceUri" (change)="refreshChangeTimeline()">
 
 <div *ngIf="showNotSupport">
-  {{selectedResourceType}} is not supported on Change Analysis. Please see the below list of supported resources.
-  <ul>
-    <li *ngFor="let resource of supportedResources">
-      {{ resource }}
-    </li>
-  </ul>
+  {{selectedResourceType}} is not supported on Change Analysis. View other dependency resource changes to isolate potential causes to an issue.
 </div>
 <h5 *ngIf="showLoader">
   <span class="fa fa-circle-o-notch fa-spin spin-icon"></span> Loading change groups for {{selectedResourceId}}...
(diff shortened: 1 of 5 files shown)


AngularApp/projects/diagnostic-data/src/lib/components/app-dependencies/app-dependencies.component.html
```diff
@@ -3,12 +3,7 @@
   <input type="hidden" id="resourceUri" (change)="refreshChangeTimeline()">
 
 <div *ngIf="showNotSupport">
-  {{selectedResourceType}} is not supported on Change Analysis. Please see the below list of supported resources.
-  <ul>
-    <li *ngFor="let resource of supportedResources">
-      {{ resource }}
-    </li>
-  </ul>
+  {{selectedResourceType}} is not supported on Change Analysis. View other dependency resource changes to isolate potential causes to an issue.
```
nmallick1: Is it too much of work to detect if other resources changed and only depending on that advice the user to view other resources ?

AngularApp/projects/diagnostic-data/src/lib/components/changesets-view/changesets-view.component.ts
```diff
@@ -89,7 +89,7 @@ export class ChangesetsViewComponent extends DataRenderBaseComponent implements
                 // Convert UTC timestamp to user readable date
                 this.scanDate = rows[0][6] != '' ? 'Changes were last scanned on ' + moment(rows[0][6]).format("ddd, MMM D YYYY, h:mm:ss a") : this.noScanMsg + ' Please enable Change Analysis using Change Analysis Settings.';
             } else {
-                this.scanDate = '';
+                this.scanDate = rows[0][6] != '' ? 'Changes were last scanned on ' + moment(rows[0][6]).format("ddd, MMM D YYYY, h:mm:ss a") : 'No recent scans were performed on this resource.';
```
nmallick1: Is there a possibility that the value in rows[0][6] might be some string but not a datetime ? Can the moment() call fail ?

AngularApp/projects/diagnostic-data/src/lib/components/app-dependencies/app-dependencies.component.html
rekhaswaminathan: So we have a work item for future which includes a UI change to display the count of changes for each resource on the graph. That will give the user an idea if other resources has changes.

AngularApp/projects/diagnostic-data/src/lib/components/changesets-view/changesets-view.component.ts
rekhaswaminathan: The way it can happen is if the model in the gist changes. For now it is the datetime string in the model.

PR #454, opened 2019-08-03
Nmallick1/partner case submission

nmallick1:
Enable Diagnose And Solve for case submission and portal for ARM resources


AngularApp/projects/app-service-diagnostics/src/app/shared/services/generic-arm-config.service.ts
```diff
@@ -0,0 +1,384 @@
...
+    const baseUri: string = 'armResourceConfig/' + resourceDesc.provider + '/' + resourceDesc.types.join('/') + '/';
+
+    let override = this._http.get<ArmResourceConfig>(baseUri + 'override.json')
+    let config = this._http.get<ArmResourceConfig>(baseUri + 'config.json')
+
+    return forkJoin(override, config).pipe(catchError(err => of(err)),
+      map(configs => {
+        if (!(configs instanceof Array)) {
+          console.log(`Error occurred while reading arm resource config. Details...`);
+          console.log(configs);
```
rekhaswaminathan: was this meant for debugging only?

nmallick1: Yeah, but I do want to stay in there. Just in case some partner messes up the config, the portal will not load for them. I want the console logs to show what error was encountered while reading the config.

AngularApp/projects/app-service-diagnostics/src/app/shared/services/generic-arm-config.service.ts
```diff
@@ -0,0 +1,384 @@
...
+
+  getArmResourceConfig(resourceUri: string, recurse?: boolean): ArmResourceConfig {
+    let returlValue: ArmResourceConfig;
+    if (this.resourceMap.length > 0) {
+      this.resourceMap.some((resource: ArmResourceConfig) => {
+        const matchPattern: RegExp = new RegExp(`${resource.matchRegEx.replace(/[-\/\\^$*+?.()|[\]{}]/g, '\\$&')}`, "i");
+        var result = resourceUri.match(matchPattern);
+        if (result && result.length > 0) {
+          returlValue = resource;
+          return true;
```
rekhaswaminathan: why do we need to return true explicitly here ?

AngularApp/projects/app-service-diagnostics/src/app/shared/services/generic-arm-config.service.ts
rekhaswaminathan: Got it, do you want to consider log the exception instead of console logging?

nmallick1: The exception will be auto logged by the unhandled exception handler that's already in place. Logging it in console will just let us get to the issue quicker.

AngularApp/projects/app-service-diagnostics/src/app/shared/services/generic-arm-config.service.ts
nmallick1: Yeah, return true will terminate the execution of Array.some so I do not end up processing additional elements.

AngularApp/projects/app-service-diagnostics/src/app/shared/services/generic-arm-config.service.ts
rekhaswaminathan: Oh i see. got it

AngularApp/projects/app-service-diagnostics/src/app/home/components/home/home.component.html
```diff
@@ -20,12 +20,18 @@
...
-    </p>
+    <span *ngIf = "homePageText === null">
+      <h2 tabindex="0">App Service Diagnostics</h2>
+      <p id="page-description" tabindex="0">
+          App Service Diagnostics to investigate how your app is performing, diagnose issues, and discover how to
+          improve your application. Select the problem category that best matches the information or tool that you're
+          interested in:
+      </p>
+    </span>
+    <span *ngIf = "homePageText !== null">
```
hforeste: I recommend just having a default `HomePageText` for App Service Diagnostics that way you can keep this HTML simple.

AngularApp/projects/app-service-diagnostics/src/app/home/components/home/home.component.ts
```diff
@@ -26,16 +27,37 @@ export class HomeComponent implements OnInit {
...
+
+    if (!!_resourceService.armResourceConfig && !!_resourceService.armResourceConfig.homePageText
+      && !!_resourceService.armResourceConfig.homePageText.title && _resourceService.armResourceConfig.homePageText.title.length > 1
+      && !!_resourceService.armResourceConfig.homePageText.description && _resourceService.armResourceConfig.homePageText.description.length > 1
+      && !!_resourceService.armResourceConfig.homePageText.searchBarPlaceHolder && _resourceService.armResourceConfig.homePageText.searchBarPlaceHolder.length > 1) {
+      this.homePageText = _resourceService.armResourceConfig.homePageText;
+      this.searchPlaceHolder = this.homePageText.searchBarPlaceHolder;
+    }
+    else {
+      this.homePageText = null;
```
hforeste: As mentioned above about a default `HomePageText` for App Service Diagnostics

AngularApp/projects/app-service-diagnostics/src/app/shared-v2/services/category.service.ts
```diff
@@ -12,10 +13,19 @@ export class CategoryService {
 
   private _commonCategories: Category[] = [];
 
-  constructor() {
+  constructor(private _genericArmConfigService?: GenericArmConfigService) {
     this._addCategories(this._commonCategories);
   }
 
+  public initCategoriesForArmResource(resourceUri: string) {
+    if (!!this._genericArmConfigService) {
```
hforeste: nitpik: I think you can technically just say `if(this._genericArmConfigService)` and it will have the same effect. https://medium.com/better-programming/javascript-bang-bang-i-shot-you-down-use-of-double-bangs-in-javascript-7c9d94446054

nmallick1: Funny :) I actually wasn't using double bangs to begin with. Our UX code however follows a practice of using double bangs specifically so I thought this is the best practice.

I can either use / not use double bangs based on what we decide. Happy to change :)

AngularApp/projects/app-service-diagnostics/src/app/shared-v2/services/resource.service.ts
```diff
@@ -26,30 +28,145 @@ export class ResourceService {
...
+    }
+    else {
+      return null;
+    }
+  }
+
   public get searchSuffix(): string {
-    return 'Azure';
+    if (!!this._genericArmConfigService) {
+      let currConfig: ArmResourceConfig = this._genericArmConfigService.getArmResourceConfig(this.resource.id);
```
hforeste: Nazeef, would it be valuable that we also create an ArmResourceConfig for Microsoft.Web too? That way you don't need these multiple if else cases for generic and our default experience.

AngularApp/projects/app-service-diagnostics/src/app/shared/services/arm.service.ts
```diff
@@ -19,18 +20,37 @@ export class ArmService {
...
+
+    }
+
+    getApiVersion(resourceUri: string, apiVersion?: string): string {
+        if (!!apiVersion) {
+            return apiVersion;
+        }
+        else {
+            if (!!this._genericArmConfigService) {
+                apiVersion = this._genericArmConfigService.getApiVersion(resourceUri);
```
hforeste: Same idea here, maybe create an ArmConfigService for Microsoft.Web so that you can keep this file clean.

AngularApp/projects/app-service-diagnostics/src/app/shared-v2/services/resource.service.ts
nmallick1: Actually, there are too many customizations that Microsoft.Web employs that I haven't made generic yet. Including breaking the workflow and opening a different page (for tools) instead of showing detectors under the category. Showing custom Genie messages for Availability and Perf instead of the default messages. 

I'd leave Microsoft.Web as is for now till the entire thing is generalized.

AngularApp/projects/app-service-diagnostics/src/app/shared-v2/services/resource.service.ts
```diff
@@ -26,30 +28,145 @@ export class ResourceService {
...
 
   public get subscriptionId(): string {
     return this._subscription;
   }
 
   public get isApplicableForLiveChat(): boolean {
-    return false;
+    if (!!this._genericArmConfigService) {
+      let currConfig: ArmResourceConfig = this._genericArmConfigService.getArmResourceConfig(this.resource.id);
+      if (currConfig.isApplicableForLiveChat) {
```
cindylovescoding: if (currConfig.isApplicableForLiveChat) {
        return currConfig.isApplicableForLiveChat;
      }
      else {
        return false;
      } 
 
this is the same as :
return currConfig.isApplicableForLiveChat;

right?

nmallick1: if (currConfig.isApplicableForLiveChat)  checks for the existence of the property. If the property is not present, the we assume that LiveChat is not enabled. If it is present, then we return the value. 

If I just did return currConfig.isApplicableForLiveChat, then if the config does not contain the flag (due to user error) then it will fail further downstream that expects a boolean.

nmallick1: For readability, I think doing a if(!!currConfig.isApplicableForLiveChat) will be better. It conveys the intent better.

PR #455, opened 2019-08-05
Fix applens url get truncated in email feedback

cindylovescoding:


nmallick1:
I see you simplified the URL however is this PR fixing applens url getting truncated ?

cindylovescoding:
@nmallick1  So Nazeef the main change is actually I am encoding the URI parameters before put them into href.  Previously "&" in "&endtime=" is not encoded correctly that's why we are not able to see endtime param in the URI.

PR #456, opened 2019-08-06
Set default end time to now - 15 minutes

nmallick1:
Kusto has a lag of 15 minutes so defaulting the end time to now - 15 minutes. This should hopefully reduce the Kusto load and avoid it from forcefully reading from disk.

### DIFF
--- a/AngularApp/projects/diagnostic-data/src/lib/services/detector-control.service.ts
+++ b/AngularApp/projects/diagnostic-data/src/lib/services/detector-control.service.ts
@@ -170,7 +170,7 @@ export class DetectorControlService {
     if (start && end) {
       startTime = moment.utc(start);
       if(moment.duration(moment.utc().diff(moment.utc(end))).asMinutes() < 16 ) {
-        //The supplied end time > now - 15 minutes. Adjust the end time so that it becomes now())-15 minutes.
+        //The supplied end time > now - 15 minutes. Adjust the end time so that it becomes now()-15 minutes.
         endTime = moment.utc().subtract(16, 'minutes');
       }
       else {
@@ -180,7 +180,7 @@ export class DetectorControlService {
     } else if (start) {
       startTime = moment.utc(start);
       if (moment.duration(moment.utc().diff(startTime.clone().add(1, 'days'))).asMinutes() < 16) {
-        //The new endtime > now() - 15 minutes of. Adjust the end time so that it becomes now())-15 minutes.
+        //No endtime was passed. If (start time + 1 day) > (now() - 15 minutes), adjust the end time so that it becomes less than now()-15 minutes.
         endTime = moment.utc().subtract(16, 'minutes');
       }
       else {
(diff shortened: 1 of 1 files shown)


AngularApp/projects/diagnostic-data/src/lib/services/detector-control.service.ts
```diff
@@ -92,13 +92,13 @@ export class DetectorControlService {
...
-      if (moment.duration(moment.utc().diff(start)).asMinutes() < 0) {
-        returnValue = 'Start date time must be less than current date time';
+      if (moment.duration(moment.utc().diff(start)).asMinutes() < 30) {
+        returnValue = 'Start date time must be 30 minutes less than current date time';
         this.timeRangeErrorString = returnValue;
         return returnValue;
       }
-      if (moment.duration(moment.utc().diff(end)).asMinutes() < 0) {
-        returnValue = 'End date time must be less than current date time';
+      if (moment.duration(moment.utc().diff(end)).asMinutes() < 0) {
```
cindylovescoding: Should this be  < 15?

AngularApp/projects/diagnostic-data/src/lib/services/detector-control.service.ts
```diff
@@ -169,12 +169,32 @@ export class DetectorControlService {
...
+      }
+      else {
+        endTime = moment.utc(end);
+      }
+      
     } else if (start) {
       startTime = moment.utc(start);
-      endTime = startTime.clone().add(1, 'days');
+      if (moment.duration(moment.utc().diff(startTime.clone().add(1, 'days'))).asMinutes() < 16) {
+        //The new endtime > now() - 15 minutes of. Adjust the end time so that it becomes now())-15 minutes.
```
cindylovescoding: Could we update the comment and make this more clear:)

AngularApp/projects/diagnostic-data/src/lib/services/detector-control.service.ts
```diff
@@ -169,12 +169,32 @@ export class DetectorControlService {
...
+        endTime = moment.utc().subtract(16, 'minutes');
+      }
+      else {
+        endTime = startTime.clone().add(1, 'days');
+      }
     } else if (end) {
-      endTime = moment.utc(end);
+      if(moment.duration(moment.utc().diff(moment.utc(end))).asMinutes() < 16) {
```
cindylovescoding: Instead of using many if-else clauses, it will be great if we have a function to validate and modify starttime and endtime based on the validation result.

AngularApp/projects/diagnostic-data/src/lib/services/detector-control.service.ts
nmallick1: I tried this earlier, however since all our detectors send EndTime as now by default, this will always show up as an error on the browser even though the user isn't the one making these changes. There are multiple places in the UI where the EndTime is being passed from, so setting it to now-15 minutes without throwing an error does the job and keeps the UI pretty.

AngularApp/projects/diagnostic-data/src/lib/services/detector-control.service.ts
nmallick1: Lets sync up offline on this.

PR #457, opened 2019-08-07
Surface asc insights

hforeste:


nmallick1:
Is this change only for mooncake or for all ? Also, after this change are we eliminating the need of the proxy project ?

PR #465, opened 2019-08-14
Enabling dynamic auth

hforeste:
Enable AAD auth and WS-Federation for National Cloud environments.

### DIFF
--- a/ApplensBackend/Program.cs
+++ b/ApplensBackend/Program.cs
@@ -1,4 +1,8 @@
-﻿using System.Reflection;
+﻿using System;
+using System.Linq;
+using System.Net;
+using System.Reflection;
+using System.Security.Cryptography.X509Certificates;
 using Microsoft.AspNetCore;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.Extensions.Configuration;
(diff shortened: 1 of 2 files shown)


ApplensBackend/Program.cs
```diff
@@ -20,11 +24,25 @@ public static IWebHost BuildWebHost(string[] args)
...
                 .UseConfiguration(config)
-                .UseStartup(assemblyName)
-                .Build();
+                .UseStartup(assemblyName);
+
+            if (config.GetValue<bool>("useHttps"))
+            {
+                var store = new X509Store(StoreName.My, StoreLocation.CurrentUser);
+                store.Open(OpenFlags.ReadOnly);
+                var serverCertificate = store.Certificates[0];
```
rekhaswaminathan: is there a reason why you always want the first certificate? Because I noticed in other places we search for a certificate by thumbprint

hforeste: Kestral requires a server certificate to run HTTPS and since I'm only running this for local development and testing, any server certificate will do.

rekhaswaminathan: I see. The config `useHttps` means you're running this for local development?

hforeste: Yep

PR #467, opened 2019-08-15
Addressed Change Analysis UI feedback

rekhaswaminathan:
- Highlight the most recent change group in the default view to indicate that it’s been selected. 
- Change groups should be monotone in color unless selected. 
- Change the cursor to a pointer style when a user hovers over a change group. 
- Make the top row from the change chart expanded by default

### DIFF
--- a/AngularApp/projects/diagnostic-data/src/lib/components/changes-view/changes-view.component.ts
+++ b/AngularApp/projects/diagnostic-data/src/lib/components/changes-view/changes-view.component.ts
@@ -96,6 +96,7 @@ export class ChangesViewComponent extends DataRenderBaseComponent implements OnI
                 });
             });
             this.tableItems.sort((i1, i2) => i1.level - i2.level);
+            this.expandedElement = this.tableItems[0];
             this.dataSource = new MatTableDataSource(this.tableItems);
         }
     }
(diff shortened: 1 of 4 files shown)


AngularApp/projects/diagnostic-data/src/lib/components/changes-view/changes-view.component.ts
```diff
@@ -96,6 +96,7 @@ export class ChangesViewComponent extends DataRenderBaseComponent implements OnI
                 });
             });
             this.tableItems.sort((i1, i2) => i1.level - i2.level);
+            this.expandedElement = this.tableItems[0];
```
yantang-msft: tableItems can have no elements

AngularApp/projects/diagnostic-data/src/lib/components/changesets-view/changesets-view.component.ts
```diff
@@ -171,6 +170,7 @@ export class ChangesetsViewComponent extends DataRenderBaseComponent implements
         // Create a Timeline
         this.changesTimeline = new Timeline(container, this.timeLineDataSet, this.sourceGroups, options);
         this.changesTimeline.on('select', this.triggerChangeEvent);
+        this.changesTimeline.setSelection(changeSets[0][0]);
```
yantang-msft: similar that changesets can have no elements

AngularApp/projects/diagnostic-data/src/lib/components/changes-view/changes-view.component.ts
rekhaswaminathan: this is wrapped in an outer if condition which checks for the rows length.

AngularApp/projects/diagnostic-data/src/lib/components/changesets-view/changesets-view.component.ts
rekhaswaminathan: same here. It is wrapped in an outer if condition which checks if there are rows present.

PR #488, opened 2019-09-09
Change filter and detector icons

cindylovescoding:


khaled-zayed:
@cindylovescoding can you please update the placeholder text to make it "Filter" instead of "filter"

PR #491, opened 2019-09-10
Updating config

yunjchoi:
Updated the config file to show Cluster Management as its own category in AKS Diagnostics and to change the placeholder text for the search bar.

### DIFF
--- a/AngularApp/projects/app-service-diagnostics/src/armResourceConfig/microsoft.containerservice/managedclusters/config.json
+++ b/AngularApp/projects/app-service-diagnostics/src/armResourceConfig/microsoft.containerservice/managedclusters/config.json
@@ -1,9 +1,9 @@
 {
-	"homePageText": {
-		"title": "Azure Kubernetes Services Diagnostics",
-		"description": "Use Azure Kubernetes Services Diagnostics to investigate how your cluster is performing, diagnose issues, and discover how to improve its reliability. Select the problem category that best matches the information or tool that you're interested in:",
-		"searchBarPlaceHolder": "Search Azure Kubernetes Services"
-	},
+    "homePageText": {
+        "title": "Azure Kubernetes Services Diagnostics",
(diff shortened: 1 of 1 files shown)


yunjchoi:
@nmallick1, we are going to surface Cluster Management detector which resides in Customer category on AppLens to customers. I set the id to Customer and name to Cluster Management. Is this the right approach for showing Cluster Management as the category name?

nmallick1:
@yunjchoi , nope, this will not work. The category array is merely to identify a category, even if there is a single detector under the category, the id and name in the file are matched against the category in applens, not the detector name

yunjchoi:
@nmallik1 I misphrased my question. Would this map the customer-facing category to Customer category in AppLens, but show the name as Cluster Management?

yunjchoi:
I'll update the config once the category name is updated in AppLens.

yunjchoi:
@nmallick1, category name is now updated to match AppLens.

AngularApp/projects/app-service-diagnostics/src/armResourceConfig/microsoft.containerservice/managedclusters/config.json
```diff
@@ -1,32 +1,24 @@
...
-		}
-	]
+    "categories": [
+        {
+            "id": "Cluster Insights",
+            "name": "Cluster Insights",
+            "description": "Is your cluster experiencing failures or unresponsiveness? Investigate and discover issues that may cause your cluster to no longer be manageable.",
+            "keywords": [ "Failures", "Node Readiness", "Node Health", "Scaling", "CRUD", "Identity" ],
+            "color": "rgb(186, 211, 245)",
+            "createFlowForCategory": false,
```
nmallick1: Please change this flag to true. Although it is set to true in override.json so there won't be any functionality issues, however the correct value for this flag is always true.

yunjchoi: Thanks Nazeef, I changed the value to true.

PR #492, opened 2019-09-10
Revert "Change reroute path and remove post request in constructor"

XiaoxuHU:
Reverts Azure/Azure-AppServices-Diagnostics-Portal#487


AngularApp/projects/app-service-diagnostics/src/app/supportbot/message-flow/cpu-analysis-chat/cpu-analysis-chat-flow.ts
```diff
@@ -45,7 +45,7 @@ export class CpuAnalysisChatFlow extends IMessageFlowProvider {
             this._siteService.currentSiteMetaData.subscribe(siteInfo => {
                 if (siteInfo) {
                     this.siteInfoMetaData = siteInfo;
-                    // this._getCpuDetectorResponse();
+                    this._getCpuDetectorResponse();
```
ShekharGupta1988: Can we still comment this line.. I dont think  this is the call causing the dotnet core server not to start..

PR #501, opened 2019-09-13
Enable search case submission flow only for OSS

SharmaAjay19:


AngularApp/projects/app-service-diagnostics/src/app/shared-v2/services/support-topic.service.ts
```diff
@@ -16,7 +16,7 @@ export class SupportTopicService {
   public pesId: string = "";
   private selfHelpContentUrl = "https://mpac.support.ext.azure.com/api/v1/selfHelpArticles?articleTypes=Generic&articleTypes=Resource";
   private supportTopicConfig = {
-    "14748": ["32444077", "32444080", "32444081", "32444082", "32444083", "32444084", "32550703", "32542218"]
+    "14748": ["32444077", "32444080", "32444081", "32444082", "32444083", "32444084", "32550703"]
```
ShekharGupta1988: nitpick - Do we need this list hardcoded anymore now?

SharmaAjay19: Yes for now we need this. We will incrementally remove this in next iteration.

PR #512, opened 2019-09-26
EventViewer and FrebViewer from v3

puneetg1983:
**IMPORTANT** - In this PR, there is a change that will hide detectors that do not have description OR Category next to them. This is done to ensure that all the detectors that are noise (like the one for the home page and even EventViewer) don't show up directly under the Search. 

Can you guys think of any detectors for any resource type that exist without Category or Description

image

image


AngularApp/projects/app-service-diagnostics/src/app/shared-v2/services/feature.service.ts
```diff
@@ -24,28 +24,30 @@ export class FeatureService {
...
-              id: detector.id,
-              description: detector.description,
-              category: detector.category,
-              featureType: FeatureTypes.Detector,
-              name: detector.name,
-              clickAction: this._createFeatureAction(detector.name, detector.category, () => {
-                this._router.navigateByUrl(`resource${startupInfo.resourceId}/analysis/${detector.id}`);
-              })
-            });
+          if (detector.category || detector.description) {
```
nmallick1: You mentioned that there is a change that will hide detectors that do not have description OR Category next to them. Wouldn't this mean that if either one is present, the detector is not hidden and only hidden when both of these are not present ?

puneetg1983: Yes, detector will be hidden if BOTH are NOT present

ShekharGupta1988: will this be true for if detector.description  = "" (not null but empty)

puneetg1983: Added an additional check for string length just to be sure.

PR #516, opened 2019-10-01
Open detector view if there is only one detector under a category els…

nmallick1:
Open detector view if there is only one detector under a category else show a list of detectors in the regular genie flow.

Newly on-boarded partners have only one public detector in a given category and for them forcing a genie experience for a single category is an overkill.


hforeste:
This behavior seems fine to me but do you want to get a PM perspective on it

nmallick1:
> This behavior seems fine to me but do you want to get a PM perspective on it

Request to support this navigation for partner teams came from PM (Yun Jung)

PR #522, opened 2019-10-08
Fix the timeseries rendering issue

cindylovescoding:
1. Now timeseries support kusto query bin by any timegrain
2. Fix the issue that sometimes timeseries data won't be rendered properly


stevenernst131:
What bugs!

PR #525, opened 2019-10-15
UI improvements for Change Analysis batch API

rekhaswaminathan:
- Display last 24h changes by default
- Noisy changes filtered out by default
- Last Scan timestamp, timestamp on each change, timeline displayed in UTC
- Removed noise icon from table legend

### DIFF
--- a/AngularApp/projects/diagnostic-data/src/lib/components/changesets-view/changesets-view.component.ts
+++ b/AngularApp/projects/diagnostic-data/src/lib/components/changesets-view/changesets-view.component.ts
@@ -569,9 +569,6 @@ export class ChangesetsViewComponent extends DataRenderBaseComponent implements 
     }
 
     getBatchChanges(data: DataTableResponseObject): any[][] {
-        if(data.rows.length < 1) {
-            return [];
-        }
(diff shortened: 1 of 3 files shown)


AngularApp/projects/diagnostic-data/src/lib/components/changesets-view/changesets-view.component.ts
```diff
@@ -565,4 +567,18 @@ export class ChangesetsViewComponent extends DataRenderBaseComponent implements
         }
         return [changeSet[initiatedByIndex]];
     }
+
+    getBatchChanges(data: DataTableResponseObject): any[][] {
+        if(data.rows.length < 1) {
+            return [];
+        }
```
cindylovescoding: Do you need this logic in this function? 
"  if(data.rows.length < 1) {
            return [];
        }"
Nitpick: this is already handled by the logic below

rekhaswaminathan: Good catch @cindylovescoding. I can remove it.

PR #526, opened 2019-10-16
Fix the routing to analysis in case there is a single public detector in a category

nmallick1:
Added a check to see if the detector is analysis or a detector and route accordingly.

### DIFF
--- a/AngularApp/projects/app-service-diagnostics/src/app/home/components/category-tile/category-tile.component.ts
+++ b/AngularApp/projects/app-service-diagnostics/src/app/home/components/category-tile/category-tile.component.ts
@@ -3,7 +3,7 @@ import { Category } from '../../../shared-v2/models/category';
 import { ActivatedRoute, NavigationExtras, Router } from '@angular/router';
 import { NotificationService } from '../../../shared-v2/services/notification.service';
 import { LoggingV2Service } from '../../../shared-v2/services/logging-v2.service';
-import { DiagnosticService, DetectorMetaData } from 'diagnostic-data';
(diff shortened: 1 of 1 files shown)


AngularApp/projects/app-service-diagnostics/src/app/home/components/category-tile/category-tile.component.ts
```diff
@@ -37,7 +37,11 @@ export class CategoryTileComponent implements OnInit {
       if (currentCategoryDetectors.length === 1) {
         this._notificationService.dismiss();
         this._logger.LogTopLevelDetector(currentCategoryDetectors[0].id, currentCategoryDetectors[0].name, this.category.id);
-        this._router.navigateByUrl(`resource${this._resourceService.resourceIdForRouting}/detectors/${currentCategoryDetectors[0].id}`);
+        if(currentCategoryDetectors[0].type === DetectorType.Detector) {
+          this._router.navigateByUrl(`resource${this._resourceService.resourceIdForRouting}/detectors/${currentCategoryDetectors[0].id}`);
+        }  else if (currentCategoryDetectors[0].type === DetectorType.Analysis) {
```
puneetg1983: What happens when category has more than one detector? Why we checking for first detector only?

nmallick1: This is the case when if a category has only one detector, it needs to directly open that single detector under that category. If there are more than one detectors, then we open the category view and initiate the whole genie flow.

puneetg1983: I see

PR #527, opened 2019-10-17
Fixed few bugs in batch api

rekhaswaminathan:
- Display last scan timestamp in UTC
- Removed .length to prevent null reference


AngularApp/projects/diagnostic-data/src/lib/components/changesets-view/changesets-view.component.ts
```diff
@@ -572,7 +572,7 @@ export class ChangesetsViewComponent extends DataRenderBaseComponent implements
         let changeDetails = [];
         let columnIndex = DataTableUtilities.getColumnIndexByName(data, "Inputs");
         data.rows.forEach(row => {
-            if (row[columnIndex] != null || row[columnIndex].length > 0) {
```
cindylovescoding: Maybe change this to  if (row[columnIndex] != null && row[columnIndex].length > 0) ?

rekhaswaminathan: Thanks Cindy. Changed.

PR #529, opened 2019-10-21
Nationalclouddev

hforeste:
This fixes https://dev.azure.com/app-service-diagnostics-portal/app-service-diagnostics-portal/_workitems/edit/803/


hforeste:
What this will do is render the page of detectors without the user images 
image

PR #530, opened 2019-10-21
Updated Change Analysis insight title, removed caching for getDetector API call

rekhaswaminathan:
- Updated Change Analysis insight title to reflect number of changes instead of change groups.
  Current UI:
image

 New UI:
image

- Removed caching call for getDetector API call by passing `shouldRefresh` parameter as `true`

### DIFF
--- a/AngularApp/projects/diagnostic-data/src/lib/components/changes-view/changes-view.component.ts
+++ b/AngularApp/projects/diagnostic-data/src/lib/components/changes-view/changes-view.component.ts
@@ -89,7 +89,7 @@ export class ChangesViewComponent extends DataRenderBaseComponent implements OnI
                     'jsonPath': jsonPath,
                     'originalModel': ChangeAnalysisUtilities.prepareValuesForDiffView(oldValue),
                     'modifiedModel': ChangeAnalysisUtilities.prepareValuesForDiffView(newValue),
-                    'initiatedByList' : row['initiatedByList'] ? row['initiatedByList'] : this.initiatedByList
+                    'initiatedByList' : row['initiatedByList'] ? this.getInitiatedByUsers(row) : this.initiatedByList
                 });
             });
             this.tableItems.sort((i1, i2) => i1.level - i2.level);
@@ -176,6 +176,10 @@ export class ChangesViewComponent extends DataRenderBaseComponent implements OnI
         };
 
         return eventProps;
+    }
+
+    getInitiatedByUsers(row: any[]): any[] {
+        return row['initiatedByList'].length > 0 ? row['initiatedByList'] : ['Unable to Determine'];
     }
 }
 
--- a/AngularApp/projects/diagnostic-data/src/lib/components/changesets-view/changesets-view.component.ts
+++ b/AngularApp/projects/diagnostic-data/src/lib/components/changesets-view/changesets-view.component.ts
@@ -79,7 +79,8 @@ export class ChangesetsViewComponent extends DataRenderBaseComponent implements 
         let resourceName = this.changeAnalysisService.getCurrentResourceName();
         let isAppService = this.changeAnalysisService.getAppService();
         if (rows.length > 0 && rows[0].length > 0) {
-            this.changeSetText = rows.length === 1 ? `1 change group detected` : `${rows.length} change groups have been detected`;
+            let batchChanges = this.getBatchChanges(data);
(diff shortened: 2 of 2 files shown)


AngularApp/projects/diagnostic-data/src/lib/components/changes-view/changes-view.component.ts
```diff
@@ -177,5 +177,9 @@ export class ChangesViewComponent extends DataRenderBaseComponent implements OnI
 
         return eventProps;
     }
+
+    getInitiatedByUsers(row: any[]): any[] {
```
cindylovescoding: Is this input "row" an array?

AngularApp/projects/diagnostic-data/src/lib/components/changes-view/changes-view.component.ts
```diff
@@ -89,7 +89,7 @@ export class ChangesViewComponent extends DataRenderBaseComponent implements OnI
                     'jsonPath': jsonPath,
                     'originalModel': ChangeAnalysisUtilities.prepareValuesForDiffView(oldValue),
                     'modifiedModel': ChangeAnalysisUtilities.prepareValuesForDiffView(newValue),
-                    'initiatedByList' : row['initiatedByList'] ? row['initiatedByList'] : this.initiatedByList
+                    'initiatedByList' : row['initiatedByList'] ? this.getInitiatedByUsers(row) : this.initiatedByList
```
SharmaAjay19: This "row" looks like a json object, I see a below usage as array.

AngularApp/projects/diagnostic-data/src/lib/components/changesets-view/changesets-view.component.ts
```diff
@@ -252,7 +253,7 @@ export class ChangesetsViewComponent extends DataRenderBaseComponent implements
             }
 
             this.diagnosticService.getDetector(this.detector, this.detectorControlService.startTimeString, this.detectorControlService.endTimeString,
-            this.detectorControlService.shouldRefresh, this.detectorControlService.isInternalView, queryParams).subscribe((response: DetectorResponse) =>{
+            true, this.detectorControlService.isInternalView, queryParams).subscribe((response: DetectorResponse) =>{
```
cindylovescoding: So the changesets view will always refresh now?

AngularApp/projects/diagnostic-data/src/lib/components/changes-view/changes-view.component.ts
rekhaswaminathan: Yes Cindy, its the `rows` from `DataTableResponseObject`

cindylovescoding: But you are parsing a single row into this right Rekha?

AngularApp/projects/diagnostic-data/src/lib/components/changesets-view/changesets-view.component.ts
rekhaswaminathan: So this is a special case where we click on buttons from the UI and invoke getDetector API with additional query string parameters. What I noticed this, it was retrieving from cache because `this.detectorControlService.shouldRefresh` was always returning `false`. So explicitly passing `true`

hforeste: So we want to get fresh data each time?

AngularApp/projects/diagnostic-data/src/lib/components/changes-view/changes-view.component.ts
rekhaswaminathan: @cindylovescoding @SharmaAjay19 Sorry for the confusion guys. I was not clear before.
So `rows` is actually a json object which is sent part of the `DataTableResponseObject`. It has the properties needed to display in the UI. Also `initiatedByList` is actually a list of strings that contains the usernames for those initiated changes. 

image

So I'm parsing each row and `getInitiatedByUsers` determines if `initiatedByList` has any values

AngularApp/projects/diagnostic-data/src/lib/components/changesets-view/changesets-view.component.ts
rekhaswaminathan: Yes we want to do that each time to eliminate any possibility of getting cached data

PR #533, opened 2019-10-23
Enable support documents in analysis view for case submission flow

SharmaAjay19:


AngularApp/projects/app-service-diagnostics/src/app/shared-v2/services/resource.service.ts
```diff
@@ -38,6 +38,12 @@ export class ResourceService {
   }
 
   public getPesId(): Observable<string>{
+    if (this._genericArmConfigService){
```
nmallick1: Can be simplified by reusing the function armResourceConfig() function from within this file. Will save you one if condition.

PR #537, opened 2019-10-29
Stop gap aks mooncake

hforeste:
This is a temporary solution for AKS in Diagnose and Solve for China Cloud.

### DIFF
--- a/AngularApp/projects/app-service-diagnostics/src/app/home/components/home/home.component.ts
+++ b/AngularApp/projects/app-service-diagnostics/src/app/home/components/home/home.component.ts
@@ -62,6 +62,13 @@ export class HomeComponent implements OnInit {
       };
     }
 
+    if (this.cloudLocation == "portal.azure.cn"){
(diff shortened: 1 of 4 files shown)


AngularApp/projects/app-service-diagnostics/src/app/home/components/home/home.component.ts
```diff
@@ -58,6 +62,13 @@ export class HomeComponent implements OnInit {
       };
     }
 
+    if (this.cloudLocation == "portal.azure.cn"){
+      this.homePageText = {
+        title:'App Service Diagnostics',
```
rekhaswaminathan: So you're setting the title for App Service Diagnostics if its China cloud but the markdown seems to be having content for AKS?

hforeste: Diagnose and Solve is only available for AKS so right now China Cloud == AKS

PR #538, opened 2019-10-30
Enable support document in detector view in Case Submission Flow

SharmaAjay19:
image


AngularApp/projects/diagnostic-data/src/lib/components/detector-view/detector-view.component.ts
```diff
@@ -240,4 +248,25 @@ export class DetectorViewComponent implements OnInit {
     this.telemetryService.logEvent(eventMessage, eventProperties, measurements);
   }
 
+  populateSupportTopicDocument(){
```
hforeste: Are we planning to show self help content anywhere else? If so should we encapsulate this logic in some helper class or maybe a service? Other areas that I am thinking are support topics that doesn't have a detector assigned to it yet.

nmallick1: I believe for topics that do not have a detector assigned to them, our blade is not invoked by CloudES.

PR #539, opened 2019-10-30
Stop gap aks mooncake

hforeste:
- Fix staging slot. `window.parent.location.hostname` was giving DOM Exception due to CORS issue
- Fix markdown links to AKS documents to open new tab on click

### DIFF
--- a/AngularApp/projects/app-service-diagnostics/src/app/home/components/home/home.component.ts
+++ b/AngularApp/projects/app-service-diagnostics/src/app/home/components/home/home.component.ts
@@ -62,6 +62,13 @@ export class HomeComponent implements OnInit {
       };
     }
 
+    if (this.cloudLocation == "portal.azure.cn"){
+      this.homePageText = {
+        title:'App Service Diagnostics',
+        description: 'Explore ways to diagnose and troubleshoot the common problems of your cluster from CRUD operations to connection problems. Click on any of the documents below to start troubleshooting.',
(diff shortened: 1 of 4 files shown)


AngularApp/projects/app-service-diagnostics/src/app/shared/services/arm.service.ts
```diff
@@ -27,12 +27,13 @@ export class ArmService {
     }
 
     get armUrl(): string {
+        let browserUrl = (window.location != window.parent.location) ? document.referrer : document.location.href;
```
rekhaswaminathan: So `document.referrer` is the case where we load the app inside Portal and `document.location.href` is when we run locally?

AngularApp/projects/app-service-diagnostics/src/app/home/components/home/home.component.ts
```diff
@@ -62,7 +62,7 @@ export class HomeComponent implements OnInit {
       };
     }
 
-    if (this.cloudLocation == "portal.azure.cn"){
+    if (this.isPublicAzure == false){
```
rekhaswaminathan: nit: This can also be `!this.isPublicAzure`

AngularApp/projects/app-service-diagnostics/src/app/shared/services/arm.service.ts
hforeste: Right

PR #551, opened 2019-11-01
Cindy/highchartfix

cindylovescoding:
Fix several issues:
1. Increase the height
2. Non wrap for x-axis lables
3. Keyboard navigation issue for the plotting area
4. Add more ticks in y axis

### DIFF
--- a/AngularApp/projects/diagnostic-data/src/lib/components/highcharts-graph/highcharts-graph.component.ts
+++ b/AngularApp/projects/diagnostic-data/src/lib/components/highcharts-graph/highcharts-graph.component.ts
@@ -157,8 +157,10 @@ export class HighchartsGraphComponent implements OnInit {
             chart: {
                 reflow: false,
                 height: 200,
+                display: 'block!important',
                 type: 'line',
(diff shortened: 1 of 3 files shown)


cindylovescoding:
Screenshot:
image

AngularApp/projects/diagnostic-data/src/lib/components/highcharts-graph/highcharts-graph.component.ts
```diff
@@ -138,23 +138,35 @@ export class HighchartsGraphComponent implements OnInit {
...
+                    wrapAround: true,
+                    skipNullPoints: true
+                },
             },
             caption: {
                 text: `${this.chartDescription}`,
             },
             chart: {
-                reflow: true,
+                reflow: false,
```
SharmaAjay19: Should we set reflow to true, so the chart can expand and shrink with window size?

cindylovescoding: Good point! Revert it back!

PR #552, opened 2019-11-01
Filter left nav by item id and name both

SharmaAjay19:
image


michimune:
Thanks! This should help me a lot!

PR #555, opened 2019-11-06
Stop gap aks mooncake2

hforeste:
Get IE back up in running


AngularApp/projects/app-service-diagnostics/src/app/home/components/home/home.component.ts
```diff
@@ -39,6 +39,10 @@ export class HomeComponent implements OnInit {
       '';
   }
 
+  get isIE_Browser(): boolean {
+    return  /msie\s|trident\//i.test(window.navigator.userAgent);
```
rekhaswaminathan: Does this work for Chrome, Firefox, Edge? 😛

PR #556, opened 2019-11-06
Chartheight

cindylovescoding:
Fix tooltip overflow

### DIFF
--- a/AngularApp/projects/applens/src/styles.scss
+++ b/AngularApp/projects/applens/src/styles.scss
@@ -157,3 +157,16 @@ body {
   min-width: 70vw !important;
   min-height: 70vh !important;
 }
+
+/* highchart container configuration */
+.highcharts-container {
+    overflow: visible !important;
+  }
+
+.highcharts-container svg {
+    overflow: visible !important;
+}
+
+highcharts-chart {
+    overflow: visible!important;
+}
--- a/AngularApp/projects/diagnostic-data/src/lib/components/highcharts-graph/highcharts-graph.component.ts
+++ b/AngularApp/projects/diagnostic-data/src/lib/components/highcharts-graph/highcharts-graph.component.ts
@@ -200,7 +200,8 @@ export class HighchartsGraphComponent implements OnInit {
             tooltip: {
                 shared: true,
                 enabled: true,
-                valueDecimals: 2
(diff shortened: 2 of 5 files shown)


cindylovescoding:
Previously the tooltip gets cut because of the default highchart-container attribute:
image

Fix that in this PR:


image

AngularApp/projects/diagnostic-data/src/lib/components/highcharts-graph/highcharts-graph.component.ts
```diff
@@ -200,7 +200,8 @@ export class HighchartsGraphComponent implements OnInit {
             tooltip: {
                 shared: true,
                 enabled: true,
-                valueDecimals: 2
+                valueDecimals: 2,
+                useHTML: true,
```
SharmaAjay19: Can we add an opacity setting here to resolve this 
image

puneetg1983: +1 on this

cindylovescoding: the background above the axis is not coming through. Changing opacity won't work

AngularApp/projects/applens/src/styles.scss
```diff
@@ -157,3 +157,16 @@ body {
...
   min-height: 70vh !important;
 }
+
+/* highchart container configuration */
+.highcharts-container {
+    overflow: visible !important;
+  }
+
+.highcharts-container svg {
+    overflow: visible !important;
```
puneetg1983: Just wanted to confirm - will we get a vertical scrollbar if there are double the number of legends as shown in the screenshot or will the height of the legend tooltip keep increasing

ShekharGupta1988:
> Previously the tooltip gets cut because of the default highchart-container attribute:
> image
> 
> Fix that in this PR:
> 
> image

Very nice :)

AngularApp/projects/applens/src/styles.scss
cindylovescoding: I think it will keep increasing just like what we had before in nvd3.

PR #557, opened 2019-11-07
update spring cloud service logo, add config file for spring cloud service

Sneezry:


AngularApp/projects/app-service-diagnostics/src/armResourceConfig/microsoft.appplatform/spring/config.json
```diff
@@ -0,0 +1,24 @@
...
+	"matchRegEx": "/Microsoft.AppPlatform/Spring/",
+	"searchSuffix": "Spring",
+	"azureServiceName": "Azure Spring Cloud",
+	"armApiVersion": "2019-05-01-preview",
+	"isSearchEnabled": true,
+	"isApplicableForLiveChat": false,
+	"categories": [{
+			"id": "Troubleshooting",
+			"name": "Troubleshooting",
+			"description": "Explore the most common problems for your resource. Follow do-it-yourself troubleshooting steps, or explore a wide range of troubleshooting tools.",
```
yunjchoi: Are there plans to add detectors that will detect these common problems by running resource-specific checks? The current description sounds like the detectors will just show the same self-help docs, which is not the best use case of detectors.

AngularApp/projects/app-service-diagnostics/src/armResourceConfig/microsoft.appplatform/spring/config.json
```diff
@@ -0,0 +1,24 @@
...
+			"keywords": ["Failures", "Service Up"],
```
yunjchoi: Can we add more keywords? We've received multiple feedback that these keywords are helpful in setting the right expectation for the category.

PR #567, opened 2019-12-03
Ensure Storage Account names are always lowercase

puneetg1983:


AngularApp/projects/app-service-diagnostics/src/app/shared/components/daas/configure-storage-account/configure-storage-account.component.ts
```diff
@@ -63,7 +63,7 @@ export class ConfigureStorageAccountComponent implements OnInit {
   setDefaultValues() {
     this.chosenStorageAccount = this.storageAccounts[0];
     this.saveEnabled = true;
-    this.newStorageAccountName = this._siteService.currentSiteStatic.name.substring(0, 6) + this.randomHash();
+    this.newStorageAccountName = (this._siteService.currentSiteStatic.name.substring(0, 6) + this.randomHash()).toLowerCase();
```
hforeste: Storage API doesn't allow capitals in the storage account name? It gives a Bad Request I assume.

puneetg1983: Forgot the exact status code, I think was bad request only... but response had the good thing was that detailed error in response that tells you the issue 

`TestApt0yobh is not a valid storage account name. Storage account name must be between 3 and 24 characters in length and use numbers and lower-case letters only`

PR #568, opened 2019-12-04
Integrated Blade Diagnostics

nmallick1:
Provide an ability to link detectors from targeted Ibiza blades.. For example, the Ibiza blade would like look like below.

image

When someone clicks on Investigate Failure (or some other better text for the button), we load the detector targeted for operations initiated from that blade. In this case, we load the following

image

Have updated the mapping for SSL and backup detectors for now.


AngularApp/projects/app-service-diagnostics/src/app/shared/components/portal-referrer-resolver/portal-referrer-resolver.component.ts
```diff
@@ -0,0 +1,104 @@
...
+  public loadingMessage: string;
+  public isEvaluating: boolean;
+  public detectorId: string;
+
+  constructor(private _authService: AuthService, private _router: Router, private _activatedRoute: ActivatedRoute,
+    private _resourceService: ResourceService, private _logService: TelemetryService) {
+    this.isEvaluating = true;
+    this.loadingMessage = 'Collecting data to analyze your issue...';
+
+    if (this._resourceService instanceof WebSitesService) {
```
ShekharGupta1988: the reason you have Detector Id in this map instead of getting from Ibiza as a param is you want to save Ibiza deployment ?

nmallick1: Yes, that is correct.

ShekharGupta1988: My only concern is having to build this map for every resource here. you can offload this work to the resource service instead and let every resource define their own map

AngularApp/projects/app-service-diagnostics/src/app/shared/components/portal-referrer-resolver/portal-referrer-resolver.component.ts
```diff
@@ -0,0 +1,104 @@
...
+        path = `${path}/detectors/${referrerMatch.DetectorId}`;
+        this.detectorId = referrerMatch.DetectorId;
+      }
+      else {
+        if (referrerMatch.DetectorType === DetectorType.Analysis) {
+          path = `${path}/analysis/${referrerMatch.DetectorId}`;
+        }
+      }
+      this._logService.logEvent('IntegratedDiagnostics', {
+        resourceUri: this._resourceService.resource.id,
```
ShekharGupta1988: resource Uri is already getting logged as part of common columns in LogEvent

nmallick1: Cool !! Didn't realize it. Consider it not logged :)

AngularApp/projects/app-service-diagnostics/src/app/shared/components/portal-referrer-resolver/portal-referrer-resolver.component.ts
nmallick1: Good point, How about I do both ? If the information is passed, then I use that, else I look at the map. The reason I am suggesting this is because for Antares, it gives us flexibility and quick iteration. For partners, we can let them pass what they want and we just use it which saves us from deploying and gives the control to partners. What do you think ?

nmallick1: @ShekharGupta1988 Updates now. Please check

ShekharGupta1988:
Looks good 👍

PR #569, opened 2019-12-06
Change tooltip opacity and reflow behavior

cindylovescoding:
1. In highcharts, texts and labels are given in HTML but rendered as SVG. When useHTML both set to be true, tooltip will be laid below XAxis labels. Remove Xaxis "useHTML=true" and set "tooltip.outside" to prevent this behavior in Highcharts 6.1.1. 

image

2. Change select "All" and "None" behavior.
series.hide() and series.show() will try to reflow the chart. If there are too many series, UI will freeze.
However setVisible is only set the visible status, which works instantaneously.

3. Change the default height of linechart from "200px" to "300px".
image


puneetg1983:
much needed change! Thanks for fixing this

PR #572, opened 2019-12-10
Nmallick1/CXP chat api integration

nmallick1:
CXP chat integration. The flow will look as follows during case submission.

Chat button displayed at the end of the detector / analysis.
image

User is given an option to confirm the chat 

image

Chat is started in a new popup when the user clicks on Confirm & Start

image


AngularApp/projects/app-service-diagnostics/src/app/shared/services/cxp-chat.service.ts
```diff
@@ -0,0 +1,225 @@
...
+import { Verbs } from '../models/portal';
+import { Guid } from '../utilities/guid';
+import { TelemetryService, TelemetryEventNames } from 'diagnostic-data';
+
+@Injectable()
+export class CXPChatService {
+  public isChatSupported:boolean;
+  private supportedSupportTopicIds:string[];
+  public pesId: string = '';
+  public readonly cxpChatTagName: string = 'webapps';
```
ShekharGupta1988: I am not sure what role does this value is used but does it have it to be always 'webapps' irrespective of the product chat is opening for like linux, functions, aks etc

AngularApp/projects/app-service-diagnostics/src/app/shared/services/cxp-chat.service.ts
```diff
@@ -0,0 +1,225 @@
...
+
+    this._portalService.postMessage(Verbs.buildChatUrl, JSON.stringify(input));
+
+
+    //Wait for the response from the CXP chat API call.
+    return this._portalService.buildChatUrl().pipe(flatMap( (chatUrl) =>{
+
+      if(chatUrl && chatUrl != '') {
+        this._telemetryService.logEvent(TelemetryEventNames.BuildCXPChatUrl, {
+          "trackingId": trackingIdGuid,
```
ShekharGupta1988: is this the same as workflowId for case submission. If yes, its automatically getting logged with every event.

AngularApp/projects/app-service-diagnostics/src/app/shared/services/cxp-chat.service.ts
```diff
@@ -0,0 +1,225 @@
...
+          "returnValue": chatUrl
+        });
+
+        return Observable.of(chatUrl);
+      }
+      else {
+        if(chatUrl === '') {
+          this._telemetryService.logEvent(TelemetryEventNames.BuildCXPChatUrl, {
+            "trackingId": trackingIdGuid,
+            "passedInput" : JSON.stringify(input),
```
ShekharGupta1988: You can reduce the code little bit here but storing returnValue in a variable for case when chat url is empty and when its null. Then just log it once.

ShekharGupta1988:
@nmallick1 - Just a small comment on chat button positioning. I initially had freshchat button also on the right bottom hand side but in a week we noticed not many interactions with that button. We tried putting on left hand side and we saw increased interaction from customers that time. Something you can consider doing right from the beginning

AngularApp/projects/app-service-diagnostics/src/app/shared/services/cxp-chat.service.ts
nmallick1: yeah, its just a tag for them too be able to pull up queues configured for a support topic. I confirmed this with Wilson from CXP and he said using the same tag will work for other products too.

AngularApp/projects/app-service-diagnostics/src/app/shared/services/cxp-chat.service.ts
nmallick1: It's not the workflowId, it's a unique id that I am creating for every time a detector is loaded. Although it's not a bad idea to reuse the workflowId. Where do I get it from in our code ?

nmallick1: While I was making a change to use workflowId instead, I realized why I use a new tracking Id instead. When using Analysis, when someone navigates to a detector by clicking on More Info and then comes back, the component is recreated and a new call to portal is made. We need a unique tracking Id per call to that portal API and hence I am generating a new one and not reusing the workflowId.

cindylovescoding:
Hey @nmallick1, is this chat window another Iframe? How will this look like if there is a parent detector with too many child detector list?

nmallick1:
@cindylovescoding the chat window is actually a popup.

"Chat is started in a new popup when the user clicks on Confirm & Start"

ShekharGupta1988:
> Hey @nmallick1, is this chat window another Iframe? How will this look like if there is a parent detector with too many child detector list?

Hey @cindylovescoding  - I think this is a new browser window (small sized) that opens up when you click on that button.

AngularApp/projects/app-service-diagnostics/src/app/shared/services/cxp-chat.service.ts
ShekharGupta1988: got it. Thanks

nmallick1:
Chat button the left now

image


image

AngularApp/projects/app-service-diagnostics/src/app/shared/services/cxp-chat.service.ts
```diff
@@ -3,7 +3,7 @@ import { flatMap } from 'rxjs/operators';
 import { Observable, of, forkJoin, ReplaySubject } from 'rxjs';
 import { PortalService } from '../../startup/services/portal.service';
 import { ResourceService } from '../../shared-v2/services/resource.service';
-import { Verbs } from '../models/portal';
+import { Verbs, StartupInfo } from '../models/portal';
```
cindylovescoding: Why do we need this?

nmallick1: Good catch @cindylovescoding I was testing something and missed removing it. It's gone now.

nmallick1:
UI updated to look like below now

image


image

AngularApp/projects/app-service-diagnostics/src/app/app.module.ts
```diff
@@ -1,5 +1,5 @@
 import {
-    CommsService, DiagnosticDataModule, DiagnosticService, DiagnosticSiteService,
+    CXPChatService, CommsService, DiagnosticDataModule, DiagnosticService, DiagnosticSiteService,
```
ShekharGupta1988: you can remove this import

AngularApp/projects/app-service-diagnostics/src/app/home/home.module.ts
```diff
@@ -21,7 +21,9 @@ import { GenericAnalysisComponent } from '../shared/components/generic-analysis/
 import { DiagnosticsSettingsComponent } from './components/diagnostics-settings/diagnostics-settings.component';
 import { SupportTopicService } from '../shared-v2/services/support-topic.service';
 import { MarkdownModule } from 'ngx-markdown';
+import { CXPChatService } from 'diagnostic-data' ;//'../shared/services/cxp-chat.service';
```
ShekharGupta1988: nitpick : remove the commented part

AngularApp/projects/app-service-diagnostics/src/app/shared-v2/services/cxp-chat-caller.service.ts
```diff
@@ -0,0 +1,214 @@
...
+import { Guid } from '../../shared/utilities/guid';
+import { TelemetryService, TelemetryEventNames } from 'diagnostic-data';
+
+@Injectable()
+export class CXPChatCallerService {
+  public isChatSupported: boolean;
+  private supportedSupportTopicIds: string[];
+  public pesId: string = '';
+  public readonly cxpChatTagName: string = 'webapps';
+  public readonly applensUserAgentForCXPChat: string = 'applensDiagnostics';
```
rekhaswaminathan: Just out of curiosity, the caller is actually case submission from Portal right? Is this string fixed one ?

nmallick1: Yes, the cxpChatTagName this will be fixed. This will be used by the chat team select the queues that they open for us and then they will run a secondary filter on the support topic that we pass to them.

AngularApp/projects/app-service-diagnostics/src/app/shared-v2/services/cxp-chat-caller.service.ts
```diff
@@ -0,0 +1,214 @@
...
+        if (chatUrl === '') {
+          stringToLog = 'Empty URL returned. Likely cause, no engineer available.';
+        }
+        else {
+          stringToLog = 'NULL object returned. Likely cause, unknown. Followup with CXP team with trackingId.';
+        }
+      }
+
+      this._telemetryService.logEvent(TelemetryEventNames.GetCXPChatURL, {
+        "trackingId": trackingIdGuid,
```
ShekharGupta1988: You prob want to use a different name than 'trackingId' as this term is used as case workflowId in CloudES kusto in some tables. So just to avoid confusion, prob name this field as 'cxpChatTrackingId'

AngularApp/projects/app-service-diagnostics/src/app/shared-v2/services/cxp-chat-caller.service.ts
rekhaswaminathan: Oh no I meant to ask about `applensUserAgentForCXPChat` string.

nmallick1: Yes, the caller is actually case submission from portal. The user agent string will be used by them to identify us. They see us as the applens team and want to isolate that the call is coming from the applens team. case submission for them is everyone else. The string is what they asked us to put. We can always change it if required, but we'll need to communicate it to them. The string value is not used for any filtering / blocking but is used for their monitoring / metrics.

rekhaswaminathan: Nice. Thanks for the explanation!

PR #575, opened 2019-12-11
Adding a RadioButtonList as an input type

puneetg1983:
The UI change related to https://github.com/Azure/Azure-AppServices-Diagnostics/pull/365

image

### DIFF
--- a/AngularApp/projects/diagnostic-data/src/lib/components/form/form.component.ts
+++ b/AngularApp/projects/diagnostic-data/src/lib/components/form/form.component.ts
@@ -87,8 +87,7 @@ export class FormComponent extends DataRenderBaseComponent {
               formInputs[ip]["inputId"],
               formInputs[ip]["inputType"],
               formInputs[ip]["label"],
-              formInputs[ip]["isRequired"],
-              formInputs[ip]["items"]));
+              formInputs[ip]["isRequired"]));
           }
         }
       }
--- a/AngularApp/projects/diagnostic-data/src/lib/components/form/form.component.ts
+++ b/AngularApp/projects/diagnostic-data/src/lib/components/form/form.component.ts
@@ -1,7 +1,7 @@
 import { Component, Inject } from '@angular/core';
 import { DataRenderBaseComponent } from '../data-render-base/data-render-base.component';
 import { DiagnosticData, Rendering, DataTableResponseObject, DetectorResponse } from '../../models/detector';
-import { Form, FormInput, InputType, FormButton, ButtonStyles } from '../../models/form';
+import { Form, FormInput, InputType, FormButton, ButtonStyles, RadioButtonList } from '../../models/form';
 import { DiagnosticService } from '../../services/diagnostic.service';
 import { DetectorControlService } from '../../services/detector-control.service';
 import { Router, ActivatedRoute } from '@angular/router';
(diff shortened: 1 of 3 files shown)


AngularApp/projects/diagnostic-data/src/lib/components/form/form.component.ts
```diff
@@ -168,7 +183,7 @@ export class FormComponent extends DataRenderBaseComponent {
       let formToExecute = this.detectorForms.find(form => form.formId == detectorQueryParams.fId);
       let queryParams = `&fId=${detectorQueryParams.fId}&btnId=${detectorQueryParams.btnId}`;
       detectorQueryParams.inputs.forEach(ip => {
-        queryParams += `&inpId=${ip.inpId}&val=${ip.val}`;
+        queryParams += `&inpId=${ip.inpId}&val=${ip.val}&inpType=${ip.inpType}`;
```
ShekharGupta1988: Hey @puneetg1983 , do we want to send input type to backend. Since you are sending input id already, the author would be able to recognize the element based on id. 

One thing to be mindful here this is subjected to Url truncation.  @rekhaswaminathan did some calculations and but some limits on textbox size etc to be within the safe limit. I just worry if this can regress that.

AngularApp/projects/diagnostic-data/src/lib/models/form.ts
```diff
@@ -17,23 +17,38 @@ export class FormInput {
...
+        this.buttonStyle = buttonStyle != undefined ? buttonStyle : ButtonStyles.Primary;
+    }
+}
+
+export class RadioButtonList extends FormInput {
+    items: ListItem[] = [];
+    constructor(internalId: string, id: number, inputType: InputType, label: string, items: ListItem[]) {
+        super(internalId, id, inputType, label, false);
+        this.items = items;
+        items.forEach(x=>{
```
ShekharGupta1988: nit: fix spacing on the lines. just format the whole document

AngularApp/projects/diagnostic-data/src/lib/components/form/form.component.ts
puneetg1983: Maybe I don't understand the concept very well....

If you don't pass the inputType back to the server, how would the server know which type of control to create in FormModelBinder

https://github.com/Azure/Azure-AppServices-Diagnostics/blob/eb66ad962be1c3591c10fbc9570dda246e36d1e1/src/Diagnostics.RuntimeHost/Models/FormModelBinder.cs#L73

rekhaswaminathan: @puneetg1983 I think what Shekhar is referring to is FormModelBinder creates cxt.Form which detector author uses. FormModelBinder is mostly responsible for creating cxt.Form based on the current inputs selection. Let me know when you have few mins, I can quickly explain this to you

ShekharGupta1988: Ah I see. Makes sense. The value that gets passed as query string is a enum number or the string?
If it's a number, then we are adding almost 10 chars more for every input. 
@rekhaswaminathan - correct me here but I think we have a limit on number of inputs a particular form can have to be 5. Will this be fine and not get into URL truncation?

rekhaswaminathan: yes the limit of number of inputs is 5. Assuming its enum, we should be fine and won't run into URL truncation

puneetg1983: So am I ok to merge this or I need to change something?

ShekharGupta1988: Is the input type value an enum number or string that gets passed in the url?

puneetg1983: It is a number

PR #580, opened 2019-12-17
Disable features based on sub location placementid

rekhaswaminathan:


AngularApp/projects/app-service-diagnostics/src/app/shared/services/arm.service.ts
```diff
@@ -275,7 +278,11 @@ export class ArmService {
         return this._cache.get(url, request, invalidateCache);
     }
 
-    getHeaders(etag?: string): HttpHeaders {
+    getSubscriptionLocation(subscriptionId: string): Observable<HttpResponse<any>> {
+        return this.getResourceFullResponse<any>(`/subscriptions/${subscriptionId}`, false, '2019-06-01');
+    }
+
+    getHeaders(etag?: string, subscriptionLocation?: string): HttpHeaders {
```
ShekharGupta1988: Should this method instead of taking subscriptionLocation as param take a more generic structure like name value pair list. This will make it more expandable in future

AngularApp/projects/app-service-diagnostics/src/app/resources/web-sites/services/site-feature.service.ts
```diff
@@ -44,11 +46,23 @@ export class SiteFeatureService extends FeatureService {
...
       this.addProactiveTools(startupInfo.resourceId);
       this.addPremiumTools();
+      this.subscriptionId = startupInfo.resourceId.split("subscriptions/")[1].split("/")[0];
     });
   }
 
   sortFeatures() {
     let featureDisplayOrder = this._featureDisplayOrder;
+    let locationPlacementId = '';
+     this.armService.getResourceFullResponse<any>( `/subscriptions/${this.subscriptionId}`, false, '2019-06-01').subscribe(response => {
```
ShekharGupta1988: I see the same call done above as well. Its prob beneficial to create a SubscriptionService and encapsulate all these calls there. It would be more maintainable in future as well. Imagine if you want to change the versions or location value checks etc.

ShekharGupta1988:
Looks great. Thanks @rekhaswaminathan  for doing this change so quickly.

PR #581, opened 2019-12-18
Logging subscription location 

rekhaswaminathan:


AngularApp/projects/app-service-diagnostics/src/app/home/components/home/home.component.ts
```diff
@@ -127,24 +127,31 @@ export class HomeComponent implements OnInit {
     if (!this._detectorControlService.startTime) {
       this._detectorControlService.setDefault();
     }
+    let locationPlacementId =  '';
+    this.subscriptionPropertiesService.getSubscriptionProperties(this.subscriptionId).subscribe((response : HttpResponse<{}>) => {
+        let subscriptionProperties = response.body['subscriptionPolicies'];
+        locationPlacementId = subscriptionProperties['locationPlacementId'];
```
ShekharGupta1988: we should prob check if properties is empty or null here.

PR #582, opened 2019-12-19
Add new header for diagnostic call

hforeste:


AngularApp/projects/applens/src/app/modules/dashboard/dashboard/dashboard.component.ts
```diff
@@ -104,6 +104,7 @@ export class DashboardComponent implements OnDestroy {
         else if (serviceInputs.resourceType.toString() === 'Microsoft.Web/sites')
         {
             this._diagnosticApiService.GeomasterServiceAddress = this.resource["GeomasterServiceAddress"];
+            this._diagnosticApiService.GeoRegion = this.resource["WebSpace"];
```
ShekharGupta1988: Instead of calling this field as Georegion,  can we call this webspace only. Just to avoid confusion in future with geo region services.

hforeste: Added a commit to instead rename the field to `Location` as that is what I am intending to use it for. I think this should suffice for now.

PR #584, opened 2019-12-19
Stg1201

yiningwei:
Storage: AppLens: Set PesId as 15629 for storage account management


AngularApp/projects/applens/src/assets/enabledResourceTypes.json
```diff
@@ -176,5 +176,25 @@
             "pesId": "15614",
             "staticSelfHelpContent": ""
         }
+        {
```
nmallick1: This json is malformed. Kindly fix the JSON.

yiningwei: Fixed

PR #587, opened 2019-12-24
Making gist version history scrollable

SharmaAjay19:
Old experience in applens gist version history:
image
image

New experience:
image


cindylovescoding:
Thank you for fixing this Ajay! Is there anyway that we can also make the confirm and cancel buttons visible without scrolling to the bottom of the page?

SharmaAjay19:
> Thank you for fixing this Ajay! Is there anyway that we can also make the confirm and cancel buttons visible without scrolling to the bottom of the page?

Yup, that is what this fix is all about 👍. Earlier the confirm and cancel buttons were not visible without scrolling to the end of the page, now they are visible (see the third snapshot), because we have mad the upper div scrollable.

PR #588, opened 2019-12-24
First Commit App Insights enablement

puneetg1983:
When App Insights is enabled and Connected

image

When it is enabled only and not **connected**

image

Within a detector...
image


Backend/Services/Common/CommonService.cs
```diff
@@ -0,0 +1,37 @@
...
+using Microsoft.Extensions.Primitives;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+
+namespace Backend.Services
+{
+    public class CommonService : ICommonService
```
ShekharGupta1988: Does this needs to be a service? This prob can just be utility class

Backend/Services/AppInsightsService/AppInsightsService.cs
```diff
@@ -0,0 +1,156 @@
...
+        {
+            _encryptionService = encryptionService;
+        }
+
+        private async Task CheckAndRemoveExistingApiKeysAsync(string appInsightsResourceId, string bearerToken)
+        {
+            HttpRequestMessage request = new HttpRequestMessage
+            {
+                Method = HttpMethod.Get,
+                RequestUri = new Uri($"https://management.azure.com{appInsightsResourceId}/ApiKeys?api-version=2015-05-01"),
```
ShekharGupta1988: Have this endpoint configurable via app settings as this would be different for N-Clouds

Backend/Controllers/AppInsightsController.cs
```diff
@@ -0,0 +1,75 @@
...
+        private readonly IAppInsightsService _appInsightsService;
+        private readonly ICommonService _commonService;
+
+        public AppInsightsController(IArmService armService, IAppInsightsService appInsightsService, ICommonService commonService)
+        {
+            _armService = armService;
+            _appInsightsService = appInsightsService;
+            _commonService = commonService;
+        }
+        [HttpGet]
```
ShekharGupta1988: This should be a POST call instead of GET. If the operation is idempotent i.e. you'll generate key if already present, then consider using PUT

Backend/Services/Common/CommonService.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        public bool GetHeaderValue(IHeaderDictionary headers, string headerName, out string headerValue)
```
ShekharGupta1988: nitpick : just following the naming convention of system lib, consider using TryGetHeaderValue as the name of the method.

Backend/Services/Common/CommonService.cs
puneetg1983: Converted to utility class.

Backend/Services/AppInsightsService/AppInsightsService.cs
puneetg1983: Good point! Done.

Backend/Controllers/AppInsightsController.cs
puneetg1983: Changed to PUT

Backend/Services/Common/CommonService.cs
puneetg1983: Changed to TryGetHeaderValue

AngularApp/projects/diagnostic-data/src/lib/components/app-insights-enablement/app-insights-enablement.component.html
```diff
@@ -0,0 +1,53 @@
+<div *ngIf="isEnabledInProd" class="animated fadeIn panel panel-info width-90">
+  <div class="banner">
```
ShekharGupta1988: We would prob need to rework this component in Fabric Library later, so that the look and feel is consistent with the new redesign.

puneetg1983: Yup, totally agreed. This is just temporary UI

PR #590, opened 2020-01-03
Adding header x-ms-diagversion for calls to ARM

rekhaswaminathan:
Adding this header to ARM calls which GeoMaster will consume for the following APIs:

- ListDetectors
- GetDetector
- GetDiagnosticsProperties

### DIFF
--- a/AngularApp/projects/app-service-diagnostics/src/app/shared/services/arm.service.ts
+++ b/AngularApp/projects/app-service-diagnostics/src/app/shared/services/arm.service.ts
@@ -73,6 +73,7 @@ export class ArmService {
         });
         let additionalHeaders = new Map<string, string>();
         additionalHeaders.set('x-ms-subscription-location-placementid', subscriptionLocation);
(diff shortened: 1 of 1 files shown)


AngularApp/projects/app-service-diagnostics/src/app/shared/services/arm.service.ts
```diff
@@ -73,6 +73,7 @@ export class ArmService {
         });
         let additionalHeaders = new Map<string, string>();
         additionalHeaders.set('x-ms-subscription-location-placementid', subscriptionLocation);
+        additionalHeaders.set('x-ms-diagversion', '1');
```
ShekharGupta1988: nitpick : can you add a class variable at the top and its value set to 1. Also good to a comment explaining what 0 is and what 1 is.

rekhaswaminathan: Sure I will do that

PR #600, opened 2020-01-21
Add VnetName and LinuxFxVersion to the resource info window

michimune:


AngularApp/projects/applens/src/app/modules/dashboard/dashboard/dashboard.component.ts
```diff
@@ -148,6 +149,20 @@ export class DashboardComponent implements OnDestroy {
   }
 
   openResourceInfoModal() {
+    if (this.keys.indexOf('VnetName') == -1)
```
dagrooms52: I'm wondering if there is a way to queue this work when the service starts up, instead of starting the work only when the user requests the Details view. It may be unclear that more info is loading when we open Details.

ApplensBackend/Services/ObserverClientService/DiagnosticObserverClientService.cs
```diff
@@ -54,7 +54,7 @@ public Task<ObserverResponse> GetSite(string siteName)
             return GetSiteInternal(null, siteName);
         }
 
-        public Task<ObserverResponse> GetSite(string stamp, string siteName)
+        public Task<ObserverResponse> GetSite(string stamp, string siteName, bool details)
```
dagrooms52: Can we default `details` to `false` to allow the previous simple usage of this API?

michimune: Got it. I will update it.

AngularApp/projects/applens/src/app/modules/dashboard/dashboard/dashboard.component.ts
michimune: I discussed this with Shekhar and he says it's a good compromise to keep balance of the number of extra API calls and AppLens' start up time. We could always revisit later, though.

ShekharGupta1988: Michi - will there be a small spinning icon next to VNET Name in modal while this call is finishing up?

AngularApp/projects/applens/src/app/modules/dashboard/dashboard/dashboard.component.ts
```diff
@@ -148,6 +149,20 @@ export class DashboardComponent implements OnDestroy {
...
+    {
+      this._observerService.getSiteRequestDetails(this.resource.SiteName, this.resource.StampName).subscribe(siteInfo => {
+        console.log('siteInfo: sitename = ' + siteInfo.siteName);
```
ShekharGupta1988: get rid of console logs before merging the PR

michimune: Got it. Updated

AngularApp/projects/applens/src/app/modules/dashboard/dashboard/dashboard.component.ts
michimune: Well as far as I tried, the VNET name shows up in less than a second, so I don't think we need an extra UI for it. But, hmm, if the API takes time for some reason we need it. I will think of it.

PR #601, opened 2020-01-22
Cognitive search component for better customer experience

SharmaAjay19:
Added a web-search component that has below features:
1. Is embeddable in both external UI and applens.
2. Uses the context of the azure resource to search bing.
3. Uses the context of app stack type to search bing.
4. Maximum number of results to be displayed can be specified.
5. Is deep-linkable (i.e. can communicate the search term using route parameters).
Screenshots below:
In Diagnose and Solve
image

Case submission flow
image

AppLens
image


AngularApp/projects/app-service-diagnostics/src/app/shared-v2/services/content.service.ts
```diff
@@ -46,25 +45,20 @@ export class ContentService {
   searchWeb(questionString: string, resultsCount: string = '3'): Observable<any> {
 
     const searchSuffix = this._resourceService.searchSuffix;
-    const query = encodeURIComponent(`${questionString} AND ${searchSuffix}`);
+    var stackTypeSuffix = this._webSiteService["appStack"]? ` ${this._webSiteService["appStack"]}`: "";
```
ShekharGupta1988: One thing to check here Ajay - can the stack value be Unknown,None, Static or NoTraffic for new/inactive sites for which analytics either doesnt have data or there is no traffic

AngularApp/projects/applens/src/assets/enabledResourceTypes.json
```diff
@@ -18,7 +19,8 @@
             "versionPrefix": "v2/",
             "azureCommImpactedServicesList": "appservice",
             "pesId": "16533",
-            "staticSelfHelpContent": "microsoft.ase"
+            "staticSelfHelpContent": "microsoft.ase",
+            "searchSuffix": ""
```
ShekharGupta1988: App Service Environment?

SharmaAjay19: Thanks, wasn't sure about the exact public string!

AngularApp/projects/app-service-diagnostics/src/app/shared-v2/services/content.service.ts
SharmaAjay19: Will check all the distinct values it can hold and take appropriate decision here accordingly.

SharmaAjay19: I talked to Nazeef regarding this and all other insignificant values are already handled by the app stack query except the "Static only". For that I added a handling here.

PR #603, opened 2020-01-24
Monitor SCM process in proactive cpu monitoring

puneetg1983:
- Adding support to monitor Scm processes and
- increasing the default time to 2 weeks instead of one.

image


AngularApp/projects/app-service-diagnostics/src/app/shared/components/daas/cpu-monitoring/cpu-monitoring-configuration/cpu-monitoring-configuration.component.ts
```diff
@@ -164,6 +164,10 @@ export class CpuMonitoringConfigurationComponent implements OnInit, OnChanges {
     }
 
     this.ruleSummary += ` Monitoring will stop automatically after <b>${this.monitoringSession.MaximumNumberOfHours > 24 ? (this.getValueRounded(this.monitoringSession.MaximumNumberOfHours / 24)) + " days" : this.monitoringSession.MaximumNumberOfHours + " hours"}</b>.`;
+
+    if (this.monitoringSession.MonitorScmProcesses){
+      this.ruleSummary += " The worker process for the kudu site and all the webjobs under this app will also be monitored."
```
hforeste: nitpik: I don't think you have to say `all the webjobs`, you can just say `the webjobs`. When I read it with just `the webjobs`, it kind of already refers to all of them.

PR #605, opened 2020-01-26
Temporary access provision for AppLens

SharmaAjay19:
Temporary Access Provision for AppLens. It defaults to 7 days and can be availed one time only.
Screenshots:
When user does not have access to applens
image

When temporary access is approved
image

And user is redirected to home page
image

When user has already used one-time temporary access
image


ShekharGupta1988:
@SharmaAjay19 - Awesome .. This was a super fast implementation :)

PR #607, opened 2020-01-27
Temporary access provision for AppLens

SharmaAjay19:
Temporary Access Provision for AppLens. It defaults to 7 days and can be availed one time only.
Screenshots:
When user does not have access to applens
image

When temporary access is approved
image

And user is redirected to home page
image

When user has already used one-time temporary access
image

### DIFF
--- a/ApplensBackend/Controllers/TemporaryAccessController.cs
+++ b/ApplensBackend/Controllers/TemporaryAccessController.cs
@@ -14,13 +14,15 @@ namespace AppLensV3.Controllers
     public class TemporaryAccessController : Controller
     {
         ICosmosDBHandler<TemporaryAccessUser> _cosmosDBHandler;
-        int temporaryAccessDays = 7;
+        long temporaryAccessDurationInSeconds = 7 * 24 * 60 * 60;
 
         public TemporaryAccessController(ICosmosDBHandler<TemporaryAccessUser> cosmosDBHandler, IConfiguration configuration)
         {
             _cosmosDBHandler = cosmosDBHandler;
             var accessDurationInDays = configuration["ApplensTemporaryAccess:AccessDurationInDays"];
+            int temporaryAccessDays = 7;
             int.TryParse(accessDurationInDays.ToString(), out temporaryAccessDays);
+            temporaryAccessDurationInSeconds = temporaryAccessDays * 24 * 60 * 60;
         }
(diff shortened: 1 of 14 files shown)


ApplensBackend/Controllers/TemporaryAccessController.cs
```diff
@@ -0,0 +1,70 @@
...
+                if (userId != null)
+                {
+                    var result = await _cosmosDBHandler.GetItemAsync(userId);
+                    if (result == null)
+                    {
+                        TemporaryAccessUser newUser = new TemporaryAccessUser()
+                        {
+                            Id = userId,
+                            AccessStartDate = DateTime.UtcNow,
+                            UserPrincipalName = userId
```
rekhaswaminathan: `Id` and `UserPrincipalName` are taking same value `userId`. Do we need both?

ApplensBackend/Services/CosmosDBHandler/CosmosDBHandler.cs
```diff
@@ -0,0 +1,111 @@
...
+        private string Key;
+        private string DatabaseId;
+        private string CollectionId;
+        private string PartitionKey = "/UserPrincipalName";
+        private DocumentClient client;
+
+        public CosmosDBHandler(IConfiguration configuration)
+        {
+            Endpoint = configuration["ApplensTemporaryAccess:Endpoint"];
+            Key = configuration["ApplensTemporaryAccess:Key"];
```
rekhaswaminathan: This `Key` is actually meant for connecting to CosmosDB right? Should we rename to it something like "AppLensCosmosDB". Same for `ApplensTemporaryAccess:Endpoint`

ApplensBackend/Controllers/TemporaryAccessController.cs
SharmaAjay19: Yes, we need a partition key as well as unique id in cosmos db.

ApplensBackend/Services/CosmosDBHandler/CosmosDBHandler.cs
SharmaAjay19: Yes we will be using this terminology for the detectors cosmos. Here the sole purpose of the handler is for temporary access hence the name.

PR #609, opened 2020-01-28
Fix Publishing Email and Small Code Improvements

ShekharGupta1988:
Changes:
- Fix in the Detector Publishing Email to include the correct alias of the person who modified the detector
- Small code refactoring to make headers code more generic and scalable.


AngularApp/package-lock.json
```diff
@@ -11,7 +11,7 @@
       "dev": true,
```
puneetg1983: Do you really want to commit this file or this got added accidentally?

PR #610, opened 2020-01-28
Use LoggedIn User alias for Publishing Email in case Modified Header is not present

ShekharGupta1988:


ApplensBackend/Controllers/DiagnosticController.cs
```diff
@@ -168,11 +169,25 @@ private static string GetHeaderOrDefault(IHeaderDictionary headers, string heade
 
         private InvokeHeaders ProcessInvokeHeaders()
         {
+            var authorization = Request.Headers["Authorization"].ToString();
+            string userId = string.Empty;
+
+            if (string.IsNullOrWhiteSpace(authorization))
```
SharmaAjay19: Maybe we want to use !string.IsNullOrWhiteSpace(authorization) here?

ShekharGupta1988: Good catch.. forgot to commit this change. Added

PR #611, opened 2020-01-31
Temporary Access expiry alert on Applens

SharmaAjay19:
Users who are running on temporary access will receive a notification that informs them about how many days are remaining till their access expires. The notification also contains a direct link to apply for AppLens Access.
Home page:
image

Applens Dashboard:
image

View upon scrolling down in a detector, the banner still shows
image

They can hit close and ignore the notification, it will reappear the next time they open AppLens (if their access is still temporary).
When they hit close button:
image

### DIFF
--- a/ApplensBackend/Controllers/TemporaryAccessController.cs
+++ b/ApplensBackend/Controllers/TemporaryAccessController.cs
@@ -25,7 +25,7 @@ namespace AppLensV3.Controllers
             temporaryAccessDurationInSeconds = temporaryAccessDays * 24 * 60 * 60;
         }
 
-        [HttpGet("requestAccess")]
+        [HttpPost("requestAccess")]
         public async Task<IActionResult> RequestAccess()
         {
             string userId = null;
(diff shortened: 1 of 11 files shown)


khaled-zayed:
❤

ApplensBackend/Controllers/TemporaryAccessController.cs
```diff
@@ -25,7 +25,7 @@ public TemporaryAccessController(ICosmosDBHandler<TemporaryAccessUser> cosmosDBH
             temporaryAccessDurationInSeconds = temporaryAccessDays * 24 * 60 * 60;
         }
 
-        [HttpGet("requestAccess")]
+        [HttpPost("requestAccess")]
```
hforeste: Why did you need to make this a post?

SharmaAjay19: Because we are doing a write operation in backend for this. We write a document to cosmos db.

PR #612, opened 2020-01-31
Fix AppLens to load for sites with non-ascii chars

puneetg1983:
The actual issue why it was failing was this exception

`Error while sending request: Failed to execute setRequestHeader on XMLHttpRequest: Value is not a valid ByteString.`


hforeste:
Can you put an example URI where it gave this exception?

puneetg1983:
Hey Hawk, details are in https://dev.azure.com/app-service-diagnostics-portal/app-service-diagnostics-portal/_workitems/edit/1044/

PR #616, opened 2020-02-07
SearchComponent: Cognitive search optimization and Detector search updates

SharmaAjay19:
Changes in cognitive search:
1. Use only particular types of stacks.
2. Show cognitive search at most 10 seconds into the rendering of the search component - and not wait for all detectors to execute.

Changes in detector search:
1. Do not run detectors that are already rendered in parent analysis.

No changes in the UI so no screenshots.

### DIFF
--- a/AngularApp/projects/diagnostic-data/src/lib/components/detector-list-analysis/detector-list-analysis.component.ts
+++ b/AngularApp/projects/diagnostic-data/src/lib/components/detector-list-analysis/detector-list-analysis.component.ts
@@ -83,6 +83,8 @@ export class DetectorListAnalysisComponent extends DataRenderBaseComponent imple
     showPreLoadingError: boolean = false;
     isSearchEmbedded: boolean = false;
     showSuccessfulChecks: boolean = true;
+    showWebSearch: boolean = false;
+    showWebSearchTimeout: any = null;
 
     constructor(private _activatedRoute: ActivatedRoute, private _router: Router,
         private _diagnosticService: DiagnosticService, private _detectorControl: DetectorControlService,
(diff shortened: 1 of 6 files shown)


AngularApp/projects/diagnostic-data/src/lib/components/detector-list-analysis/detector-list-analysis.component.ts
```diff
@@ -312,6 +314,10 @@ export class DetectorListAnalysisComponent extends DataRenderBaseComponent imple
     }
 
     startDetectorRendering(detectorList) {
+        if (this.showWebSearchTimeout){
+            clearTimeout(this.showWebSearchTimeout);
+        }
+        this.showWebSearchTimeout = setTimeout(() => {this.showWebSearch = true;}, 10000);
```
ShekharGupta1988: So we will wait for 10 seconds to show web results, is it? it seems pretty long wait specially in case submission.
Should this wait time be dynamically calculated based on search API response time. So scenarios where no detector was found in search, cognitive results can show instantaneously instead of waiting for 10 seconds

SharmaAjay19: If there are no search results then congitive search will immediately fire. This logic does not make the cognitive search to wait for 10 seconds. In the earlier implementation the cognitive search had to wait until all the detector run completely. So now if it's 10 seconds and detectors are still running, cognitive search will show up.

SharmaAjay19: Cognitive search will show up in "atmost" 10 seconds.

PR #617, opened 2020-02-07
Add copy-to-clipboard buttons to Resource Infomation dialog

michimune:
This PR adds a copy-to-clipboard button for each property in the Resource Information dialog box.


AngularApp/projects/applens/src/app/modules/dashboard/dashboard/dashboard.component.ts
```diff
@@ -169,6 +169,23 @@ export class DashboardComponent implements OnDestroy {
     this.ngxSmartModalService.getModal('resourceInfoModal').open();
   }
 
+  copyToClipboard(item, event) {
+    //console.log(event.target);
```
cindylovescoding: Could you remove the comment:)

PR #619, opened 2020-02-10
Highlight the analytics row when hovering

michimune:


puneetg1983:
@michimune - Did you mean to say highlight the **analysis** row on hover?

PR #620, opened 2020-02-11
Diagnose And Solve for fairfax

nmallick1:
Enable Diagnose And Solve for Fairfax. The UI will appear as follows.

image

### DIFF
--- a/AngularApp/projects/app-service-diagnostics/src/app/home/components/home/home.component.html
+++ b/AngularApp/projects/app-service-diagnostics/src/app/home/components/home/home.component.html
@@ -1,6 +1,5 @@
 <div class="home-container">
-  <div class="input-group input-group-lg search-input-group" (keyup.escape)="clearSearch()" role="search" 
-  *ngIf="this.armService.isPublicAzure || this._resourceService.parseResourceUri(this._resourceService.resourceIdForRouting).provider.toLowerCase() != 'microsoft.containerservice'">
+  <div class="input-group input-group-lg search-input-group" (keyup.escape)="clearSearch()" role="search" *ngIf="!isAKSOnNationalCloud">
     <input tabindex="0" type="search" class="form-control" placeholder="{{searchPlaceHolder}}"
       [ngModel]="searchValue" (ngModelChange)="updateSearchValue($event)" (focus)="onSearchBoxFocus(event)"
       (focusout)="onSearchLostFocus()" aria-controls="search-results">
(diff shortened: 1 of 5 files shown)


AngularApp/projects/app-service-diagnostics/src/app/shared/services/arm.service.ts
```diff
@@ -21,23 +22,59 @@ export class ArmService {
...
+        return this.isMooncake || this.isFairfax || this.isBlackforest || this.isUsnat;
+    }
+
+    get armUrl(): string {
+        if(this.armEndpoint !='' && this.armEndpoint.length > 1 ) {
+            return  this.armEndpoint;
         }
+        else {
+            let browserUrl = (window.location != window.parent.location) ? document.referrer : document.location.href;
+            let armUrl = this.publicAzureArmUrl;
```
rekhaswaminathan: So this will work in both cases where diagnose and solve is loaded within portal or as separate project (localhost:3000)?

AngularApp/projects/app-service-diagnostics/src/app/home/components/home/home.component.html
```diff
@@ -1,5 +1,6 @@
 <div class="home-container">
-  <div class="input-group input-group-lg search-input-group" (keyup.escape)="clearSearch()" role="search" *ngIf="isPublicAzure || isIE_Browser">
+  <div class="input-group input-group-lg search-input-group" (keyup.escape)="clearSearch()" role="search" 
+  *ngIf="this.armService.isPublicAzure || this._resourceService.parseResourceUri(this._resourceService.resourceIdForRouting).provider.toLowerCase() != 'microsoft.containerservice'">
```
rekhaswaminathan: `this.armService.isPublicAzure || this._resourceService.parseResourceUri(this._resourceService.resourceIdForRouting).provider.toLowerCase() != 'microsoft.containerservice`

Should we make this a property in the component? Seems like it is repeated 3 times in the same file.

AngularApp/projects/app-service-diagnostics/src/app/shared/services/arm.service.ts
nmallick1: Yes, it should. We will use the value supplied by Ibiza only if present else default to what our existing behavior has been. Till now we have been ignoring the value sent to us by Ibiza. We were fine so far since we were always public cloud but now we'll have to depend on the Ibiza supplied value.

PR #621, opened 2020-02-11
Making the table that shows DaaS sessions keyboard accessible

puneetg1983:


AngularApp/projects/app-service-diagnostics/src/app/shared/components/daas-sessions/daas-sessions.component.html
```diff
@@ -77,14 +77,14 @@
                         <div class="inline tool-tip hotspot-clickable" style="font-size:medium">
                             <span style="margin-left:30px;width:250px;margin-top:30px" class="tool-tip-text">Click here
                                 to delete this session and any collected logs and reports generated for this session</span>
-                            <i class="fa fa-trash" aria-hidden="true" (click)="deleteSession(i)"></i>
+                            <i tabindex="0" class="fa fa-trash" aria-hidden="true" (click)="deleteSession(i)" (keypress)="deleteSession(i)"></i>
```
cindylovescoding: Do we want to specify which key will trigger keypress event?

puneetg1983: how are we doing at other places? We are looking for a specific key?

puneetg1983: ah found it, we are using (keyup.enter), will update the PR..Thanks

PR #623, opened 2020-02-11
Point National Cloud requests to liberation

nmallick1:
Point requests from National cloud to liberation and bypass the diag role.


AngularApp/projects/app-service-diagnostics/src/app/shared/services/arm.service.ts
```diff
@@ -71,6 +71,10 @@ export class ArmService {
             }
             else if(browserUrl.includes("azure.us")){
                 armUrl = this.usGovernmentAzureArmUrl;
+            } else if(browserUrl.includes("azure.cn")) {
```
hforeste: It looks like we have two conditions for china cloud

PR #626, opened 2020-02-12
Configure kusto mapping

hforeste:
- Add UI for Kusto Mapping Configuration
- Wire up backend to the kusto mappings configuration API

### DIFF
--- a/ApplensBackend/Controllers/GithubController.cs
+++ b/ApplensBackend/Controllers/GithubController.cs
@@ -46,6 +46,19 @@ namespace AppLensV3.Controllers
         }
 
         /// <summary>
+        /// Get template.
+        /// </summary>
+        /// <param name="name">File name.</param>
+        /// <param name="fileExtension">File extension.</param>
+        /// <returns>Task for getting template.</returns>
+        [HttpGet("template/{name}/{fileExtension}")]
+        public async Task<IActionResult> GetTemplate(string name, string fileExtension)
+        {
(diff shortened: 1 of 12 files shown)


ApplensBackend/Controllers/GithubController.cs
```diff
@@ -45,6 +45,19 @@ public async Task<IActionResult> GetTemplate(string name)
...
+        /// <summary>
+        /// Get template.
+        /// </summary>
+        /// <param name="name">File name.</param>
+        /// <param name="fileExtension">File extension.</param>
+        /// <returns>Task for getting template.</returns>
+        [HttpGet("template/{name}/{fileExtension}")]
+        public async Task<IActionResult> GetTemplate(string name, string fileExtension)
+        {
+            string content = await GithubService.GetRawFile(GithubConstants.TemplatePath.Replace("{filename}", name).Replace("csx", fileExtension));
```
SharmaAjay19: Will the config file be present under data/templates folder?

SharmaAjay19: Also the template path points to the master branch, but our latest and up to date data is in master2. I would rather recommend creating a new constant like
internal const string ConfigPathFormat = "https://api.github.com/repos/{0}/{1}/contents/{2}/package.json?ref={3}";

This takes care of the branch.

hforeste: Config is under data/templates.

PR #627, opened 2020-02-12
Fix issue: Applens editor hangs on empty code

SharmaAjay19:
Applens editor hangs when the code is emptied. This happens for detectors/gists/configuration files.
Repro steps:
Open applens and open any detector for editing or go to create new detector
image

Then empty the editor, the editor screen hangs indefinitely
image

The fix:
We use a boolean for rendering the monaco code editor instead of binding it directly to "code" variable. We set the boolean once the pre-development requests are complete.
After the fix, upon emptying the editor one can have the view
image


rekhaswaminathan:
Thank you for fixing this!

PR #630, opened 2020-02-13
Nmallick1/Disable fresh chat for non Public subs

nmallick1:
Disable fresh chat for non public subs


AngularApp/projects/app-service-diagnostics/src/app/shared-v2/services/livechat.service.ts
```diff
@@ -32,27 +33,32 @@ export class LiveChatService {
...
+                                                            supportTopicId: startupInfo.supportTopicId,
+                                                            product: this._resourceService.azureServiceName
+                                                        });
+                                                    }, 1000);
+                                                }
+                                            }));
+                                        }
+                                        else {
+                                            this.logger.LogLiveChatWidgetSkipped('Jedi subscription');
+                                        }
```
rekhaswaminathan: I believe we should remove this string also. We can log something like "Chat not supported for this subcription"

rekhaswaminathan:
@nmallick1 Just a minor comment on the log message. Other things look good. I'm approving this.

PR #638, opened 2020-02-14
Make Dynamic Insight Consistent With Regular Insights

dagrooms52:
- Status icon and rating are the same on insights and dynamic insights:

dynamic-insight-style

- Changed column header to indicate sort, used input text box placeholder as column name:

column-header-style


AngularApp/projects/diagnostic-data/src/lib/components/data-table/data-table.component.ts
```diff
@@ -125,4 +125,8 @@ export class DataTableComponent extends DataRenderBaseComponent implements After
 
   }
 
+  onInputClicked(event: any) {
+    event.stopPropagation();
```
cindylovescoding: What's this event about?

dagrooms52: Each column header is an `<input>` box, but clicking on that input box also triggers a click on the containing `<div>` which would call `sort()` for the column. This blocks the click event from going out to the div and sorting when you just want to type a filter.

PR #639, opened 2020-02-17
Improve AI detection logic when Portal RPC service returns nothing

puneetg1983:
When this portal service fails to find the tag for App Insights resource, we can get appsettings for the app and reverse lookup all the AI resources in the subscription to find out the AI resource linked for the app.

### DIFF
--- a/AngularApp/projects/diagnostic-data/src/lib/components/app-insights-enablement/app-insights-enablement.component.ts
+++ b/AngularApp/projects/diagnostic-data/src/lib/components/app-insights-enablement/app-insights-enablement.component.ts
@@ -26,7 +26,7 @@ export class AppInsightsEnablementComponent implements OnInit {
   connecting: boolean = false;
   error: any;
 
-  isEnabledInProd: boolean = false;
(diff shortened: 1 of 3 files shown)


AngularApp/projects/diagnostic-data/src/lib/components/app-insights-enablement/app-insights-enablement.component.ts
```diff
@@ -26,7 +26,7 @@ export class AppInsightsEnablementComponent implements OnInit {
   connecting: boolean = false;
   error: any;
 
-  isEnabledInProd: boolean = false;
+  isEnabledInProd: boolean = true;
```
cindylovescoding: Why changing the default value?

puneetg1983: Changing this enables or disables this component in the UI. I disabled it last week to ensure that this component doesn't even show up and with this PR, I am enabling it back.

PR #645, opened 2020-02-24
Remove Row Limit For DataTable

dagrooms52:
- Row limit was previously worked around with `TableOptions = new { scrollbarV = true}` in detector rendering properties
- Setting a height limit on the table and removing row limit automatically adds the vertical scroll bar, achieving the same effect


ShekharGupta1988:
Hey @dagrooms52  - I see this PR active for 15 days. Mind checking if this is needed and merge/abort it accordingly.

PR #647, opened 2020-02-26
DAAS Icons and Wording update

puneetg1983:


nmallick1:
Suggestion, when the profiler has been started, thenicon turn green but the text still reads starting profiler. Can we change the text too ?

puneetg1983:
> Suggestion, when the profiler has been started, thenicon turn green but the text still reads starting profiler. Can we change the text too ?

done

PR #661, opened 2020-03-09
Changing to AppInsights Rendering based on DataTable

puneetg1983:
This is the frontend code change request w.r.t https://github.com/Azure/Azure-AppServices-Diagnostics/pull/439

### DIFF
--- a/AngularApp/projects/diagnostic-data/src/lib/components/app-insights-markdown/app-insights-markdown.component.ts
+++ b/AngularApp/projects/diagnostic-data/src/lib/components/app-insights-markdown/app-insights-markdown.component.ts
@@ -40,6 +40,7 @@ export class AppInsightsMarkdownComponent extends DataRenderBaseComponent {
   diagnosticDataSet: DiagnosticData[] = [];
   loadingAppInsightsResource: boolean = true;
   loadingAppInsightsQueryData: boolean = true;
+  showSectionHeader: boolean = false;
 
   constructor(public _appInsightsService: AppInsightsQueryService, private _diagnosticService: DiagnosticService, private _router: Router,
     private _activatedRoute: ActivatedRoute, protected telemetryService: TelemetryService, private _navigator: FeatureNavigationService, @Inject(DIAGNOSTIC_DATA_CONFIG) config: DiagnosticDataConfig) {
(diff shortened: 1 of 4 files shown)


AngularApp/projects/diagnostic-data/src/lib/models/app-insights.ts
```diff
@@ -1,17 +1,17 @@
...
     query: string;
     poralBladeInfo: BladeInfo;
     renderingProperties: any;
+    dataTable: DataTableResponseObject;
   }
   
   export class AppInsightData {
     title: string;
     description: string;
-    table: any;
```
cindylovescoding: Don't remember why this was added at the first place..

AngularApp/projects/diagnostic-data/src/lib/components/app-insights-markdown/app-insights-markdown.component.ts
```diff
@@ -70,55 +71,76 @@ export class AppInsightsMarkdownComponent extends DataRenderBaseComponent {
...
-                    columnName: column.ColumnName,
-                    dataType: column.DataType,
-                    columnType: column.ColumnType,
-                  }
+      this.appInsightQueryMetaDataList.forEach(appInsightData => {
 
-                  )
-
-                });
+        if (appInsightData.dataTable !== null) {
```
cindylovescoding: So this table will take precedence over the table result from the passed in query, right?

AngularApp/projects/diagnostic-data/src/lib/models/app-insights.ts
puneetg1983: Yea I tried finding this but found no references to this property

AngularApp/projects/diagnostic-data/src/lib/components/app-insights-markdown/app-insights-markdown.component.ts
puneetg1983: Yes if the detector is passing the DataTable, it means it has done the hard-work of getting data via AppInsights and in those cases, we want to avoid another call completely so this should take precedence over the query.

PR #664, opened 2020-03-10
fix drop-down menu can't accessible issue

XiaoxuHU:
To fix the drop-down menu can't accessible issue.
[WorkItem](https://dev.azure.com/app-service-diagnostics-portal/app-service-diagnostics-portal/_workitems/edit/1031)

### DIFF
--- a/AngularApp/projects/diagnostic-data/src/lib/components/dropdown/dropdown.component.html
+++ b/AngularApp/projects/diagnostic-data/src/lib/components/dropdown/dropdown.component.html
@@ -16,7 +16,7 @@
           </button>
           <ul class="dropdown-menu" aria-labelledby="dropdownMenuId">
             <li *ngFor="let key of keys">
-              <a (click)="selectKey(key)">
+              <a tabindex="0" (keyup.enter)="selectKey(key)" (click)="selectKey(key)">
                   <span [innerHtml]="key"></span>
               </a>
             </li>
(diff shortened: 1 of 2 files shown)


AngularApp/projects/diagnostic-data/src/lib/components/dropdown/dropdown.component.html
```diff
@@ -16,7 +16,7 @@
           </button>
           <ul class="dropdown-menu" aria-labelledby="dropdownMenuId">
             <li *ngFor="let key of keys">
-              <a (click)="selectKey(key)">
+              <a href="#" (click)="selectKey(key,$event)">
```
cindylovescoding: Why are we doing this?

XiaoxuHU: I check the bootstarp3 example and figure out need to add a dummy href attributes for  accessibility. 
You can try to remove herf and it will lose accessibility[here](https://www.w3schools.com/bootstrap/tryit.asp?filename=trybs_dropdown-menu&stacked=h)

cindylovescoding: Hmmm interesting. So that's like a known issue and this is the workaround for bootstrap 3?

XiaoxuHU: Yes, the workaround is add href attributes.
To prevent navigate to href link, I added **e.preventDefault()** in .ts file

PR #669, opened 2020-03-12
Simplify Comms logic and Add ICM number to Comms

ShekharGupta1988:
Fixes:
- Simply Comms Logic
- Add ICM number as per request from CSS

image

### DIFF
--- a/Backend/Services/OutageCommunicationService/OutageCommunicationService.cs
+++ b/Backend/Services/OutageCommunicationService/OutageCommunicationService.cs
@@ -25,7 +25,7 @@ namespace Backend.Services
         GetCommunicationsBySubIdAndDateRange(@'{SUBSCRIPTION}', startDate, endDate) 
         | where CommunicationType == 'Outage'
         | order by PublishedTime asc
-        | project CommunicationId, PublishedTime, Title, RichTextMessage, Status, Severity, IncidentId, CommunicationType, ImpactedServices
+        | project CommunicationId, PublishedTime, Title, RichTextMessage, Status, Severity, IncidentId, CommunicationType, ImpactedServices, ExternalIncidentId
         ";
 
         public OutageCommunicationService(IKustoQueryService kustoQueryService)
@@ -74,7 +74,8 @@ namespace Backend.Services
                     Title = row["Title"].ToString(),
                     RichTextMessage = row["RichTextMessage"].ToString(),
                     Status = row["Status"].ToString().Equals("Active", StringComparison.OrdinalIgnoreCase) ? CommunicationStatus.Active : CommunicationStatus.Resolved,
-                    IncidentId = row["IncidentId"].ToString()
+                    IncidentId = row["IncidentId"].ToString(),
+                    IcmId = row["ExternalIncidentId"].ToString()
                 };
 
                 comm.ImpactedServices = GetImpactedRegions(row["ImpactedServices"].ToString());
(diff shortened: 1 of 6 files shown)


Backend/Services/OutageCommunicationService/OutageCommunicationService.cs
```diff
@@ -83,43 +84,28 @@ public async Task<List<Communication>> GetCommunicationsAsync(string subscriptio
...
-                        mostRecentImpactedServiceComm.IsAlert = true;
-                        mostRecentImpactedServiceComm.IsExpanded = false;
-                        // NOTE:- For now, resolved incidents will be collapsed by Default.
-                        // Uncommenting below line would make resolved incidents expanded by default for certain timespan.
-                        //mostRecentImpactedServiceComm.IsExpanded = ((currentTimeUTC - mostRecentImpactedServiceComm.PublishedTime) <= _commExpandedWindow);
-                        impactedServiceComm = mostRecentImpactedServiceComm;
-                    }
-                }
+                // Sometimes Communications belonging to same IncidentId may have different Impacted Services.
+                // To accurately identify the latest update:
```
cindylovescoding: So each communication row here will only have one impacted service right? And the published time for each impacted service won't be the same?

Backend/Services/OutageCommunicationService/OutageCommunicationService.cs
```diff
@@ -83,43 +84,28 @@ public async Task<List<Communication>> GetCommunicationsAsync(string subscriptio
...
+                // 1) Figure out the latest communication containing impacted service as your service (object : latestCommContainingImpactedService)
```
cindylovescoding: I forget why do we need this in UI controllers for both Applens and D&S? Instead of putting this into our runtimehost backend as a common functionality?

Backend/Services/OutageCommunicationService/OutageCommunicationService.cs
ShekharGupta1988: No, it can have multiple. 

Recently there was a outage in Core compute VM.
What I noticed was there was a comm published with impacted service as app service and status active. After couple of hours, there was another comm with same incident id but resolved status. The second one didnt had app service mentioned in impacted service,it had  "all" string.

So even though the issue was resolved, our previous logic still showed active comm with app service as impacted service.

Backend/Services/OutageCommunicationService/OutageCommunicationService.cs
ShekharGupta1988: If there is a geomaster or ARM outage going on, then our detectors will not work. 

But in those case as well, we need to show comms

hforeste:
Shekhar could you also show tracking id in the UI so that next time we have an issue we can easily find this ID and correlate with comms team on any issue? The service health portal also has the tracking ID. Customers will be able to correlate our comms UI with the incidents in service health portal.

Backend/Services/OutageCommunicationService/OutageCommunicationService.cs
rekhaswaminathan: We are showing comms inside of detector right? If call to GetDetector fails, how can we show the comms?

ShekharGupta1988:
> Shekhar could you also show tracking id in the UI so that next time we have an issue we can easily find this ID and correlate with comms team on any issue? The service health portal also has the tracking ID. Customers will be able to correlate our comms UI with the incidents in service health portal.

Ya, good idea. I'll put the Tracking Id there.

 For external UI, we also show a link that opens up azure portal with Tracking Id.
ex:  https://app.azure.com/h/CV95-ND8

image

ShekharGupta1988:
Added Tracking Id as well

image

Backend/Services/OutageCommunicationService/OutageCommunicationService.cs
ShekharGupta1988: Nope, we are not showing inside detector view. Its outside Detector view. Even if detector call fails, Comms still show up

PR #673, opened 2020-03-16
Fix observer 4xx issue for /details route

michimune:


AngularApp/projects/applens/src/app/modules/dashboard/dashboard/dashboard.component.ts
```diff
@@ -156,11 +156,11 @@ export class DashboardComponent implements OnDestroy {
   }
 
   openResourceInfoModal() {
-    if (this.keys.indexOf('VnetName') == -1 && this.resourceReady !== undefined && this.resourceDetailsSub === undefined)
+    if (this.keys.indexOf('VnetName') == -1 && typeof this.resourceReady !== 'undefined' && typeof this.resourceDetailsSub === 'undefined')
```
dagrooms52: Not sure if string comparison is the best way to check this, they could be `null` as well right?
Typically this will check if the var is null or undefined: `variable == null`

michimune: I see. I will change it.

michimune: Updated

PR #680, opened 2020-03-20
Fix bugs from bug bash

XiaoxuHU:


AngularApp/projects/app-service-diagnostics/src/app/fabric-ui/components/detector-command-bar/detector-command-bar.component.ts
```diff
@@ -29,13 +25,15 @@ export class DetectorCommandBarComponent {
   }
 
   toggleOpenTimePicker() {
-    this.globals.openTimePicker = !this.globals.openTimePicker;
+    // setTimeout(() => {this.globals.openTimePicker = !this.globals.openTimePicker},0);
```
ShekharGupta1988: get rid of comment line

PR #681, opened 2020-03-20
Nationalclouddev

hforeste:
- Add geomastername in the request headers
- Remove static configuration for Observer
- Redirect using 302 instead of 301 for Dsts Auth since it was causing cache issues with the browser
- View application exceptions in the HTTP response body
- Configure Observer for different cloud AAD authorities


ApplensBackend/Configuration/StartupNationalCloud.cs
```diff
@@ -112,6 +120,22 @@ public virtual void ConfigureServices(IServiceCollection services)
...
+            app.UseExceptionHandler(errorApp =>
+            {
+                errorApp.Run(async context =>
+                {
+                    context.Response.StatusCode = 500;
+                    context.Response.ContentType = "text/html";
+                    await context.Response.WriteAsync("<html lang=\"en\"><body>\r\n");
+                    await context.Response.WriteAsync("ERROR!<br><br>\r\n");
+                    var exceptionHandlerPathFeature = context.Features.Get<IExceptionHandlerPathFeature>();
+                    await context.Response.WriteAsync(exceptionHandlerPathFeature?.Error?.ToString());
```
rekhaswaminathan: Can you provide me some context on why we need this?

hforeste: It is not required but I would say it is nice to have so that we can see why something is failing in applens itself. It was helpful when I could use the F12 developer window and see the exception in the response in real time and when I didn't have application insights hooked up to the app.

rekhaswaminathan: Does this apply to detector failures also? If so, shouldn't this change be done in RuntimeHost?

hforeste: Detector failures won't count. Its only exceptions inside of applens app code its self. So stuff like failed to execute an observer call.

PR #686, opened 2020-03-24
Synchronize charts on detector

michimune:


cindylovescoding:
Hey Michi, Is this including the fix for accessibility to iterate series data through keyboard?

AngularApp/projects/diagnostic-data/src/lib/components/highcharts-graph/highcharts-graph.component.ts
```diff
@@ -282,6 +346,22 @@ export class HighchartsGraphComponent implements OnInit {
...
                 },
+                events: {
+                    setExtremes: function(evt: Highcharts.AxisSetExtremesEventObject) {
+                        if (evt.trigger !== 'sync') { // Prevent feedback loop
+                            for(let i = 0; i < Highcharts.charts.length; i++)
+                            {
+                                let chart = Highcharts.charts[i];
+
+                                if (chart && this.chart !== chart) {
+                                    if (chart.xAxis[0].setExtremes) { // It is null while updating
```
cindylovescoding: We may want to add an insight later on to remind users that insights on the same detector are not synced for the same time range that we select for graphs.

michimune: Yes. Detectors need to add such a note. I think we should create a new gist which many detectors can use in common.

michimune:
For the accessibility, it accepts left and right arrow keys. But unfortunately I could not implement to draw plot lines to all the charts. As a workaround, when a key is pressed we show tooltips without plot lines. It's not a perfect solution but at least consistency is provided. I will keep trying to improve the experience.

PR #687, opened 2020-03-24
fix padding in search result homepage, Api and Mobile App will navigate only to old UI

XiaoxuHU:


AngularApp/projects/app-service-diagnostics/src/app/fabric-ui/version-test.service.ts
```diff
@@ -44,6 +44,6 @@ export class VersionTestService {
 
     private isWindowsWebAppResource(site: Site, resourceType: ResourceType): boolean {
         let isLinuxPlatform = site && site.kind && site.kind.toLowerCase().indexOf('linux') >= 0;
-        return resourceType === ResourceType.Site && site && site.appType === AppType.WebApp && !isLinuxPlatform;
+        return resourceType === ResourceType.Site && site && site.appType === AppType.WebApp && !isLinuxPlatform && site.kind === 'app';
```
ShekharGupta1988: any reason for additional check. The kind can also be webapp

PR #692, opened 2020-03-26
Add a check for storage length till ANT88

puneetg1983:
Auto-healing server side code is encountering a buffer overrun if the storage account number is longer than 18 characters. The server side fix is applied but will go with ANT 88 which might take more than a month. Adding a check in UI code for Storage name length so that customers dont end up configuring a storage account which may then cause Autohealing to fail silently.

image

For pre-configured Storage accounts that are already exceeding in length, Auto-Heal UI will show this warning.

image

### DIFF
--- a/AngularApp/projects/app-service-diagnostics/src/app/betaSubscriptions.ts
+++ b/AngularApp/projects/app-service-diagnostics/src/app/betaSubscriptions.ts
@@ -10,26 +10,26 @@ export class DemoSubscriptions {
         'ef90e930-9d7f-4a60-8a99-748e0eea69de', // AntaresDemo
         'fb2c25dc-6bab-45c4-8cc9-cece7c42a95a', // antps14
         'edcc99a4-b7f9-4b5e-a9a1-3034c51db496', // OtherDemo?
-        'ce678eba-8ab3-4a0c-9610-97fee4ee6b34',  //Ranjith linux test sub
-        '88c8fe3d-1993-4fac-8e44-0f3232cc60ce',   // antps29,
-        'da511dea-6e00-4728-93ff-6302ad7fe284',   // Nazeef's Sub
-        '15c0f6ba-acb7-4f9a-aed0-7249c140b4f3', //Finbar' sub
+        'ce678eba-8ab3-4a0c-9610-97fee4ee6b34', // Ranjith linux test sub
+        '88c8fe3d-1993-4fac-8e44-0f3232cc60ce', // antps29,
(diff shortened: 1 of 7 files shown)


AngularApp/projects/app-service-diagnostics/src/app/betaSubscriptions.ts
```diff
@@ -10,26 +10,26 @@ export class DemoSubscriptions {
...
         '5abde51d-cc72-4bcc-b0d7-3c86b4db2a7c', // Shekhar's subscription
-        '6a205b0a-6b5a-4346-8468-f4b6367f7efa', //Private Test Sub HAWFOR (Hawk's subscription)
-        '6b6db65f-680e-4650-b97d-e82ed6a0f583', //Private Test Sub PUNEETG (Puneet's subscription)
+        '6a205b0a-6b5a-4346-8468-f4b6367f7efa', // Private Test Sub HAWFOR (Hawk's subscription)
+        '6b6db65f-680e-4650-b97d-e82ed6a0f583', // Private Test Sub PUNEETG (Puneet's subscription)
         '76679aa7-6638-4d84-b1c6-97dafcd8ebeb', // Juntao Zhu's Subscription
         '1c685347-a4a8-4cbc-8362-7a7b3c18d654', // Anik Shen's Subscription
-        '05c54750-4e91-4d98-981c-8c3e3356e9fa', //Ji Young Lee Susbscription
+        '05c54750-4e91-4d98-981c-8c3e3356e9fa', // Ji Young Lee Susbscription
         '8eeeb512-e3d7-4732-b40c-36df9ff628d0', // Nikhil Vetteth's Susbscription
```
hforeste: I know you didn't add this to your PR but can we remove these comments as some of them have people's full name in them...

ShekharGupta1988:
I'll go ahead and merge this PR in order to catch the afternoon deployment train.

puneetg1983:
> I'll go ahead and merge this PR in order to catch the afternoon deployment train.

Thanks Shekhar, I was away from my computer and didnt notice the comments.

AngularApp/projects/app-service-diagnostics/src/app/betaSubscriptions.ts
puneetg1983: so you are saying we can put alias instead of names?

hforeste: I may sound paranoid but I don't think its appropriate to have those either.

PR #696, opened 2020-03-26
Fix search suffix for windows web app

cindylovescoding:
Websites service is already defined as an injectable to be provided in "Root". To make sure the service is a singleton, it should not be provide in a lazy loaded module in websitesmodule again.


SharmaAjay19:
Nice! You are shooting down bugs in their cradle. Hopefully, now we can enable the genie flow for other subscriptions :)

PR #702, opened 2020-04-03
Call refresh on dynamic activate router component

cindylovescoding:
1. Add a separate service specifically for detector command in the new UI
2. Calling refresh() function of the current activated router component


cindylovescoding:
Fixed this in another PR.

PR #703, opened 2020-04-06
Hard refresh for refreshing request

cindylovescoding:
This seems like the only solution I can find so far to make this refreshing completely working. 
https://ms.portal.azure.com/?websitesextension_ext=asd.env%3Dstaging#home

In this solution, I only unsubscribe the observer in ngDestroy function. Meaning as long as the component instance is still in the routing cache and getting reused, it will be still listening to the refresh signal from detector control service. I am thinking if this is the right way to do, but since we are already maintain those cached component instances, then maybe we should also maintain their subscribing states to make it a complete story.

The side effect is that it's making all the get detector response calls for the cached components whenever we refresh a detector. Need to find a way to minimize the calls.


cindylovescoding:
@XiaoxuHU Local is always working. You can help verify on staging, that will tell us if that's working on prod.
Also I fix the unnecessary get detectors calls as well in the 2nd commit.

PR #707, opened 2020-04-10
Added HTTP GET to Invoke API

hforeste:
Currently, Invoke API is HTTP POST so it requires it either be called with some tool like postman or fiddler or from Applens UI. 
This PR adds HTTP GET for the Invoke API so the API can be directly called from the browser
This is just a scenario for when you only need a JSON response from the backend and its unnecessary to add UI elements to the Applens UI.
Can quickly be called from the browser with https://applens.azurewebsites.net?path={some_diagnostic_backend_path}&method={http_verb}

### DIFF
--- a/ApplensBackend/Controllers/DiagnosticController.cs
+++ b/ApplensBackend/Controllers/DiagnosticController.cs
@@ -11,12 +11,14 @@ using System.Net;
 using System.Net.Http;
 using System.Net.Http.Headers;
 using System.Threading.Tasks;
+using AppLensV3.Helpers;
 using AppLensV3.Services;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.Extensions.Configuration;
+using Microsoft.Net.Http.Headers;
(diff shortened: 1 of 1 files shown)


ApplensBackend/Controllers/DiagnosticController.cs
```diff
@@ -85,6 +87,35 @@ public IActionResult Ping()
...
+                return BadRequest(reason);
+            }
+
+            this.HttpContext.Request.Headers.Add(PathQueryHeader, path);
+            this.HttpContext.Request.Headers.Add(MethodHeader, method);
+            this.HttpContext.Request.Headers.Add(HeaderNames.ContentType, "application/json");
+
+            try
+            {
+                response = await Invoke(null);
```
rekhaswaminathan, line 108: Does this pass the `Request.Headers` also? It look like you're calling the ControllerMethod Invoke right?

hforeste: Yep just calling the controller method. The necessary headers are passed in the above 
```         
this.HttpContext.Request.Headers.Add(PathQueryHeader, path);
this.HttpContext.Request.Headers.Add(MethodHeader, method);
this.HttpContext.Request.Headers.Add(HeaderNames.ContentType, "application/json");

```

PR #716, opened 2020-04-20
CXP chat case creation

nmallick1:
Take diagnostics consent from the customer in chat and popup and pass those options to the CXP API's to be used during case creation.

The updated UI looks as follows.

image


hforeste:
Does this include permission for things like event logs and other logs from OSS frameworks like php and etc?

nmallick1:
@hforeste Yes it does. Here's a link to public documentation that we open while clicking on information icon.

https://azure.microsoft.com/en-us/support/legal/support-diagnostic-information-collection/

PR #728, opened 2020-04-27
Fix Side Nav in Applens Staging

XiaoxuHU:


AngularApp/projects/diagnostic-data/src/lib/services/telemetry/telemetry.service.ts
```diff
@@ -43,14 +44,14 @@ export class TelemetryService {
         }
     ];
     constructor(private _appInsightsService: AppInsightsTelemetryService, private _kustoService: KustoTelemetryService,
-        @Inject(DIAGNOSTIC_DATA_CONFIG) private config: DiagnosticDataConfig, private _versionService: VersionService, private _activatedRoute: ActivatedRoute, private _router: Router,private _diagnosticSiteService:DiagnosticSiteService) {
+        @Inject(DIAGNOSTIC_DATA_CONFIG) config: DiagnosticDataConfig, private _versionService: VersionService, private _activatedRoute: ActivatedRoute, private _router: Router,private _diagnosticSiteService:DiagnosticSiteService) {
```
cindylovescoding, line 47: Why we want to make the config this public?

XiaoxuHU: Just for VS Code can highlight the **config** parameter after we use it in this service. (I think there is a bug in VS Code, even though we already used **config** parameter in our code,  it still can't highlight this parameter if we add **private** modifier after **@Inject**)

PR #731, opened 2020-05-01
Add application insights to dotnet core project and upgrade to application insights sdk v2 for angular

cindylovescoding:


AngularApp/projects/diagnostic-data/src/lib/services/telemetry/appinsights-telemetry.service.ts
```diff
@@ -1,43 +1,81 @@
...
+    public logPageView(name?: string, url?: string, properties?: any, duration?: number) {  
+        properties = properties || {};
+        properties.duration = duration === undefined || duration === null ? 0 : duration;
+
+        const pageViewTelemetry: IPageViewTelemetry = {
+            name: name,
+            uri: url,
+            properties: properties,
+          };
+        this.appInsights .trackPageView(pageViewTelemetry);
```
rekhaswaminathan, line 35: Seems like there is an extra space here? `appInsights .trackPageView`

ApplensBackend/AppLensV3.csproj
```diff
@@ -45,6 +45,7 @@
 
   <PropertyGroup>
     <CodeAnalysisRuleSet>ca.ruleset</CodeAnalysisRuleSet>
+    <ApplicationInsightsResourceId>/subscriptions/c1972e9d-b3c7-4de4-acb3-681773b28ced/resourceGroups/DiagnoseAndSolve/providers/microsoft.insights/components/DiagnoseAndSolve</ApplicationInsightsResourceId>
```
rekhaswaminathan, line 48: Do we want to use the same resource for AppLens as well?

Backend/Backend.csproj
```diff
@@ -2,13 +2,16 @@
 
   <PropertyGroup>
     <TargetFramework>netcoreapp2.0</TargetFramework>
+    <ApplicationInsightsResourceId>/subscriptions/c1972e9d-b3c7-4de4-acb3-681773b28ced/resourcegroups/DiagnoseAndSolve/providers/microsoft.insights/components/DiagnoseAndSolve</ApplicationInsightsResourceId>
```
rekhaswaminathan, line 5: Do you mind telling why this resource id needs to be added in .csproj file?

ApplensBackend/AppLensV3.csproj
ShekharGupta1988, line 48: Few comments here :
- Can we use AppSettings to load this resource instead of putting the resource Uri here. It will also help when you look to expand to different N-Clouds
-  I see we are using the same resource for Applens and Diagnose&Solve. Is there any distinction at the event level to filter them out.
- We can prob name this App Insights resource as 'AppServicesDiagnosticsPortal'

hforeste: From a configurations standpoint, all you need to do is specify the instrumentation key in the environment variable. So `APPINSIGHTS_INSTRUMENTATIONKEY` in app service app settings.

cindylovescoding: CIL

Backend/Backend.csproj
cindylovescoding: I will move this to app settings Rekha

ApplensBackend/AppLensV3.csproj
cindylovescoding: Yes I plan to use the same instance for the two dotcore projects, also the angular project

AngularApp/projects/diagnostic-data/src/lib/services/telemetry/appinsights-telemetry.service.ts
cindylovescoding: Good catch haha! Somehow this got submitted after my local build

AngularApp/projects/app-service-diagnostics/src/app/app.module.ts
```diff
@@ -11,6 +11,9 @@ import { RouteReuseStrategy, RouterModule } from '@angular/router';
 import {
   KustoTelemetryService
 } from '../../../diagnostic-data/src/lib/services/telemetry/kusto-telemetry.service';
+import {
+  AppInsightsTelemetryService
```
rekhaswaminathan, line 15: Minor feedback here. There is an import statement in the beginning of the file which imports necessary classes from `diagnostic-data`.If you like, you can add `AppInsightsTelemetryService` to that line: 
`import {
  CommsService, DiagnosticDataModule, DiagnosticService, DiagnosticSiteService,
  PUBLIC_DEV_CONFIGURATION, PUBLIC_PROD_CONFIGURATION, SolutionService, SettingsService, BackendCtrlQueryService, GenieGlobals, VersionService
} from 'diagnostic-data';`

This helps in keeping the file size short and imports from `diagnostic-data` are in one place.

AngularApp/projects/app-service-diagnostics/src/app/shared/services/portal-appinsights-telemetry.service.ts
```diff
@@ -0,0 +1,109 @@
+import { Injectable, OnInit, Inject, Input } from '@angular/core';
+import { ApplicationInsights, Snippet, IPageViewTelemetry, IEventTelemetry, IExceptionTelemetry, SeverityLevel, ITraceTelemetry, IMetricTelemetry, ITelemetryItem } from '@microsoft/applicationinsights-web'
+import { ITelemetryProvider } from 'diagnostic-data';
+import { BackendCtrlService } from './backend-ctrl.service';
+import { of, forkJoin } from 'rxjs';
+import { map, retry, catchError } from 'rxjs/operators';
+
+@Injectable()
+export class PortalAppInsightsTelemetryService implements ITelemetryProvider {
```
hforeste, line 9: Between this class and `ApplensAppinsightsTelemetryService` looks like there are some duplicate code that can be shared between the two. Maybe create a base AppInsights class?

Backend/Services/AppInsightsTelemetry/AppInsightsTelemetryInitializer.cs
```diff
@@ -0,0 +1,31 @@
...
+namespace Backend.Services
+{
+    public class AppInsightsTelemetryInitializer : ITelemetryInitializer
+    {
+        private static readonly string EnvironmentName;
+        private static readonly string WebsiteHostName;
+
+        static AppInsightsTelemetryInitializer()
+        {
+            EnvironmentName = Environment.GetEnvironmentVariable("ADS_ENVIRONMENT");
```
ShekharGupta1988, line 14: nitpick : should this be ASD_ENVIRONMENT or DIAGPORTAL_ENVIRONMENT?
ADS stands for Azure Diagnostic Services and might cause confusion later on.

ApplensBackend/Controllers/DiagnosticController.cs
```diff
@@ -76,14 +76,45 @@ public IActionResult Ping()
...
+        [HttpGet("environmentVariable/{name}")]
+        public IActionResult GetEnvironmentVariable(string name)
+        {
+            if (string.IsNullOrEmpty(name))
+            {
+                return BadRequest("Environment variable name is empty");
+            }
+
+            return Ok(Environment.GetEnvironmentVariable(name));
+        }
```
hforeste, line 98: You won't need this because in the `Startup.cs` we add EnvironmentVariables as a configuration store in the IConfiguration class. See https://github.com/Azure/Azure-AppServices-Diagnostics-Portal/blob/master/ApplensBackend/Configuration/Startup.cs#L24-L28. 

So the `GetAppSettingValue(string name)` should suffice for your needs.

ApplensBackend/Controllers/DiagnosticController.cs
```diff
@@ -76,14 +76,45 @@ public IActionResult Ping()
...
+            };
+
+            if (Env.IsDevelopment())
+            {
+                environmentConfiguration["environment"] = "local";
+                environmentConfiguration["websiteHostName"] = "applens";
+            }
+
+            return Ok(environmentConfiguration);
+        }
```
hforeste, line 116: Can you use the `GetAppSettingValue` to get these values and form the dictionary in the UI code?

ApplensBackend/Services/ApplensTelemetry/ApplensTelemetryInitializer.cs
```diff
@@ -0,0 +1,31 @@
...
+    public class ApplensTelemetryInitializer : ITelemetryInitializer
+    {
+        private static readonly string EnvironmentName;
+        private static readonly string WebsiteHostName;
+
+        static ApplensTelemetryInitializer()
+        {
+            EnvironmentName = Environment.GetEnvironmentVariable("APPLENS_ENVIRONMENT");
+            WebsiteHostName = Environment.GetEnvironmentVariable("APPLENS_HOST");
+        }
```
hforeste, line 16: Instead of directly using the Environment object maybe use the dependency injected IConfiguration object to get these values that way if you provided settings in appsettings.json and not in the environment variables then this class will still have what it needs.

ApplensBackend/Services/ApplensTelemetry/ApplensTelemetryInitializer.cs
```diff
@@ -0,0 +1,31 @@
...
+
+        public void Initialize(ITelemetry telemetry)
+        {
+            if (!telemetry.Context.GlobalProperties.ContainsKey("environment"))
+            {
+                telemetry.Context.GlobalProperties.Add("environment", EnvironmentName);
```
hforeste, line 22: Can the ITelemtry object accept null values?

cindylovescoding: Yes I tried

AngularApp/projects/app-service-diagnostics/src/app/shared/services/portal-appinsights-telemetry.service.ts
cindylovescoding: Yes I have thought about that @hforeste , here are the things I was thinking when I decided to separate this service. If we keep the appInsights telemetry service in the common library:

- We will need to add another service interface in the common module to say "_ITelemetryConfigurationService_", and two more new services will be needed to add to implement that interface _ITelemetryConfigurationService_ in applens and portal project.

- We lose the flexibility to customize telemetry for applens or portal

- Not following the other FE services pattern, like content service (used to get cognitive search result) or comms service.

AngularApp/projects/app-service-diagnostics/src/app/app.module.ts
cindylovescoding: Fixed it. Thanks @rekhaswaminathan !

Backend/Services/AppInsightsTelemetry/AppInsightsTelemetryInitializer.cs
cindylovescoding: > nitpick : should this be ASD_ENVIRONMENT or DIAGPORTAL_ENVIRONMENT?
> ADS stands for Azure Diagnostic Services and might cause confusion later on.

Sounds good Shekhar. I always thought ADS stands for **Azure Diagnose and Solve**..

ApplensBackend/Services/ApplensTelemetry/ApplensTelemetryInitializer.cs
cindylovescoding: So the intention adding this is to get environment variables which does not exist in app settings. Are you sure that I will be able to get everything in environment variables from  getAppSettings?

ApplensBackend/Controllers/DiagnosticController.cs
cindylovescoding: I see. @hforeste Do I need to change the key name, eg,. "Applens_Host" with some prefix like "Appsettings_"? Or I am safe to use whatever key I set in portal?

ApplensBackend/Controllers/DiagnosticController.cs
cindylovescoding: That will be two API calls though

ApplensBackend/Controllers/DiagnosticController.cs
hforeste: Your safe to use whatever you set.

ApplensBackend/Controllers/DiagnosticController.cs
hforeste: Are you worried about performance of the call?

ApplensBackend/Services/ApplensTelemetry/ApplensTelemetryInitializer.cs
hforeste: So in this case you wouldn't be using the `getAppSettings` method that is in the controller. Instead you'll reference the injected `IConfiguration` through the constructor.

So the way it works is: https://github.com/Azure/Azure-AppServices-Diagnostics-Portal/blob/master/ApplensBackend/Configuration/Startup.cs#L24-L28

            ```var builder = new ConfigurationBuilder()
                .SetBasePath(env.ContentRootPath)
                .AddJsonFile("appsettings.json", optional: false, reloadOnChange: true)
                .AddJsonFile($"appsettings.{env.EnvironmentName}.json", optional: true)
                .AddEnvironmentVariables();```

Whenever you use `IConfiguration` to get configuration value it will use this list to find the values and it is in the order of least to greatest precedence and this case environment variables have the highest precedence so even if a value was specified in appsettings.json it will be overridden by the value in environment variables.

PR #732, opened 2020-05-01
Xiaoxu/accessibility

XiaoxuHU:
This PR fix all the [Sev1 bugs](https://msazure.visualstudio.com/Antares/_queries?tempQueryId=8e7d68fe-811b-43d9-8ba8-afc7da1af938) on my plate and also update the style for DynamicInsight rendering type to be same as insight

### DIFF
--- a/AngularApp/projects/diagnostic-data/src/lib/components/dropdown-v4/dropdown-v4.component.ts
+++ b/AngularApp/projects/diagnostic-data/src/lib/components/dropdown-v4/dropdown-v4.component.ts
@@ -82,12 +82,13 @@ export class DropdownV4Component extends DataRenderBaseComponent {
     this.fabDropdownWidth = this.calculateFabWidth(this.options);
   }
 
-  selectKey(key: string) {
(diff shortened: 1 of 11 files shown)


AngularApp/projects/diagnostic-data/src/lib/components/dropdown-v4/dropdown-v4.component.ts
```diff
@@ -82,12 +82,13 @@ export class DropdownV4Component extends DataRenderBaseComponent {
     this.fabDropdownWidth = this.calculateFabWidth(this.options);
   }
 
-  selectKey(key: string) {
+  selectKey(key: string,event:any) {
```
cindylovescoding, line 85: What is this for?

XiaoxuHU: this is for prevent navigating to the url that I added in href attribute in html. I added href is for accessibility

PR #750, opened 2020-05-19
More accessibility fixes

puneetg1983:


AngularApp/projects/app-service-diagnostics/src/app/auto-healing/autohealing-memory-rule/autohealing-memory-rule.component.html
```diff
@@ -6,15 +6,18 @@
       Configure Private Bytes rule
     </button>
   </div>
+  <div aria-live="polite" class="mt-1" [style.visibility]="displayDeleteRuleMessage ? 'visibile' : 'hidden'">
```
cindylovescoding, line 9: Oh wow I didn't know this aria-live before, interesting learning!
aria-live: The aria-live=POLITENESS_SETTING is used to set the priority with which screen reader should treat updates to live regions - the possible settings are: off, polite or assertive. The default setting is off. This attribute is by far the most important.

puneetg1983: I am learning stuff myself....   :-)

There is so much out there, its just hard to find out everything.

PR #751, opened 2020-05-19
Accessibility bugs

rekhaswaminathan:
Fixing accessibility bugs assigned

### DIFF
--- a/AngularApp/projects/diagnostic-data/src/lib/components/changeanalysis-onboarding/changeanalysis-onboarding.component.html
+++ b/AngularApp/projects/diagnostic-data/src/lib/components/changeanalysis-onboarding/changeanalysis-onboarding.component.html
@@ -1,5 +1,5 @@
 <data-container *ngIf="!isPublic; else publicView">
-    <div class="insight-container">
+    <div class="insight-container" role="button" tabindex="0" attr.aria-label="{{onboardingText}}">
         <div class="panel panel-default insight-panel">
           <div class="panel-heading">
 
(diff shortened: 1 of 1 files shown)


AngularApp/projects/diagnostic-data/src/lib/components/changeanalysis-onboarding/changeanalysis-onboarding.component.html
```diff
@@ -10,12 +10,12 @@ <h5 class="panel-title"> <span class="fa status-icon fa-exclamation-triangle"></
 </data-container>
 
 <ng-template #publicView>
-    <div class="insight-container hyperlink" (click)="navigateToSettings()">
+    <div class="insight-container hyperlink" (click)="navigateToSettings()" role="button" tabindex="0" attr.aria-label="{{onboardingText}}">
```
cindylovescoding, line 13: Do we want to set the keyup.enter or keyup.space events to hook up with the click function?

rekhaswaminathan: So I had kept the keyup.enter event to the inner div which has the "Enable Now!" text. Do you think I should move it here as well?

cindylovescoding: Oh I see. Then it should be fine for now. The inner dive functionality will be keyboard accessible

PR #756, opened 2020-05-20
Add portalVersion and resourceType into common method for using in all logging 

XiaoxuHU:


AngularApp/projects/app-service-diagnostics/src/app/shared/services/portal-appinsights-telemetry.service.ts
```diff
@@ -56,6 +57,15 @@ export class PortalAppInsightsTelemetryService implements ITelemetryProvider {
                         envelop.data["environment"] = this.environment ? this.environment : "test";
                         envelop.data["websiteHostName"] = this.websiteHostName ? this.websiteHostName : "appservice-diagnostics";
                         envelop.data["isfrontend"] = true;
+
+                        try {
+                            const isLegacy = this._versionTestService.isLegacySub.value;
+                            envelop.data["portalVersion"] = isLegacy ? 'V2' : 'V3';
+                            const isWindowsApp = this._versionTestService.isWindowsWebApp.value;
```
cindylovescoding, line 64: Why do we need this?

XiaoxuHU: Just in case it won't break anything

AngularApp/projects/diagnostic-data/src/lib/services/telemetry/telemetry.service.ts
```diff
@@ -77,48 +77,51 @@ export class TelemetryService {
...
             }
-            telemetryProvider.logPageView(name, url, properties, measurements, duration);
+            if(telemetryProvider) {
+                telemetryProvider.logPageView(name, url, properties, measurements, duration);
+            }
         }
     }
 
     public logException(exception: Error, handledAt?: string, properties?: any, severityLevel?: SeverityLevel) {
+        this.addCommonAttriToLog(properties);
```
cindylovescoding, line 102: Is this fine if exception happened in "findProductName" under "addCommonAttriToLog"?
Will there be cyclic dependency?
logException() -> addCommonAttriToLog() -> findProductName() -> logException()

AngularApp/projects/diagnostic-data/src/lib/services/telemetry/telemetry.service.ts
```diff
@@ -167,4 +170,19 @@ export class TelemetryService {
         
         return productName;
     }
+
+    private addCommonAttriToLog(properties: { [name: string]: string }):void {
```
cindylovescoding, line 174: Just a nitpick: Could we rename this function to be something like addCommonLoggingProperties()?

AngularApp/projects/app-service-diagnostics/src/app/shared/services/portal-appinsights-telemetry.service.ts
cindylovescoding: No I am asking why need this: isWindowsApp

AngularApp/projects/app-service-diagnostics/src/app/shared/services/portal-appinsights-telemetry.service.ts
```diff
@@ -56,6 +57,15 @@ export class PortalAppInsightsTelemetryService implements ITelemetryProvider {
                         envelop.data["environment"] = this.environment ? this.environment : "test";
                         envelop.data["websiteHostName"] = this.websiteHostName ? this.websiteHostName : "appservice-diagnostics";
                         envelop.data["isfrontend"] = true;
+
+                        try {
+                            const isLegacy = this._versionTestService.isLegacySub.value;
+                            envelop.data["portalVersion"] = isLegacy ? 'V2' : 'V3';
```
cindylovescoding, line 63: Nitpick: PortalVersion is not inconsistent with "initialPortalVersion", which are using lower cases like "v2" and "v3". Could we change either place to be consistent?

AngularApp/projects/diagnostic-data/src/lib/services/telemetry/telemetry.service.ts
XiaoxuHU: Sure,will do

AngularApp/projects/app-service-diagnostics/src/app/shared/services/portal-appinsights-telemetry.service.ts
XiaoxuHU: As for logging, what kind of naming rule we expect to follow ? 
We can use camel case for all key and values in logging

cindylovescoding: Make your call xiaoxu. It's good as long as it is consistent

AngularApp/projects/diagnostic-data/src/lib/services/telemetry/telemetry.service.ts
XiaoxuHU: I see, I can remove **logException** in **findProductName** to break the cycle. Or I can remove the whole try catch block

AngularApp/projects/app-service-diagnostics/src/app/shared/services/portal-appinsights-telemetry.service.ts
XiaoxuHU: Yes, I edit my comment above. Will remove this variable

PR #765, opened 2020-06-02
Fix exception for no dynamic component rendering found for downtime

cindylovescoding:


AngularApp/projects/diagnostic-data/src/lib/components/dynamic-data/dynamic-data.component.ts
```diff
@@ -70,6 +70,10 @@ export class DynamicDataComponent implements OnInit {
     this.versionService.isLegacySub.subscribe(isLegacy => this.isLegacy = isLegacy);
     this.dataBehaviorSubject.subscribe((diagnosticData: DiagnosticData) => {
       const component = this._findInputComponent((<Rendering>diagnosticData.renderingProperties).type);
+      if (component == null)
+      {
+        return;
```
ShekharGupta1988, line 75: Awesome!! Small quick fix.
nitpick : Can you add a log statement stating 'no component found for rendering type : foobar'. It might help in future.

cindylovescoding: Good point Shekhar! Will add that

PR #771, opened 2020-06-05
Open Change Analysis blade for function app

rekhaswaminathan:


cindylovescoding:
Rekha we will need to make the relevant changes for the new D&S once we enable it for function app, right?

cindylovescoding:
> Rekha we will need to make the relevant changes for the new D&S once we enable it for function app, right?

Oh never mind I think that's have been taken care of already.. like for web app..

PR #776, opened 2020-06-11
Nmallick1/downtimes

nmallick1:
Support for downtimes in external portal and AppLens.

Downtimes will allow users to set focus on when the issue started thereby eliminating noise generated by unrelated events that occurred outside of the impact time. It will further reduce load on the underlying Kusto clusters by executing time restricted queries instead of for the entire 24 hours duration. Downtimes are only supported in analysis. In order to use downtimes, the main analysis csx needs to use res.AddDownTimes(new List&lt;DownTime&gt;())

image

image

image

image

image


AngularApp/projects/app-service-diagnostics/src/app/shared/components/generic-analysis/generic-analysis.component.html
```diff
@@ -1,4 +1,4 @@
-<detector-container *ngIf="displayDetectorContainer" [detector]="analysisDetector" [isAnalysisView]="true"></detector-container>
+<detector-container *ngIf="displayDetectorContainer" [detector]="analysisDetector" [isAnalysisView]="true" (downTimeChanged)="ondownTimeChanged($event)"></detector-container>
```
hforeste, line 1: Small nitpik - use camelCasing for `ondownTimeChanged`

AngularApp/projects/app-service-diagnostics/src/app/shared/components/generic-analysis/generic-analysis.component.html
```diff
@@ -7,18 +7,10 @@
...
-  <div>
-    <div style="display: inline-block;margin-left: 20px">
-      <h2>{{ detectorName }}</h2>
-    </div>
-    <div style="display: inline-block;margin-left: 10px">
-      <a tabindex="0" (keyup.enter)="goBackToAnalysis()" aria-label="Link to go back to the analysis view" role="button"
-        (click)="goBackToAnalysis()" *ngIf="detectorId !== ''">Back to Observations</a>
-    </div>
-  </div>
+<detector-list-analysis #detectorListAnalysis withinDiagnoseAndSolve="true" detectorParmName="detectorName" [resourceId]="resourceId" [targetedScore]="targetedScore" [analysisId]="analysisId" [searchTerm]="searchTerm" [searchMode]="searchMode" (onComplete)="updateLoadingStatus($event)"></detector-list-analysis>
```
rekhaswaminathan, line 11: In `DetectorControlService` there is a property `internalClient`, can you use that instead of `withinDiagnoseAndSolve`?

AngularApp/projects/diagnostic-data/src/lib/components/detector-list-analysis/detector-list-analysis.component.ts
```diff
@@ -223,77 +239,104 @@ export class DetectorListAnalysisComponent extends DataRenderBaseComponent imple
...
     }
 
+    analysisContainsDowntime() : Observable<boolean> {
+        if(this.analysisId === 'searchResultsAnalysis') {
+            return of(false);
+        }
+        return this._diagnosticService.getDetector(this.analysisId, this._detectorControl.startTimeString, this._detectorControl.endTimeString).pipe(
+            map((response: DetectorResponse) => {
+                let downTimeRenderingType = response.dataset.find(set => (<Rendering>set.renderingProperties).type === RenderingType.DownTime);
+                if(!!downTimeRenderingType) {
```
rekhaswaminathan, line 249: Are we adding a new `RenderingType `in our backend?

nmallick1: This rending type has been there for a long time now, there was no UI supporting it though up until now.

AngularApp/projects/app-service-diagnostics/src/app/shared/components/generic-analysis/generic-analysis.component.ts
```diff
@@ -54,7 +63,8 @@ export class GenericAnalysisComponent extends GenericDetectorComponent implement
       this._cxpChatService.logChatEligibilityCheck('Call to CXP Chat API skipped. Config is not Public.', JSON.stringify(checkOutcome));
     }    
     this._activatedRouteLocal.paramMap.subscribe(params => {
-      this.analysisId = this.analysisId === "" ? params.get('analysisId'): this.analysisId;
+      this.analysisId = (this.analysisId != 'searchResultsAnalysis' && !!params.get('analysisId'))? params.get('analysisId') : this.analysisId;
```
cindylovescoding, line 66: The previous logic is input 'analysisId' to this component will take precedence over params['analysisId']. The new logic will be: except 'searchResultsAnalysis', router params['analysisId'] take priority over input 'analysisId'. We should be a little bit mindful on this change.

AngularApp/projects/applens/src/app/modules/dashboard/tabs/tab-analysis/tab-analysis.component.html
```diff
@@ -1,26 +1,8 @@
...
-      <div style="display: inline-block;margin-left: 10px;margin-right:20px">
-        <button class="btn btn-sm" *ngIf="detectorId !== ''" [routerLink]="['../../']" preserveFragment
-          queryParamsHandling="preserve">Back to Observations</button>
-      </div>
-    </div>
-  </div>
-  <router-outlet (activate)="onActivate($event)"></router-outlet>
-</div>
\ No newline at end of file
+<router-outlet (activate)="onActivate($event)"></router-outlet>
```
cindylovescoding, line 8: The old code will hide <router-outlet> if detectorId === '', we probably want to add it here if removing that could cause inconsistency

AngularApp/projects/app-service-diagnostics/src/app/shared/components/generic-analysis/generic-analysis.component.ts
nmallick1: I did check this piece across multiple scenarios and confirmed that dynamic analysis is the only one that is using this as an input field. Everywhere else, the value is being populated via router. However you are correct, in case in future we decide to have a component that passes this input, the logic here also needs to be updated.

AngularApp/projects/applens/src/app/modules/dashboard/tabs/tab-analysis/tab-analysis.component.html
nmallick1: The way we have routes defined now, the router will show content only when there is a detectorId present. Also, the TabAnalysis component is not reloaded when a child path is called so the detectorId is always null here. I should in fact cleanup the code here and remove the detectorId from this file. I'll clean up this code.

PR #781, opened 2020-06-12
Fixing DAAS Storage Account selection for National Clouds

puneetg1983:
With this PR:-
1. Temporarily disabling the forcing of storage account for Memory dumps for National Clouds
2. Adding the relevant information to pass while setting the BlobSasUri when DAAS update is everywhere.

### DIFF
--- a/AngularApp/projects/app-service-diagnostics/src/app/shared/components/daas/daas-validator.component.ts
+++ b/AngularApp/projects/app-service-diagnostics/src/app/shared/components/daas/daas-validator.component.ts
@@ -34,6 +34,15 @@ export class DaasValidatorComponent implements OnInit {
   diagnosers: DiagnoserDefinition[];
 
   constructor(private _serverFarmService: ServerFarmDataService, private _siteService: SiteService, private _daasService: DaasService) {
+    
+    //
+    // Temporary fix to unblock users in national clouds.
+    // Should be removed after DAAS 1.40 update
+    //
+    
+    if (this._daasService.isNationalCloud) {
+      this.diagnosersRequiringStorageAccount=[];
+    }
   }
 
   validateDaasSettings() {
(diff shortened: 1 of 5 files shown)


AngularApp/projects/app-service-diagnostics/src/app/shared/components/daas/daas-validator.component.ts
```diff
@@ -34,6 +34,15 @@ export class DaasValidatorComponent implements OnInit {
   diagnosers: DiagnoserDefinition[];
 
   constructor(private _serverFarmService: ServerFarmDataService, private _siteService: SiteService, private _daasService: DaasService) {
+    
+    //
+    // Temporary fix to unblock users in national clouds.
+    // Should be removed after DAAS 1.40 update
+    //
+    
+    if (this._daasService.isNationalCloud) {
```
hforeste, line 43: Why not inject the ARM service and use the armclient directly to find if its national cloud?

hforeste:
So will this default back to the old behavior where memory dumps were written to D:\home?

AngularApp/projects/app-service-diagnostics/src/app/shared/components/daas/daas-validator.component.ts
ShekharGupta1988: I believe thats what he is doing in DaaSService since he is using this flag in that service as well.
Since this component already injects DaaSService, might be ok to use this method and avoid extra injection.

ShekharGupta1988: I figured out you are not using this method `_daasService.isNationalCloud` anywhere except here. 
So either approach is fine.

puneetg1983:
> So will this default back to the old behavior where memory dumps were written to D:\home?

Yes, for national clouds it will

AngularApp/projects/app-service-diagnostics/src/app/shared/components/daas/daas-validator.component.ts
puneetg1983: Somehow I thought it is easier to use it this way.

PR #783, opened 2020-06-12
Removing HTTPS from endpoint suffix

puneetg1983:


AngularApp/projects/app-service-diagnostics/src/app/shared/services/arm.service.ts
```diff
@@ -87,7 +87,8 @@ export class ArmService {
     }
 
     get storageUrl(): string {
-        return this.armUrl.replace("management", "core");
+        let storageUrl = this.armUrl.replace("management", "core").replace("https://", "");
```
hforeste, line 90: Why remove the protocol?

puneetg1983: Because endpoint suffix doesn't need https://

PR #785, opened 2020-06-13
Token refresh mechanism for portal blade

SharmaAjay19:
Service Health Issue [1333: [Portal] 401 (expired token) exceptions in Portal while talking to ARM](https://dev.azure.com/app-service-diagnostics-portal/app-service-diagnostics-portal/_workitems/edit/1333)
We see a lot of 401 (expired token) issues in portal when talking to ARM. It's because of the ARM token being expired and not getting refreshed from ibiza.
This is hampering our customer experience by displaying them nothing when requests are made. ~1,667 distinct users have been impacted due to this issue in last 1 month.

[Failure count trend weekly](https://ms.portal.azure.com/#@72f988bf-86f1-41af-91ab-2d7cd011db47/blade/Microsoft_Azure_Monitoring_Logs/LogsBlade/resourceId/%2Fsubscriptions%2Fc1972e9d-b3c7-4de4-acb3-681773b28ced%2FresourceGroups%2FDiagnoseAndSolve%2Fproviders%2Fmicrosoft.insights%2Fcomponents%2FDiagnoseAndSolvePortal/source/LogsBlade.AnalyticsShareLinkToQuery/q/H4sIAAAAAAAAA12P3WrDMAyF7wN9B5GrBEbdwmBsbXqzJ1Fs0RhiOUgypWUP37iFdez2Oz%252FSmclAbcBz7j53oT9smnkl9CT7VEGghTgQ%252B0i6aX7gMpEQWEykhmmB01oAyOEPOq4NLytjIvCZDSMrtJPZol%252FOJWQ8UyK2Ld6K0Nbn5LSM6iUuFjOrax%252B9%252F%252FIukJG3LNq%252Bbghpme07B4JhgPfdvkpaUkKJt5oubF0P4xXGyN3vp2%252FwEfpqlTpRHhP8hGLQ3AGWVKvkGgEAAA%253D%253D)
image

### DIFF
--- a/AngularApp/projects/app-service-diagnostics/src/app/startup/services/auth.service.ts
+++ b/AngularApp/projects/app-service-diagnostics/src/app/startup/services/auth.service.ts
@@ -55,7 +55,7 @@ export class AuthService {
             try{
                 var token = jwt_decode(this.currentToken);
                 var currentTime = (new Date().getTime())/1000;
-                var refreshInterval = token.exp - currentTime - 60;
+                var refreshInterval = token.exp - currentTime + 5;
                 // Keeping this console log for tracking during debugging purposes
                 console.log("Scheduling token refresh - ", " current time: ", currentTime, " refreshes in: ", refreshInterval);
                 if (refreshInterval <= 0) {
--- a/AngularApp/projects/app-service-diagnostics/src/app/startup/services/auth.service.ts
+++ b/AngularApp/projects/app-service-diagnostics/src/app/startup/services/auth.service.ts
@@ -5,6 +5,7 @@ import { StartupInfo, ResourceType } from '../../shared/models/portal';
 import { PortalService } from './portal.service';
 import { map } from 'rxjs/operators';
 import { environment } from 'projects/app-service-diagnostics/src/environments/environment';
+import * as jwt_decode from "jwt-decode";
 
 @Injectable()
 export class AuthService {
@@ -28,6 +29,10 @@ export class AuthService {
 
     constructor(private _http: HttpClient, private _portalService: PortalService) {
         this.inIFrame = window.parent !== window;
+        this._portalService.getToken().subscribe(token => {
+            this.setAuthToken(token);
+        });
+        this.scheduleNextRefresh();
     }
 
     getAuthToken(): string {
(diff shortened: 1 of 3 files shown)


AngularApp/projects/app-service-diagnostics/src/app/startup/services/auth.service.ts
```diff
@@ -44,6 +50,41 @@ export class AuthService {
...
     }
 
+    scheduleNextRefresh(retryOnFailure=true) {
+        if (this.currentToken) {
+            try{
+                var token = jwt_decode(this.currentToken);
+                var currentTime = (new Date().getTime())/1000;
+                var refreshInterval = token.exp - currentTime + 5;
+                // Keeping this console log for tracking during debugging purposes
+                console.log("Scheduling token refresh - ", " current time: ", currentTime, " refreshes in: ", refreshInterval);
```
ShekharGupta1988, line 60: Use Telemetry service instead of console log

AngularApp/projects/app-service-diagnostics/src/app/startup/services/auth.service.ts
```diff
@@ -44,6 +50,41 @@ export class AuthService {
         this.currentToken = token;
     }
 
+    scheduleNextRefresh(retryOnFailure=true) {
+        if (this.currentToken) {
+            try{
+                var token = jwt_decode(this.currentToken);
+                var currentTime = (new Date().getTime())/1000;
+                var refreshInterval = token.exp - currentTime + 5;
```
cindylovescoding, line 58: Why not refresh immediately after token expired?

AngularApp/projects/app-service-diagnostics/src/app/startup/services/portal.service.ts
```diff
@@ -133,6 +145,12 @@ export class PortalService {
         console.log('[iFrame] Received mesg: ' + methodName, event);
 
         if (methodName === Verbs.sendStartupInfo) {
+            if (this.willRefreshToken){
+                const info = <StartupInfo>data;
+                this.tokenObservable.next(info.token);
+                this.willRefreshToken = false;
```
cindylovescoding, line 151: Do we want to set up the sessionId and startupInfoObservable here as well?

cindylovescoding:
I think we have the same issue for Applens as well. Sometimes token past due and users have to refresh the page. @SharmaAjay19  Maybe we can fix that issue with same solution here?

AngularApp/projects/app-service-diagnostics/src/app/startup/services/auth.service.ts
SharmaAjay19: The token which the portal gives us is still the stale one, it gets a new token if we refresh after the existing token has expired.

AngularApp/projects/app-service-diagnostics/src/app/startup/services/portal.service.ts
SharmaAjay19: Nope. We only need the token.

AngularApp/projects/app-service-diagnostics/src/app/startup/services/auth.service.ts
cindylovescoding: Yes I mean why +5s?

AngularApp/projects/app-service-diagnostics/src/app/startup/services/auth.service.ts
SharmaAjay19: Found a better and simpler logic for refreshing the token. Will update the PR once confirmed.

AngularApp/projects/app-service-diagnostics/src/app/startup/services/auth.service.ts
SharmaAjay19: Never mind, this logic is going away. I've found a different solution for token refresh.

SharmaAjay19:
> I think we have the same issue for Applens as well. Sometimes token past due and users have to refresh the page. @SharmaAjay19 Maybe we can fix that issue with same solution here?

I don't think Applens has this issue because Applens frontend is not in an iframe in a portal. This means the frontend should be able to automatically refresh its token through AAD via the ADAL library unless there is some configuration missing there. I am not very aware of how the Applens id token is managed but the logic should be to automatically refresh the id token and generally libraries like ADAL and MSAL are supposed to do that.

AngularApp/projects/app-service-diagnostics/src/app/startup/services/auth.service.ts
```diff
@@ -28,6 +28,9 @@ export class AuthService {
 
     constructor(private _http: HttpClient, private _portalService: PortalService) {
         this.inIFrame = window.parent !== window;
+        this._portalService.getToken().subscribe(token => {
+            this.setAuthToken(token);
```
cindylovescoding, line 32: So the newlogic is: whenever portal sends a token back to the IFrame, this will be subscribe to that and reset the token?

PR #797, opened 2020-06-16
Fix for format issue in error message and timestamp format for logging

XiaoxuHU:
- Fix for error message in time picker in external UI should indicate the input time format should be "MM/DD/YY hh:mm"

- Fix for logging StartTime and EndTime as ISO string so that Kusto query can use it directly.

### DIFF
--- a/AngularApp/projects/diagnostic-data/src/lib/services/detector-control.service.ts
+++ b/AngularApp/projects/diagnostic-data/src/lib/services/detector-control.service.ts
@@ -79,17 +79,17 @@ export class DetectorControlService {
     let returnValue: string = '';
     this.timeRangeDefaulted = false;
     this.timeRangeErrorString = '';
-
+    let timeStringFormat = this.internalClient ? "YYYY-MM-DD hh:mm" : "MM/DD/YY hh:mm"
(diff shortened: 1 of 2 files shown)


AngularApp/projects/diagnostic-data/src/lib/services/detector-control.service.ts
```diff
@@ -79,17 +79,17 @@ export class DetectorControlService {
     let returnValue: string = '';
     this.timeRangeDefaulted = false;
     this.timeRangeErrorString = '';
-
+    let timeStringFormat = this.internalClient ? "YYYY-MM-DD hh:mm" : "MM/DD/YY hh:mm"
```
cindylovescoding, line 82: Why we show different formatting between internal and external?

XiaoxuHU: Since the format for external is "MM/DD/YY hh:mm" that is because the Fabric Calendar only accept "MM/DD/YY" as input. But for Applens, the format is still the old version ("YYYY-MM-DD hh:mm")

PR #828, opened 2020-07-09
Fix accessibility bugs

XiaoxuHU:
Fix 3 accessibility bugs:

|||
-- | --
| [7084192](https://msazure.visualstudio.com/Antares/_workitems/edit/7084192) | [Keyboard Navigation - App Service Diagnostics - High CPU Analysis] After selection of "High CPU Analysis" focus indicator does not land on "High CPU Analysis" Pane. |
| [7614078](https://msazure.visualstudio.com/Antares/_workitems/edit/7614078) | [Supporting the platform - App Service Diagnostics - None] At zoom 400%,Controls and keyboard focus indicator are not appearing properly on the screen. |
| | Items in Side Nav not read |

For **7614078**, I change the width and height to be percentage instead of fixed,  so when zoom in, the side-nav will not take too much space. 
And also change some styles for tooltip in side-nav, so tooltip won't take extra place within side-nav, so that we can remove fixed height for side-nav items

### DIFF
--- a/AngularApp/projects/app-service-diagnostics/src/app/fabric-ui/components/category-summary/category-summary.component.scss
+++ b/AngularApp/projects/app-service-diagnostics/src/app/fabric-ui/components/category-summary/category-summary.component.scss
@@ -4,20 +4,21 @@
     }
 
   section.sidebar {
-      width: 265px;
+      width: 20%;
       position: fixed;
       top: 0px;
       bottom: 0px;
       border-right: 1px solid #eee;
       overflow-y: scroll;
+      overflow-x: hidden;
     }
 
   section.content {
     position: fixed;
-    width:calc(100% - 265px);
+    width: 80%;
     top: 0px;
     bottom: 0px;
-    left: 265px;
+    left: 20%;
     overflow-y: auto;
     overflow-x: hidden;
   }
--- a/AngularApp/projects/diagnostic-data/src/lib/components/detector-view/detector-view.component.ts
+++ b/AngularApp/projects/diagnostic-data/src/lib/components/detector-view/detector-view.component.ts
@@ -1,7 +1,7 @@
 import { Moment } from 'moment';
 import { BehaviorSubject } from 'rxjs';
 import { animate, state, style, transition, trigger } from '@angular/animations';
-import { Component, Inject, Input, OnInit, Output, EventEmitter, Pipe, PipeTransform, ViewChild, ElementRef, AfterViewChecked } from '@angular/core';
+import { Component, Inject, Input, OnInit, Output, EventEmitter, Pipe, PipeTransform, SimpleChanges } from '@angular/core';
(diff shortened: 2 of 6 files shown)


AngularApp/projects/diagnostic-data/src/lib/components/detector-view/detector-view.component.ts
```diff
@@ -267,6 +267,14 @@ export class DetectorViewComponent implements OnInit {
             this.resetGlobals();
           }
         }
+
+        //After loading detectors, foucs indicator will land into detector title
+        //For now asynchronouslly to foucs after render, it should have better solution
+        setTimeout(() => {
```
ShekharGupta1988, line 273: this doesnt need any timeout value?

AngularApp/projects/app-service-diagnostics/src/app/fabric-ui/components/category-summary/category-summary.component.scss
```diff
@@ -4,20 +4,21 @@
     }
 
   section.sidebar {
-      width: 265px;
```
cindylovescoding, line 7: Why do we want to change this? This is set in a way to be consistent with ibiza side bar width

XiaoxuHU: For now the behavior for ibiza side bar is when it zoom in, it will hide and only show the content on right side. Do we want the same behavior as ibiza side bar ? 
Since for now if we try to zoom in the fixed width side bar will take most of place so it can't show the content properly. That's the accessibility team complain about.

AngularApp/projects/diagnostic-data/src/lib/components/detector-view/detector-view.component.ts
XiaoxuHU: No need , I just to want to make it asynchronously so that I can get the detector title element after Angular render it. 
I tried if I don't put the logic into **setTimeout**, then the code will run before the element rendered and it's not working

AngularApp/projects/diagnostic-data/src/lib/components/detector-view/detector-view.component.ts
```diff
@@ -267,6 +267,14 @@ export class DetectorViewComponent implements OnInit {
             this.resetGlobals();
           }
         }
+
+        //After loading detectors, foucs indicator will land into detector title
+        //For now asynchronouslly to foucs after render, it should have better solution
```
cindylovescoding, line 272: nitpick: typo here “fousc”

XiaoxuHU: Need to install a spell checker on VS Code....

PR #829, opened 2020-07-09
Generate new request id for each arm request retry

cindylovescoding:
Fix for this service health WI: https://dev.azure.com/app-service-diagnostics-portal/app-service-diagnostics-portal/_queries/edit/1337/?triage=true

### DIFF
--- a/AngularApp/projects/app-service-diagnostics/src/app/shared/services/arm.service.ts
+++ b/AngularApp/projects/app-service-diagnostics/src/app/shared/services/arm.service.ts
@@ -5,7 +5,7 @@ import { Subscription } from '../models/subscription';
 import { ResponseMessageEnvelope, ResponseMessageCollectionEnvelope } from '../models/responsemessageenvelope';
 import { AuthService } from '../../startup/services/auth.service';
 import { CacheService } from './cache.service';
-import { catchError, retry, map, retryWhen } from 'rxjs/operators';
+import { catchError, retry, map, retryWhen, delay } from 'rxjs/operators';
 import { HttpClient, HttpHeaders, HttpResponse, HttpErrorResponse } from '@angular/common/http';
 import { GenericArmConfigService } from './generic-arm-config.service';
 import { StartupInfo } from '../models/portal';
@@ -159,13 +159,15 @@ export class ArmService {
                 eventProps.requestId = requestId;
                 eventProps["retryCount"] = retryCount;
 
-                if (retryCount++ >= 2) {
-                    this.telemetryService.logEvent("RetryRequestFailed", eventProps);
-                    throw err;
-                }
+                return err.pipe(delay(1000), map(err => {
+                    if(retryCount++ >= 2){
+                        this.telemetryService.logEvent("RetryRequestFailed", eventProps);
+                        throw err;
+                    }
+                    this.telemetryService.logEvent("RetryRequestRoutingDetails", eventProps);
+                    return err;
+                }));
 
-                this.telemetryService.logEvent("RetryRequestRoutingDetails", eventProps);
-                return err;
             }),
             catchError(this.handleError)
         );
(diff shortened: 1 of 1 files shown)


AngularApp/projects/app-service-diagnostics/src/app/shared/services/arm.service.ts
```diff
@@ -138,17 +138,37 @@ export class ArmService {
...
         this.telemetryService.logEvent("RequestRoutingDetails", eventProps);
+
+        let requestHeaders = this.getHeaders(null, additionalHeaders);
         const request = this._http.get<ResponseMessageEnvelope<T>>(url, {
-            headers: this.getHeaders(null, additionalHeaders)
+            headers: requestHeaders
         }).pipe(
-            retry(2),
+            retryWhen(err => {
+                let requestId:string = Guid.newGuid();
```
ShekharGupta1988, line 154: will this break @nmallick1 's logic for Downtime logging? He passes the request id as part of the header.

ShekharGupta1988: Cindy - I am thinking another simple way could be :
1. If the request Id is present in the header, use it for the first invocation. For the next invocations in case of retry, append #1, #2 etc to the request Id. That should not break Nazeef's scenarios as well and will generate unique requestIds at backend. 

2. If the request Id header is not present, generate a request id like you are doing for the first invocation. In case of retry, follow the same approach of appending #1, #2 etc to the same request id.

Thoughts?

cindylovescoding: Thanks Shekhar!
@nmallick1  Could you help review this as well? I don't think this should break any changes 
@ShekharGupta1988 I am thinking since we are only using request id in "DiagnosticRole" table, do you think it makes sense we generate the request id itself in the backend and log it for individual request, instead of passing this from front end?

ShekharGupta1988: that is what we were doing before where the UI was not passing any request id and we were generating a unique request id in the backend. But with downtimes work, Nazeef wanted to easily find and correlate the requests generating from the downtimes UI whenever you change a downtime.

**In general, its a good idea to pass the client request id to server as it helps to correlate the requests.**

Another approach is we generate the request ids from the server and pass it as response header which is then logged by client. (but that might need geomaster change. Not sure)

cindylovescoding: > that is what we were doing before where the UI was not passing any request id and we were generating a unique request id in the backend. But with downtimes work, Nazeef wanted to easily find and correlate the requests generating from the downtimes UI whenever you change a downtime.
> 
> **In general, its a good idea to pass the client request id to server as it helps to correlate the requests.**
> 
> Another approach is we generate the request ids from the server and pass it as response header which is then logged by client. (but that might need geomaster change. Not sure)

Sounds good Shekhar! I will wait for Nazeef's review and change to appending request id format

nmallick1: Yeah, this will break some of the reporting that I have in place to check for impact analysis of downtime. I like @ShekharGupta1988's suggestion of appending the retry count to the original request id (i.e.. requestId#1, requestId#2 etc.). I can update the reporting to use startswith instead of using == and we will have unique request Id's and will not break the current reporting. We will also be able to track how many retries are happening in general rather easily.

nmallick1: Spoke to @cindylovescoding , I'll make a few modifications to the queries I use for impact analysis but this should be fine.

PR #831, opened 2020-07-10
Fix deeplinking from analysis to detector in drill down view

nmallick1:
Make downtimes shareable when opened in drill down mode. 
Maintain the current drill down view even when downtime selection is changed. The detector view updates to reflect the latest downtime selection instead of switching to observation list.

### DIFF
--- a/AngularApp/projects/diagnostic-data/src/lib/components/detector-list-analysis/detector-list-analysis.component.ts
+++ b/AngularApp/projects/diagnostic-data/src/lib/components/detector-list-analysis/detector-list-analysis.component.ts
@@ -349,7 +349,6 @@ export class DetectorListAnalysisComponent extends DataRenderBaseComponent imple
                                                 }
                                             });
                                         }
-                                        //console.log('I am here >>>>');
                                         return;
                                     } else {
                                         this.detectorEventProperties = {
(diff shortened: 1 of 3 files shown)


AngularApp/projects/diagnostic-data/src/lib/components/detector-list-analysis/detector-list-analysis.component.ts
```diff
@@ -266,7 +266,10 @@ export class DetectorListAnalysisComponent extends DataRenderBaseComponent imple
             this._activatedRoute.paramMap.subscribe(params => {
                 this.analysisId = (this.analysisId != 'searchResultsAnalysis' && !!params.get('analysisId')) ? params.get('analysisId') : this.analysisId;
                 this.detectorId = params.get(this.detectorParmName) === null ? "" : params.get(this.detectorParmName);
-                if (this.analysisId != 'searchResultsAnalysis') this.goBackToAnalysis();
+                if (this.detectorId == "" && !!this._activatedRoute.firstChild && this._activatedRoute.firstChild.snapshot.paramMap.has(this.detectorParmName) && this._activatedRoute.firstChild.snapshot.paramMap.get(this.detectorParmName).length > 1) {
+                    this.detectorId = this._activatedRoute.firstChild.snapshot.paramMap.get(this.detectorParmName);
```
rekhaswaminathan, line 270: Should it be `this._activatedRoute.firstChild.snapshot.paramMap.get(this.detectorParmName).length > 0` to check if it is not empty?

nmallick1: I'm going length >1 to check for a presence of a valid value. We do not expect a detector name to be one character length  long and this will also guard against the case when detector name is a valid string but just a space (by mistake) that can especially happen in case someone shares a link and people copy paste the link from a URL. I've copied an extra space numerous times :)

PR #843, opened 2020-07-20
Nmallick1/UI null ref exception fix

nmallick1:


cindylovescoding:
@nmallick1 why do we need to update shrinkwrap too?

nmallick1:
> @nmallick1 why do we need to update shrinkwrap too?

My local build was failing with the latest highcharts update untill I did an npm install. I "think" this update will save others from that build failure. I might be mistaken though in my understanding of shrinkwrap js file.

PR #866, opened 2020-08-06
Nmallick1/Reduce logging for request routed

nmallick1:
Move logging inside cache service when the call to API is actually made. Avoid logging when requests are served from in-flight observable or from cached response.

### DIFF
--- a/AngularApp/projects/app-service-diagnostics/src/app/shared/services/cache.service.ts
+++ b/AngularApp/projects/app-service-diagnostics/src/app/shared/services/cache.service.ts
@@ -3,16 +3,10 @@ import { throwError as observableThrowError, Observable, Subject, of } from 'rxj
 import { Injectable } from '@angular/core';
 import { tap, finalize } from 'rxjs/operators';
 import { PortalKustoTelemetryService } from './portal-kusto-telemetry.service';
+import { TelemetryPayload } from 'diagnostic-data';
 
 interface CacheContent {
     value: any;
-}
-
-export interface LogInfo {
-    eventMessage: string,
-    properties: {
-        [name: string]: string
-    }
 }
 
 @Injectable()
@@ -25,7 +19,7 @@ export class CacheService {
 
     constructor(private telemetryService?: PortalKustoTelemetryService) { }
 
-    get(key: string, fallback?: Observable<any>, invalidateCache: boolean = false, logDataForActualCall?: LogInfo): Observable<any> | Subject<any> {
+    get(key: string, fallback?: Observable<any>, invalidateCache: boolean = false, logDataForActualCall?: TelemetryPayload): Observable<any> | Subject<any> {
 
         if (this.has(key)) {
             if (invalidateCache) {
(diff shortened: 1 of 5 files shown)


AngularApp/projects/app-service-diagnostics/src/app/shared/services/cache.service.ts
```diff
@@ -1,12 +1,20 @@
...
+import { throwError as observableThrowError, Observable, Subject, of } from 'rxjs';
 import { Injectable } from '@angular/core';
 import { tap, finalize } from 'rxjs/operators';
+import { PortalKustoTelemetryService } from './portal-kusto-telemetry.service';
 
 interface CacheContent {
     value: any;
 }
 
+export interface LogInfo {
```
ShekharGupta1988, line 11: Do you expect this model class can be reused in other areas in future apart from Cache service. If yes, we prob want to move this as a separate model ts file.

ShekharGupta1988: Also, nitpick : do you want to call it as "LogPayload" instead of LogInfo. :)
LogInfo seems like an action to log informational event (similarly like LogWarning, LogError etc)

nmallick1: Maybe @cindylovescoding  can comment if the structure is consistent with logging elsewhere. If it is, then it should certainly  reside as a stand alone model.

AngularApp/projects/applens/src/app/shared/services/cache.service.ts
```diff
@@ -1,14 +1,22 @@
...
+import { tap } from 'rxjs/operators';
 import { Injectable } from '@angular/core';
 import { environment } from '../../../environments/environment';
+import { TelemetryService } from 'diagnostic-data';
 
 interface CacheContent {
     value: any;
 }
 
+export interface LogInfo {
```
ShekharGupta1988, line 13: same here. I think it will be good to move this  model as part of diagnostics-data. I beleive base logger classes are also present there.

AngularApp/projects/app-service-diagnostics/src/app/shared/services/cache.service.ts
nmallick1: @ShekharGupta1988  Updated the code, kindly review.

AngularApp/projects/applens/src/app/shared/services/diagnostic-api.service.ts
```diff
@@ -257,7 +259,7 @@ export class DiagnosticApiService {
     }
 
     let keyPostfix = internalClient === true ? "-true" : "-false";
-    return useCache ? this._cacheService.get(this.getCacheKey(method, path + keyPostfix), request, invalidateCache) : request;
+    return useCache ? this._cacheService.get(this.getCacheKey(method, path + keyPostfix), request, invalidateCache, logData) : request;
```
hforeste, line 262: I didn't look to the code to see if we ever pass `false` for using cache but in the case that we do you are intentionally planning not to log the request routing details?

nmallick1: Good point, updated.

PR #868, opened 2020-08-07
Bug fixes in the Crash Monitoring UX

puneetg1983:
Fixing issues based on testing done by Yun Jung.

@cindylovescoding - Please check the change to `category-nav.component.ts` and review closely to see if it can break anything.

The issue we found was that if a site name contains a tool name, the navigation tab on the left was getting pre selected. We ran in to this issue when we tested with a site called crashmonitoringtest and on that one, we could see that Crash Monitoring tile was also selected because the toolId matched crashmonitoring. Let me know if you need more context.


AngularApp/projects/app-service-diagnostics/src/app/home/components/category-nav/category-nav.component.ts
```diff
@@ -135,7 +135,7 @@ export class CategoryNavComponent implements OnInit {
                 title: 'Proactive Tools',
                 tools: this.siteFeatureService.proactiveTools.map(tool => {
                     let isSelected = () => {
-                        return this._route.url.includes("/" + tool.item.id);
+                        return this._route.url.endsWith("/" + tool.item.id);
```
nmallick1, line 138: Will the scenario where the URL is of the form /crashmonitoring/ work with this?

puneetg1983: so the _route.url here will always end with diagnostics/<toolname>for tools

AngularApp/projects/app-service-diagnostics/src/app/shared/components/tools/crash-monitoring/crash-monitoring-analysis/crash-monitoring-analysis.component.ts
```diff
@@ -103,19 +106,33 @@ export class CrashMonitoringAnalysisComponent implements OnInit, OnChanges, OnDe
...
-          let crashMonitoringDatas = this.parseData(rawTable.table);
-          this.dumpsCollected = crashMonitoringDatas.length;
-          this.populateInsights(crashMonitoringDatas);
-          if (this.dumpsCollected === this.crashMonitoringSettings.MaxDumpCount) {
-            this.monitoringEnabled = false;
-          }
+        if (this.blobSasUri) {
+          this.loadDetectorData(rawTable);
+        } else {
+          this.blobSasUriObservable.subscribe(resp => {
```
cindylovescoding, line 112: So the sas uri result will also be refreshed along with the table right?

AngularApp/projects/app-service-diagnostics/src/app/home/components/category-nav/category-nav.component.ts
```diff
@@ -153,7 +153,7 @@ export class CategoryNavComponent implements OnInit {
                 title: 'Diagnostic Tools',
                 tools: this.siteFeatureService.diagnosticTools.filter(tool => this.stackMatchedForTools(tool)).map(tool => {
                     let isSelected = () => {
-                        return this._route.url.includes("/" + tool.item.id);
+                        return this._route.url.endsWith("/" + tool.item.id);
```
cindylovescoding, line 156: Looks good to me! Initially I thought the url might include query parameters, but it doesn't look like so. Thanks for fixing this!

AngularApp/projects/app-service-diagnostics/src/app/shared/components/tools/crash-monitoring/crash-monitoring-analysis/crash-monitoring-analysis.component.ts
puneetg1983: So this was the real issue why I made this code change. We get the BlobSasUri in the ngOnInit of this component. Since this is a DAAS call, it might take longer than the call to the crashmonitoring detector. The detector is returning us only relative path but we need the BLOB SAS Uri so that we can construct the full link to the file. 

Without this change, the link was incomplete or empty. With this change, we just check if blobSasUri is empty, then we subscribe to the observer again to get this else we reuse the cached copy of the BlobSasUri.

If there is a better way to achieve this, I would love to know.

PR #872, opened 2020-08-11
Adding event message body to integrate with case submission flow

cindylovescoding:
Currently the new Frame control from case submission team won't be able to get the message if there is not message object payload. Adding the payload with event type so they can register listener on these events.


AngularApp/projects/app-service-diagnostics/src/app/startup/services/portal.service.ts
```diff
@@ -146,14 +147,14 @@ export class PortalService {
     }
 
     private iframeReceivedMsg(event: Event): void {
-
+        console.log("[iFrame] Received messsage", event);
```
ShekharGupta1988, line 150: nit: remove console logging

PR #876, opened 2020-08-13
Fix Cannot read property 'id' of undefined in feature service

XiaoxuHU:
Fix **Cannot read property 'id' of undefined  in feature service**
[USER STORY 1474](https://dev.azure.com/app-service-diagnostics-portal/app-service-diagnostics-portal/_workitems/edit/1474/)

I only modified code in **getCategoryIdByCategoryName** method in feature service, other places are just format issue.

For now, if can't match categoryId in category-summary component, it will hide side-nav bar.

image

### DIFF
--- a/AngularApp/projects/app-service-diagnostics/src/app/shared-v2/services/feature.service.ts
+++ b/AngularApp/projects/app-service-diagnostics/src/app/shared-v2/services/feature.service.ts
@@ -22,10 +22,10 @@ export class FeatureService {
   protected _features: Feature[] = [];
   protected _featureDisplayOrder = [];
   private categories: Category[] = [];
-  private _featureSub:BehaviorSubject<Feature[]> = new BehaviorSubject<Feature[]>([]);
-  protected isLegacy:boolean;
+  private _featureSub: BehaviorSubject<Feature[]> = new BehaviorSubject<Feature[]>([]);
+  protected isLegacy: boolean;
   constructor(protected _diagnosticApiService: DiagnosticService, protected _contentService: ContentService, protected _router: Router, protected _authService: AuthService,
-    protected _logger: TelemetryService, protected _siteService: SiteService, protected _categoryService: CategoryService, protected _activatedRoute: ActivatedRoute,protected _portalActionService:PortalActionService,protected versionTestService:VersionTestService) {
+    protected _logger: TelemetryService, protected _siteService: SiteService, protected _categoryService: CategoryService, protected _activatedRoute: ActivatedRoute, protected _portalActionService: PortalActionService, protected versionTestService: VersionTestService) {
     this.versionTestService.isLegacySub.subscribe(isLegacy => this.isLegacy = isLegacy);
     this._categoryService.categories.subscribe(categories => this.categories = categories);
     this._authService.getStartupInfo().subscribe(startupInfo => {
@@ -44,14 +44,14 @@ export class FeatureService {
                 clickAction: this._createFeatureAction(detector.name, detector.category, () => {
                   //Remove after A/B test
                   if (this.isLegacy) {
-                      if(detector.id === 'appchanges') {
-                          this._portalActionService.openChangeAnalysisBlade();
-                      } else {
-                        this._router.navigateByUrl(`resource${startupInfo.resourceId}/detectors/${detector.id}`);
-                      }
+                    if (detector.id === 'appchanges') {
+                      this._portalActionService.openChangeAnalysisBlade();
+                    } else {
+                      this._router.navigateByUrl(`resource${startupInfo.resourceId}/detectors/${detector.id}`);
+                    }
                   } else {
                     const categoryId = this.getCategoryIdByCategoryName(detector.category);
-                    this.navigatTo(startupInfo,categoryId,detector.id,DetectorType.Detector);
+                    this.navigatTo(startupInfo, categoryId, detector.id, DetectorType.Detector);
                   }
                 })
               });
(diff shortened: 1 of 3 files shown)


AngularApp/projects/app-service-diagnostics/src/app/shared-v2/services/feature.service.ts
```diff
@@ -149,47 +148,49 @@ export class FeatureService {
   }
 
   private getCategoryIdByCategoryName(name: string): string {
-    let categoryId: string;
+    //Default set to first category id, to make sure routing is working
+    let categoryId: string = this.categories[0].id;
```
cindylovescoding, line 152: Will categories be empty here when you call this method?

AngularApp/projects/app-service-diagnostics/src/app/shared-v2/services/feature.service.ts
```diff
@@ -149,47 +148,49 @@ export class FeatureService {
...
       const category = this.categories.find(category => category.name === name);
       categoryId = category.id;
     }
+    //In category-overview page and uncategoried detector,return current categoryId
+    else if (currentCategoryId) {
+      categoryId = currentCategoryId;
+    }
     //In home page,no categoryId in router,return category as availability&perf
-    else if (!currentCategoryId){
+    else if (this.categories.find(category => category.name === "Availability and Performance")) {
```
cindylovescoding, line 168: Are we sure that all the partner teams will have "Availabity and Performance" category?

AngularApp/projects/app-service-diagnostics/src/app/shared-v2/services/feature.service.ts
```diff
@@ -149,47 +148,49 @@ export class FeatureService {
...
 
   private getCategoryIdByCategoryName(name: string): string {
-    let categoryId: string;
+    //Default set to first category id, to make sure routing is working
+    let categoryId: string = this.categories[0].id;
     const currentCategoryId = this._activatedRoute.root.firstChild.firstChild.firstChild.firstChild.snapshot.params["category"];
     //If category name is "XXX Tools" then should belong to Diagnostic Tool Category
-    if (name && name.includes('Tools')) {
-      categoryId = 'DiagnosticTools';
+    if (name && name.includes('Tools') && this.categories.find(category => category.name === "Diagnostic Tools")) {
```
cindylovescoding, line 155: Should we still check name.includes('Tools') if you are already checking this.categories.find(category => category.name === "Diagnostic Tools")?

AngularApp/projects/app-service-diagnostics/src/app/shared-v2/services/feature.service.ts
XiaoxuHU: Not sure. that's why I need to set the default categoryId to be the first one

AngularApp/projects/app-service-diagnostics/src/app/shared-v2/services/feature.service.ts
XiaoxuHU: I update so even if categories is empty and it in landing page(cannot get categoryId from Url), at least routing is working to open a new blade and render detector

AngularApp/projects/app-service-diagnostics/src/app/shared-v2/services/feature.service.ts
```diff
@@ -151,8 +151,8 @@ export class FeatureService {
     //Default set to first category id, to make sure routing is working
     let categoryId: string = this.categories.length > 0 ? this.categories[0].id : "";
     const currentCategoryId = this._activatedRoute.root.firstChild.firstChild.firstChild.firstChild.snapshot.params["category"];
-    //If category name is "XXX Tools" then should belong to Diagnostic Tool Category
```
cindylovescoding, line 154: Actually this should work for dedicated function app as well

XiaoxuHU: if category is empty string, routing will still working just no side nav and command-bar
image

AngularApp/projects/app-service-diagnostics/src/app/shared-v2/services/feature.service.ts
XiaoxuHU: I think we need, for Windows Web App, the feature categories will be Diagnostic Tool, Proactive Tool, Support Tool. All these three will be route to under DiagnosticTool category.

I update code to make sure only if  they have the same pattern as Windows App(categorized as three tools above and has DiagnosticTool in as one of their categories) will route into DiagnosticTool category

AngularApp/projects/app-service-diagnostics/src/app/shared-v2/services/feature.service.ts
cindylovescoding: I think this is a bug. @XiaoxuHU

XiaoxuHU: The screenshot is not repro for exception. It's when we return as empty string for **getCategoryIdByCategoryName** method.

I think that's not a bug. I check the logs, it seems like when exception('id' of undefined) happens it won't even open a new blade for detector after they clicked searched result.

cindylovescoding: We should fix the new exception on showing side nav and command bar. Otherwise this is just transfering  exception from one place to another.

XiaoxuHU:
> We should fix the new exception on showing side nav and command bar when the category id is empty. Otherwise this is just transfering exception from one place to another.

I fixed to show command bar in detector page

PR #881, opened 2020-08-14
Nationalclouddev

hforeste:
Enable other clouds the ability to run and compile their detectors and find out why it is failing


ShekharGupta1988:
@hforeste - since the detector changes made in N-Clouds will not sync back to public applens, should we add a warning for NClouds stating that you would need to take the changes and re-publish the detector in public version of applens as well?

PR #885, opened 2020-08-21
Moving AI Controller code to frontend due to ARM token restrictions

puneetg1983:
With this PR:-
1. Moving the Code from AppInsights Backend API controller over to FrontEnd project
2. Addressing feedback from Designer for Crash monitoring UI


ShekharGupta1988:
@puneetg1983 - while you are at it, can you also hide the app insights notification message if App Insights is already connected with our system?

puneetg1983:
Hey @ShekharGupta1988 - Any reason why we don't want to show anything? IMO its better to give a message that it is integrated with ASD but love to hear your thoughts more.

ShekharGupta1988:
> Hey @ShekharGupta1988 - Any reason why we don't want to show anything? IMO its better to give a message that it is integrated with ASD but love to hear your thoughts more.

Actuall nvm for this PR. We'll discuss it later and take the changes(if any) in another PR. I am reviewing your PR now.

Backend/Services/AppInsightsService/AppInsightsService.cs
```diff
@@ -38,196 +29,30 @@ private HttpClient _httpClient
...
-        }
-
-        private async Task<JObject> GetExistingTagsAsync(string resourceId, string bearerToken)
-        {
+            var query = WebUtility.UrlEncode("requests|take 1");
             HttpRequestMessage request = new HttpRequestMessage
             {
                 Method = HttpMethod.Get,
-                RequestUri = new Uri($"https://{_armEndpoint}{resourceId}?api-version=2018-02-01"),
+                RequestUri = new Uri($"https://api.applicationinsights.io/v1/apps/{appInsightsAppId}/query?timespan=1H&query={query}")
```
ShekharGupta1988, line 45: This is good :) (Forgot we can directly hit app insights API)

BTW Puneet - is this feature expected to work for Nclouds? Earlier since we were using ARM url everywhere, I think it would have been functional. Just check once and either disable the feature for NClouds or have this API endpoint configurable per NCloud.

puneetg1983:
Hey @cindylovescoding - Can you check once and confirm that changes to `AngularApp/projects/app-service-diagnostics/src/app/shared/services/arm.service.ts` wont break any logic to your error logging / handling. There are many more calls to HandleError now and I want to make sure that you are aware and we are not breaking anything.

Backend/Services/AppInsightsService/AppInsightsService.cs
puneetg1983: yea, let me find this information. We were using this endpoint earlier too but we never validated this to work in n-clouds.

puneetg1983: I would sync up with Hawk and see if we can find this information. So far, I am not able to find this info by searching.

PR #889, opened 2020-08-28
Add tooltip for form input

rekhaswaminathan:
Adding tooltip for form inputs:
image
image

Allow icon as input:
image

### DIFF
--- a/AngularApp/projects/diagnostic-data/src/lib/components/form/form.component.html
+++ b/AngularApp/projects/diagnostic-data/src/lib/components/form/form.component.html
@@ -8,7 +8,7 @@
         <div class="col-sm-8">
           <label class="col-sm-4 col-md-2 col-form-label"> {{formInput.inputLabel}}
             <fab-tooltip-host  *ngIf="formInput.tooltip != ''" [content]="formInput.tooltip" [directionalHint]="directionalHint">
-              <span class="fa form-tooltip fa-info-circle"></span>
+              <span class="fa form-tooltip" [ngClass]="formInput.tooltipIcon"></span>
             </fab-tooltip-host>
           </label>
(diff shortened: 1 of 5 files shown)


puneetg1983 (approved):

ShekharGupta1988:
Hi Rekha, couple of things:
1. Can we make sure basic accessibility is intact with this PR.
2. We should prob use the tooltip that Fabric offers. In general, it would be better as that control is built with all the default functionality like accessibility, responsiveness etc.

puneetg1983:
+1 to what @ShekharGupta1988 said.

AngularApp/projects/diagnostic-data/src/lib/components/form/form.component.html
```diff
@@ -6,23 +6,28 @@ <h3 class="form-title">{{element.formTitle}}</h3>
     <ng-template ngFor let-formInput [ngForOf]="element.formInputs">
       <div class="form-group row" *ngIf="isText(formInput.inputType)">
         <div class="col-sm-8">
-          <label class="col-sm-4 col-md-2 col-form-label"> {{formInput.inputLabel}} </label>
+          <label class="col-sm-4 col-md-4 col-form-label"> {{formInput.inputLabel}}
+            <span class="fa form-tooltip fa-info-circle" *ngIf="formInput.tooltip != ''"> <span class="form-tooltiptext"> {{formInput.tooltip}} </span> </span>
```
nmallick1, line 10: Can we take the tooltip icon as an optional input from the rendering type defaulting to what what you already have? It can be a fa icon in string format. We do not need to validate the class, if it is invalid, the icon rendering will fail.  It'll allow the user to customize the icon displayed.

Not really important, but just some flexibility.

PR #891, opened 2020-08-31
Crash Monitoring local cache changes

puneetg1983:
With this PR, introducing these changes.

1. If an app has Local Cache enabled, they will see a warning message (below) when they try to enable Crash Monitoring
2. Added code in our project to generate SAS URI in the angular project and then store it as an App Setting (`WEBSITE_DAAS_STORAGE_SASURI`). We will use this code path when ANT90 is deployed. Backend App Service PR in progress. Will share once I have tested it fully.
3. Add a small workaround to fix the Profiler checking for .NET Core 3.1 version. The change is not 100% accurate but we are waiting on .net core team to confirm. Till then, this can disable the warnings for customers who are on .NET CORE 3.1

image


AngularApp/projects/app-service-diagnostics/src/app/fabric-ui/components/create-storage-account-panel/create-storage-account-panel.component.ts
```diff
@@ -195,6 +203,23 @@ export class CreateStorageAccountPanelComponent implements OnInit {
       });
   }
 
+  generateSasKey(storageAccountId: string, storageAccountName: string, storageKey: string) {
+    this._storageService.generateSasKey(storageAccountId, storageKey).subscribe(generatedSasUri => {
+      if (generatedSasUri) {
+        let storageAccountProperties: StorageAccountProperties = new StorageAccountProperties();
+        storageAccountProperties.name = storageAccountName;
+        storageAccountProperties.sasUri = `https://${storageAccountName}.blob.${this._armService.storageUrl}/${this._daasService.defaultContainerName}?${generatedSasUri}`;
+        console.log(storageAccountProperties.sasUri);
```
cindylovescoding, line 212: Remove console logging?

PR #897, opened 2020-09-08
IFrame integration hackathon for case submission flow

cindylovescoding:
This logic is integrated behind a feature flag, will not have any production impact.

### DIFF
--- a/AngularApp/projects/app-service-diagnostics/src/app/startup/services/portal.service.ts
+++ b/AngularApp/projects/app-service-diagnostics/src/app/startup/services/portal.service.ts
@@ -153,7 +153,12 @@ export class PortalService {
 
         const data = event.data.data;
         const methodName = event.data.kind;
-        console.log('[iFrame] Received validated mesg: ' + methodName, event);
+        console.log('[iFrame] Received validated mesg: ' + methodName, event, event.srcElement, event.srcElement.location, event.srcElement.location.host);
+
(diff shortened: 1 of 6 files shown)


AngularApp/projects/app-service-diagnostics/src/app/startup/services/portal.service.ts
```diff
@@ -30,6 +32,8 @@ export class PortalService {
         this.tokenObservable = new ReplaySubject<string>(1);
         this.appInsightsResourceObservable = new ReplaySubject<any>(1);
 
+        this.isIFrameForCaseSubmissionSolution = new ReplaySubject<boolean>(1);
```
XiaoxuHU, line 35: Nitpick, do we need to initialize  isIFrameForCaseSubmissionSolution as **ReplaySubject<boolean>(true)** ?

XiaoxuHU (approved):

AngularApp/projects/app-service-diagnostics/src/app/startup/services/portal.service.ts
cindylovescoding: This is not the initialized value, it's the buffer size of the replay subject:

It buffers a set number of values and will emit those values immediately to any new subscribers in addition to emitting new values to existing subscribers.

https://rxjs-dev.firebaseapp.com/api/index/class/ReplaySubject

PR #906, opened 2020-09-16
Dynamic adjust table height and check PII info 

XiaoxuHU:
I change the logic so that if estimated table height is less than threshold(50% of window height)and height is not passed from detector code, it will be set by Fabric UI. Otherwise we will override table height as before.

I also added PII replacement for feedback input

### DIFF
--- a/AngularApp/projects/diagnostic-data/src/lib/components/data-table-v4/data-table-v4.component.ts
+++ b/AngularApp/projects/diagnostic-data/src/lib/components/data-table-v4/data-table-v4.component.ts
@@ -37,11 +37,16 @@ export class DataTableV4Component extends DataRenderBaseComponent implements Aft
       }
     }
 
-    let tableHeight = `${this.calculateTableHeight()}px`;
+    let tableHeight = "";
+    if (this.estimateTableHeight() >= this.heightThreshold) {
+      tableHeight = `${this.heightThreshold}px`;
+    }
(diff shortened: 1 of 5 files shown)


AngularApp/projects/diagnostic-data/src/lib/components/data-table-v4/data-table-v4.component.ts
```diff
@@ -37,11 +37,16 @@ export class DataTableV4Component extends DataRenderBaseComponent implements Aft
...
-    let tableHeight = `${this.calculateTableHeight()}px`;
+    let tableHeight = "";
+    if (this.estimateTableHeight() >= this.heightThreshold) {
+      tableHeight = `${this.heightThreshold}px`;
+    }
     if (this.renderingProperties.height) {
       tableHeight = this.renderingProperties.height;
     }
-    this.fabDetailsList.styles = { root: { height: tableHeight } };
+    if (tableHeight !== "") {
```
cindylovescoding, line 47: What if tableHeight == "" as initiated (The two if conditions are not met above), will any height be set to fabDetailList?

XiaoxuHU: Yes, correct. DetailList will take care for height

cindylovescoding (approved):

PR #925, opened 2020-10-01
Updated Deep Search to be enabled only for specific Products

vijaysaayi:
- Added a new Config file which has a list of Products for which Deep Search is enabled
- Added methods in component to get PesId 
- Deep Search's IsEnabled method considers PesId and IsPublic


SharmaAjay19 (approved):
Nice! 👍

PR #933, opened 2020-10-07
Validate host name for security

XiaoxuHU:


AngularApp/projects/app-service-diagnostics/src/app/startup/services/portal.service.ts
```diff
@@ -25,6 +25,15 @@ export class PortalService {
     private shellSrc: string;
     private tokenObservable: ReplaySubject<string>;
 
+    private acceptedOriginsSuffix = [
```
ShekharGupta1988, line 28: do we need to ms.portal as well?

XiaoxuHU: Not sure... I just copied this array from Elliott code

https://github.com/Azure/azure-functions-ux/blob/7e2a387b1b50f659afbde77d5379ab643c8b58b5/client-react/src/portal-communicator.ts#L416

What is the url for ms portal ?

ehamai: No you shouldn't need it since you're just doing a suffix match

ShekharGupta1988 (approved):

hforeste (approved):

PR #934, opened 2020-10-09
Fix calculateFabWidth and add exception for "Missing Subscription" 

XiaoxuHU:
- This PR should fix calculateFabWidth length undefined issue
- For "Missing Subscription" bug, it should throw exception and not make Arm Call when resource Uri is empty. This PR will not fix this issue but we can call stack when happens again and do further investigation

### DIFF
--- a/AngularApp/projects/app-service-diagnostics/src/app/shared/services/arm.service.ts
+++ b/AngularApp/projects/app-service-diagnostics/src/app/shared/services/arm.service.ts
@@ -121,6 +121,9 @@ export class ArmService {
     }
 
     createUrl(resourceUri: string, apiVersion?: string) {
+        if(resourceUri === "/"){
+            throw new Error("Empty ResourceUri for ARM Call");
+        }
         return `${this.armUrl}${resourceUri}${resourceUri.indexOf('?') >= 0 ? '&' : '?'}` +
             `api-version=${this.getApiVersion(resourceUri, apiVersion)}`;
     }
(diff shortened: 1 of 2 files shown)


AngularApp/projects/app-service-diagnostics/src/app/shared/services/arm.service.ts
```diff
@@ -121,6 +121,9 @@ export class ArmService {
     }
 
     createUrl(resourceUri: string, apiVersion?: string) {
+        if(resourceUri === "/"){
+            throw new Error("Empty ResourceUri for ARM Call");
```
cindylovescoding, line 125: Is this gonna be an unhandled error? What will happen on D&S portal if we get to this scenario?

XiaoxuHU: I am not able to repro this issue. But from AppInsight, I think it will show as blank or failed message, I saw in a categoryoverview page when this exception happened, the next customer event is **CategorySelect**.

I adding this is because for now we do not have meaningful stack trace for this exception. We need full picture about which function called **createUri** with empty resource Uri so we can investigate further to the root cause

XiaoxuHU: I tested when throw the unhandled error, it will not send request to ARM, thus the corresponding response will not rendering(like detector view or detector list, ect). This error should not able to make whole page into blank.

AngularApp/projects/diagnostic-data/src/lib/components/dropdown-v4/dropdown-v4.component.ts
```diff
@@ -111,6 +111,10 @@ export class DropdownV4Component extends DataRenderBaseComponent {
         length = Math.max(length, option.text.length);
       }
     });
-    return (length + this.label.length) * 11 ;
+    let lableLength = 0;
+    if(this.label && this.label.length > 0) {
+      lableLength = this.label.length;
```
cindylovescoding, line 116: Nitpick: Typo

cindylovescoding (approved):

PR #942, opened 2020-10-14
Fix exception for ‘Uncaught (in promise): Code:MissingSubscription

XiaoxuHU:
Try to fix [1512](https://dev.azure.com/app-service-diagnostics-portal/app-service-diagnostics-portal/_workitems/edit/1512) ‘Uncaught (in promise): Code:MissingSubscription The request did not have a subscription or a valid tenant level resource provider. at M’
All Exceptions are from **registerResource** in resource.resolver
If not able to get resourceUri from activatedRouteSnapshot, then try to get from RouterStateSnapshot. 
If both are not working then throw exception message for next step debug

image

### DIFF
--- a/AngularApp/projects/app-service-diagnostics/src/app/home/resolvers/resource.resolver.ts
+++ b/AngularApp/projects/app-service-diagnostics/src/app/home/resolvers/resource.resolver.ts
@@ -21,13 +21,23 @@ export class ResourceResolver implements Resolve<Observable<{} | ArmResource>> {
             .filter(x => x.path !== 'new' && x.path !== 'categories')
             .map(x => x.path)
             .join('/');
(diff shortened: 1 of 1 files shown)


ShekharGupta1988:
this is lovely investigation @XiaoxuHU . This error was particularly hard to chase and you had to chase the fix in multiple steps.

ShekharGupta1988 (approved):

cindylovescoding (approved):

XiaoxuHU:
> this is lovely investigation @XiaoxuHU . This error was particularly hard to chase and you had to chase the fix in multiple steps.

Thanks for @cindylovescoding integrated AppInsight within our App that I can find stack trace when throw an exception. I am not sure this fix will be working or not but at least we can get more information about this exception for further investigation

PR #953, opened 2020-10-28
Fix for cannot read property 'replace' of null at t.project

XiaoxuHU:


cindylovescoding (approved):
Thanks for the quick fix xiaoxu!

PR #956, opened 2020-10-29
Remove duplicate tools in category-nav

XiaoxuHU:
This is for fixing duplicate tools in category-nav component. 

This is caused by not filtering stack for tools and I added **stackMatchedForTools()** method back to resolve issue.

Before fixing,
image

After fixing,
image

### DIFF
--- a/AngularApp/projects/app-service-diagnostics/src/app/home/components/category-nav/category-nav.component.ts
+++ b/AngularApp/projects/app-service-diagnostics/src/app/home/components/category-nav/category-nav.component.ts
@@ -127,7 +127,7 @@ export class CategoryNavComponent implements OnInit {
             stack: '',
             item: {
                 title: 'Proactive Tools',
-                tools: this.siteFeatureService.proactiveTools.map(tool => {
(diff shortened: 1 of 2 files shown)


ShekharGupta1988 (approved):

cindylovescoding:
I am thinking the right way to fix this is probably add the filter logic in siteFeatureService.
Also you only added this for diagnostic tools and support tools, and leave the behavior a little bit inconsistent with line 138 for "proactiveTools"

XiaoxuHU:
> I am thinking if the right way to fix this is to add the filter logic in siteFeatureService.
> Also you only added this for diagnostic tools and support tools, and leave the behavior a little bit inconsistent with line 138 for "proactiveTools"
> 
> Please help verify this is working for premiumV3 sku.

Yes, make sense.  I can move this method to site-feature service and also add into proactive tools

BTW, how to verify whether is working with P3 SKU ?
I created a test App with P3V2 SKU, and below is the diagnostic tools page looks like, **dose the side nav show all tools for P3** ?

image

image

image

cindylovescoding (approved):

PR #961, opened 2020-11-04
Provide resource service to dynamic analysis in genie

SharmaAjay19:
image


cindylovescoding (approved):
Thanks for the quick turnaround Ajay!

PR #973, opened 2020-11-17
fix createFlowForCategory undefined

XiaoxuHU:


AngularApp/projects/app-service-diagnostics/src/app/supportbot/message-flow/v2-flows/generic-category.flow.ts
```diff
@@ -51,7 +51,7 @@ export class GenericCategoryFlow extends IMessageFlowProvider {
   }
 
   createMessageFlowForCategory(category: Category): Observable<MessageGroup[]> {
-    if (!category.createFlowForCategory || this.categoriesCreated.indexOf(category) >= 0) { return of([]); }
+    if (!category || !category.createFlowForCategory || this.categoriesCreated.indexOf(category) >= 0) { return of([]); }
```
cindylovescoding, line 54: I don't know this is used in our system or not
Can you double check you are fixing the right service?

XiaoxuHU: image

I checked we still call this service in old UI when init category page

cindylovescoding (approved):

PR #975, opened 2020-11-18
Changing Storage account creation to polling based mechanism

puneetg1983:
With this PR, there are 3 main changes
- Switching to a polling based mechanism while Storage Account creation. This is to address the issue `Code:StorageAccountOperationInProgress An operation is currently performing on this storage account that requires exclusive access.`
- Removing the PHP Process Analyzer tool as that is no more there in DAAS. It had zero usage and was buggy and was removed from DAAS in the last update.
- Ensuring that the logging inside `handleError `method works by ensuring that `this `object is available. Following the recommendation given in [this](https://stackoverflow.com/questions/52020486/cant-use-services-inside-a-catcherror-function) article. @cindylovescoding - Please check this one.


AngularApp/projects/app-service-diagnostics/src/app/shared/services/storage.service.ts
```diff
@@ -38,13 +39,16 @@ export class StorageService {
...
     let requestBody = new NewStorageAccount();
     requestBody.location = location;
-    return this._armClient.putResourceWithoutEnvelope<any, NewStorageAccount>(url, requestBody, this.apiVersion)
-      .pipe(map(response => {
-        return response;
+    return this._armClient.putResourceFullResponse(url, requestBody, true, this.apiVersion).pipe(
+      map((response: HttpResponse<{}>) => {
+        let locationHeader = response.headers.get('Location');
+        if (locationHeader != null) {
+          return locationHeader;
```
hforeste, line 50: Why not return response as opposed to just the `locationHeader` like you had before so that tomorrow if you need some other property from the response you wouldn't have to change any code here.

hforeste (approved):

cindylovescoding (approved):

AngularApp/projects/app-service-diagnostics/src/app/shared/services/storage.service.ts
puneetg1983: This makes the caller code simple so leaving it as is for now. If this requirement comes, will change this piece.

PR #986, opened 2020-12-10
openShiftManagedClusters was incorrectly mapped

khaled-zayed:
Moved it from Microsoft.ContainerService to Microsoft.RedHatOpenShift

### DIFF
--- a/AngularApp/projects/applens/src/assets/enabledResourceTypes.json
+++ b/AngularApp/projects/applens/src/assets/enabledResourceTypes.json
@@ -51,14 +51,14 @@
         },
         {
             "service": "ResourceService",
-            "resourceType": "Microsoft.ContainerService/openShiftManagedClusters",
+            "resourceType": "Microsoft.RedHatOpenShift/openShiftManagedClusters",
             "templateFileName": "OpenShiftManagedClusters",
             "imgSrc": "assets/img/Azure-KubernetesService-Logo.png",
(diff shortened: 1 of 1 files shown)


nmallick1 (approved):

AngularApp/projects/applens/src/assets/enabledResourceTypes.json
```diff
@@ -51,14 +51,14 @@
         },
         {
             "service": "ResourceService",
-            "resourceType": "Microsoft.ContainerService/openShiftManagedClusters",
+            "resourceType": "Microsoft.RedHatOpenShift/openShiftManagedClusters",
```
nmallick1, line 54: Can you check if their detectors are also using the earlier provider name? If so, we might have to update all their detectors with the new resource type string.

khaled-zayed: I just fixed their detector/gist templates:
https://github.com/Azure/Azure-AppServices-Diagnostics/blob/main/data/templates/Detector_OpenShiftManagedClusters.csx & https://github.com/Azure/Azure-AppServices-Diagnostics/blob/main/data/templates/Gist_OpenShiftManagedClusters.csx

They have only 1 detector, which I'll fix once this PR is merged.

XiaoxuHU (approved):

PR #991, opened 2020-12-22
Fix api version for arm resource type

cindylovescoding:


XiaoxuHU (approved):

AngularApp/projects/app-service-diagnostics/src/app/shared/components/support-tools/support-tools.component.ts
```diff
@@ -32,7 +32,7 @@ export class SupportToolsComponent {
 
         this._authService.getStartupInfo()
             .subscribe((startUpInfo: StartupInfo) => {
-                if (startUpInfo.resourceType === ResourceType.Site) {
+                if (startUpInfo.resourceType === ResourceType.Site && startUpInfo.resourceId.toLowerCase().includes("/providers/microsoft.web/")) {
```
puneetg1983, line 35: Wouldn’t a better fix be to add one more type (like Other) to ResourceType enum and then improve this logic in auth.service.ts?                    

```
if (!this.resourceType) {
    this.resourceType = info.resourceId.toLowerCase().indexOf('hostingenvironments') > 0 ? ResourceType.HostingEnvironment : ResourceType.Site;
}
info.resourceType = this.resourceType;
```


Here we can add the check for info.resourceId.toLowerCase().includes("/providers/microsoft.web/") and then assign this.resourceType to ResourceType.Site else set it to ResourceType.Other

That way we don’t have to change other child components .

AngularApp/projects/app-service-diagnostics/src/app/startup/services/auth.service.ts
```diff
@@ -66,9 +66,9 @@ export class AuthService {
                     if (!this.currentToken){
                         this.currentToken = info.token;
                     }
-                    
+
                     if (!this.resourceType) {
-                        this.resourceType = info.resourceId.toLowerCase().indexOf('hostingenvironments') > 0 ? ResourceType.HostingEnvironment : ResourceType.Site;
+                        this.resourceType = info.resourceId.toLowerCase().indexOf('hostingenvironments') > 0 ? ResourceType.HostingEnvironment : info.resourceId.toLowerCase().includes("/providers/microsoft.web/") ? ResourceType.Site : ResourceType.Other;
```
puneetg1983, line 71: Lets change `.indexOf('hostingenvironments')` to `.indexOf('/providers/microsoft.web/hostingenvironments/')`

puneetg1983 (approved):

PR #1012, opened 2021-01-23
Diagnose And Solve portal fix for Service Fabric

nmallick1:
Fix config for Service Fabric and SignalR. Also, add logging in case of merging of config fails.

### DIFF
--- a/AngularApp/projects/app-service-diagnostics/src/app/shared/services/generic-arm-config.service.ts
+++ b/AngularApp/projects/app-service-diagnostics/src/app/shared/services/generic-arm-config.service.ts
@@ -65,7 +65,7 @@ export class GenericArmConfigService {
     return null;
   }
 
-  public distinctStringValues(value:string, index: number, self: string[]): boolean {
+  public distinctStringValues(value: string, index: number, self: string[]): boolean {
     return self.indexOf(value) === index;
   }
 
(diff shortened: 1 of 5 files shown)


AngularApp/projects/app-service-diagnostics/src/app/shared/services/generic-arm-config.service.ts
```diff
@@ -112,268 +112,337 @@ export class GenericArmConfigService {
...
-          currConfig.homePageText.title = this.getValue(this.resourceConfig.homePageText.title, this.overrideConfig.homePageText.title);
+        try {
+          if (
+            (this.overrideConfig.homePageText && this.overrideConfig.homePageText.title) ||
+            (this.resourceConfig.homePageText && this.resourceConfig.homePageText.title)
+          ) {
+            currConfig.homePageText.title = this.getValue(this.resourceConfig.homePageText.title, this.overrideConfig.homePageText.title);
+          }
+        } catch (error) {
+          console.log(`Error while merging title for resource ${resourceUri}. Error : ${error}`);
```
ShekharGupta1988, line 123: remove console logging from prod

nmallick1: Had to explicitly push to Kusto and AppInsights due to cyclic dependency causing compilation failure while using telemetry service.

ShekharGupta1988 (approved):

PR #1013, opened 2021-01-25
Extend partner config to indicate whether the response returned by arm api is base64 encoded.

nmallick1:
Extend partner config to indicate whether the response returned by arm api is base64 encoded.
Also add support for keystoneDetectorId.
Update partner configs to reflect updated schema.

### DIFF
--- a/AngularApp/projects/app-service-diagnostics/src/app/shared/services/generic-arm-config.service.ts
+++ b/AngularApp/projects/app-service-diagnostics/src/app/shared/services/generic-arm-config.service.ts
@@ -620,14 +620,14 @@ export class GenericArmConfigService {
 
   getApiVersion(resourceUri: string): string {
     let apiVersion = '';
-    if (!!this.getArmApiConfig(resourceUri) && !!this.getArmApiConfig(resourceUri).armApiVersion && this.getArmApiConfig(resourceUri).armApiVersion !== '') {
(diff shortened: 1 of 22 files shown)


AngularApp/projects/app-service-diagnostics/src/app/shared/services/generic-arm-config.service.ts
```diff
@@ -529,26 +595,44 @@ export class GenericArmConfigService {
...
+    else {
+      return null;
+    }
   }
 
   getApiVersion(resourceUri: string): string {
     let apiVersion = '';
-    if (this.getArmResourceConfig(resourceUri)) {
-      apiVersion = this.getArmResourceConfig(resourceUri).armApiVersion;
+    if (!!this.getArmApiConfig(resourceUri) && !!this.getArmApiConfig(resourceUri).armApiVersion && this.getArmApiConfig(resourceUri).armApiVersion !== '') {
```
cindylovescoding, line 623: nitpick: this.getArmApiConfig(resourceUri).armApiVersion !== '' is already preconcluded in !!this.getArmApiConfig(resourceUri).armApiVersion right?

nmallick1: Good point.

PR #1016, opened 2021-01-26
Enable diagnose and solve apim

rzhang628:
Hi Team:

@cindylovescoding @nmallick1 @yunjchoi I have pushed the PR for APIM diagnose and solve blade, can you please review and let me know if there is anything? Thanks.


nmallick1 (changes requested):
We recently added additional features to our platform and have updated the config to reflect these features. Can you please pull from latest and make sure your changes do not overwrite the newly pushed config changes for armApiConfig and keystoneDetectorId?

PR #1021, opened 2021-01-29
Update category tile style

XiaoxuHU:
In this PR,
1) Update style for category tile so that if category tile text are not same length, each tile style will still be almost  align(tile height, tile title and tile link will be align. But tile description and keywords may not)
image

2) Add keywords for windows web app and diag tool 
image

### DIFF
--- a/AngularApp/projects/app-service-diagnostics/src/app/resources/web-sites/services/sites-category.service.ts
+++ b/AngularApp/projects/app-service-diagnostics/src/app/resources/web-sites/services/sites-category.service.ts
@@ -204,16 +204,6 @@ export class SitesCategoryService extends CategoryService {
   }
 
   private _getDiagnosticToolsCategory(siteId: string): SiteFilteredItem<Category> {
-    const appType = this._resourceService.appType;
-    const appSku = this._resourceService.sku;    
-
-    let diagToolKeywords: string[] = ['Profiler', 'Memory Dump', 'Auto-Heal'];
-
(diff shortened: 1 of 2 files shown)


ShekharGupta1988 (approved):

AngularApp/projects/app-service-diagnostics/src/app/resources/web-sites/services/sites-category.service.ts
```diff
@@ -204,6 +204,16 @@ export class SitesCategoryService extends CategoryService {
   }
 
   private _getDiagnosticToolsCategory(siteId: string): SiteFilteredItem<Category> {
+    const appType = this._resourceService.appType;
+    const appSku = this._resourceService.sku;    
+
+    let diagToolKeywords: string[] = ['Profiler', 'Memory Dump', 'Auto-Heal'];
+
+    //Add one more keyword to Dedicated Function App category page for better looking
+    if(appType === AppType.FunctionApp && appSku !== Sku.Dynamic) {
```
ShekharGupta1988, line 213: we should add this keyword for windows app as well

cindylovescoding (approved):

AngularApp/projects/app-service-diagnostics/src/app/resources/web-sites/services/sites-category.service.ts
XiaoxuHU: Sure, I will remove if check and just add it for all resources

PR #1035, opened 2021-02-11
Applens Incident Assistance

SharmaAjay19:
When someone creates an ICM on the team, we validate the incident and provide a link to Applens where they can correct the information. This PR is about the Applens page that helps in this process.
When customer lands on the Applens Incident Assistance page, they'll see the information on failed validations in the CRI.
image
Now the customer can use the hints/error messages provided here to correct the information and check if the validation succeeds. For example here they correct the app name to an app that actually exists.
image
Similary they can follow up for all fields and finally when they have corrected all fields, they see the "Update Incident" button is activated.
image

When the user clicks on "Update Incident" button, the incident in ICM is updated with the corrected information and ready for the team to investigate.
image


ShekharGupta1988 (approved):
Looks good Ajay. 
One small comment - Can you make sure we are logging enough events to catch situations where we are not able to get ICM details from remote endpoints or any controller exceptions and setup alerts for ourselves to catch this.

cindylovescoding:
Hi @SharmaAjay19  Could you explain a little more what is this feature about?

SharmaAjay19:
Hi guys, this PR is not complete yet. Please wait for the review, I will send the links.
Thanks

PR #1037, opened 2021-02-16
Adding AppSetting check WEBSITE_DAAS_STORAGE_SASURI

puneetg1983:
With this PR

- Adding functionality so that configuring storage account now configures the SAS URI in app setting
- Unified the choosing storage account experience in all places. Now everywhere, a side panel will be opened to choose a storage account
- With this change, Crash Monitoring can work with LocalCache so removed that check
- Removed the messaging in AutoHeal component for process frozen for memory dumps (as we create a clone)

Also fixing these

ID | Title | Iteration Path | State | Tags | Comments | Changed Date
-- | -- | -- | -- | -- | -- | --
1674 | [Diagnose and Solve client] Cannot read property 'indexOf' of undefined at Function.e.timespanToSeconds | app-service-diagnostics-portal\Backlog | Active | Service Health | 0 | 2/17/2021, 11:28:29 PM
1672 | [Diagnose and Solve Portal] Cannot read property 'indexOf' of undefined at n.isValidStatusCodesRange | app-service-diagnostics-portal\Backlog | Active | Service Health | 0 | 2/17/2021, 11:28:24 PM


ShekharGupta1988 (approved):

AngularApp/projects/app-service-diagnostics/src/app/shared/services/daas.service.ts
```diff
@@ -151,9 +153,51 @@ export class DaasService {
...
-    getBlobSasUri(site: SiteDaasInfo): Observable<DaasSettings> {
+    setBlobSasUriAppSetting(site: SiteDaasInfo, blobSasUri: string): Observable<any> {
+        return this._siteService.getSiteAppSettings(site.subscriptionId, site.resourceGroupName, site.siteName, site.slot).pipe(
+            map(settingsResponse => {
+                if (settingsResponse && settingsResponse.properties) {
+                    if (blobSasUri) {
+                        settingsResponse.properties['WEBSITE_DAAS_STORAGE_SASURI'] = blobSasUri;
+                    } else {
+                        if (settingsResponse.properties['WEBSITE_DAAS_STORAGE_SASURI']) {
+                            delete settingsResponse.properties['WEBSITE_DAAS_STORAGE_SASURI'];
```
cindylovescoding, line 164: Just curious, how does this "delete" work?

cindylovescoding (approved):

AngularApp/projects/app-service-diagnostics/src/app/shared/services/daas.service.ts
puneetg1983: So this is how Delete works - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete but it looks like this is a bad practice to be followed on arrays https://rules.sonarsource.com/typescript/RSPEC-2870 as it might lead an 'undefined' element in the arrya.

Let me improve this code.

puneetg1983: BTW, thanks for bringing this up @cindylovescoding

puneetg1983: Actually I am going ahead with the current delete implementation only. When I try to remove the array item it is giving some silly error. I will revisit this some time later.

PR #1049, opened 2021-03-01
Add WEBSITE_CRASHMONITORING_USE_DEBUGDIAG for Crash Monitoring

puneetg1983:


AngularApp/projects/app-service-diagnostics/src/app/shared/services/site.service.ts
```diff
@@ -250,6 +255,8 @@ export class SiteService {
                         if (blobSasUri) {
                             settingsResponse.properties['WEBSITE_DAAS_STORAGE_SASURI'] = blobSasUri;
                         }
+                        settingsResponse.properties['WEBSITE_CRASHMONITORING_USE_DEBUGDIAG'] = true;
```
cindylovescoding, line 258: Hi @puneetg1983 , do we always set this to be true?

puneetg1983: Yes, If this app setting is present and if it is set to TRUE, only then DebugDiag will be used instead of ProcDump

ShekharGupta1988 (approved):

PR #1056, opened 2021-03-08
Exclude detector types 

XiaoxuHU:
This PR is to exclude some detector types from searching and showing in side navigation for category overview page. 

For now, I exclude category overview detector. In future, we can add more types to be excluded, like risk alert detector, notification detector, key stone detector...

### DIFF
--- a/AngularApp/projects/app-service-diagnostics/src/app/shared-v2/services/feature.service.ts
+++ b/AngularApp/projects/app-service-diagnostics/src/app/shared-v2/services/feature.service.ts
@@ -40,10 +40,7 @@ export class FeatureService {
           this._detectors = detectors;
           this.categories = categories;
           detectors.forEach(detector => {
-            if ((detector.category && detector.category.length > 0) ||
-              (detector.description && detector.description.length > 0) ||
-              this.checkAddDetector(detector)) {
-
+            if (this.checkAddDetector(detector)) {
               this._rewriteCategory(detector);
               if (detector.type === DetectorType.Detector) {
                 this._features.push(<Feature>{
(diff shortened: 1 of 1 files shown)


ShekharGupta1988 (approved):

AngularApp/projects/app-service-diagnostics/src/app/shared-v2/services/feature.service.ts
```diff
@@ -210,8 +213,14 @@ export class FeatureService {
     const bestPractices = "Best Practices";
     const riskAssessments = "Risk Assessments";
     //If category name is "Best Practice" and only has "Risk Assessment" category then rewrite category to "Risk Assessment"
-    if(detector.category === bestPractices && !this.categories.find(category => category.name === bestPractices) && this.categories.find(category => category.name === riskAssessments)) {
+    if (detector.category === bestPractices && !this.categories.find(category => category.name === bestPractices) && this.categories.find(category => category.name === riskAssessments)) {
       detector.category = riskAssessments;
     }
   }
+
+  private checkAddDetector(detector: DetectorMetaData): boolean {
```
cindylovescoding, line 221: Nitpick: this name is a little confusing. Can we rename this to be something like validateDetectorMetadata:)

cindylovescoding (approved):

AngularApp/projects/app-service-diagnostics/src/app/shared-v2/services/feature.service.ts
XiaoxuHU: Updated

PR #1083, opened 2021-03-24
Add Section rendering type

XiaoxuHU:
- This PR is for adding new rendering type to sectionalize our detectors, so we can sectionalize and collapsible some unimportant sections in detector.

- In this PR, I also move collapsible component from shared module to diagnostic data module so both detectors and tools page can use same collapsible component.

Here is sample code on how to adding a section into detector code,
```
public async static Task<Response> Run(DataProviders dp, OperationContext cxt, Response res){
     var sectionResponse = new Response();
     sectionResponse.AddInsight(InsightStatus.Success, "Success");
    
    //add sectionResponse into section with expanding by default
    res.AddSection(sectionResponse, "Section Title", true);
}
```
image

PR for backend change to support section rendering type
https://github.com/Azure/Azure-AppServices-Diagnostics/pull/657


AngularApp/projects/diagnostic-data/src/lib/components/sections/sections.component.ts
```diff
@@ -0,0 +1,41 @@
...
+  protected processData(data: DiagnosticData) {
+    super.processData(data);
+    this.parseData(data.table);
+  }
+
+  private parseData(table: DataTableResponseObject) {
+    const titleColumn = 0;
+    const dataColumn = 1;
+    const isExpandColumn = 2;
+    for (let i = 0; i < table.rows.length; i++) {
```
ShekharGupta1988, line 24: can you put a null check here for rows before accessing it?

ShekharGupta1988 (approved):

AngularApp/projects/diagnostic-data/src/lib/components/sections/sections.component.ts
XiaoxuHU: Yes, @ShekharGupta1988. I updated with null checking

cindylovescoding (approved):

PR #1104, opened 2021-04-06
Address asc header issue

hforeste:
- Update logging for NationalCloud (use appservicelogs when appinsights is not functioning)
- Use AppSvcUx cluster to get logging data of `locationPlacementId`
- Defer to fallback solution if subscription details are not in kusto


ApplensBackend/Controllers/DiagnosticController.cs
```diff
@@ -53,15 +58,22 @@ private class InvokeHeaders
...
             Env = env;
             DiagnosticClient = diagnosticClient;
             EmailNotificationService = emailNotificationService;
-            blackListedAscRegions = configuration.GetValue<string>("BlackListedAscRegions", string.Empty).Replace(" ", string.Empty).Split(new char[] { ',' }, StringSplitOptions.RemoveEmptyEntries);
-            diagAscHeaderValue = configuration.GetValue<string>("DiagAscHeaderValue");
+            forbiddenAscRegions = configuration.GetValue<string>("ForbiddenAscRegions", string.Empty).Replace(" ", string.Empty).Split(new char[] { ',' }, StringSplitOptions.RemoveEmptyEntries);
+            if (!forbiddenAscRegions.Any())
+            {
+                // will remove once production config is updated to use the more racially neutral term `ForbiddenAscRegions`
+                forbiddenAscRegions = configuration.GetValue<string>("BlackListedAscRegions", string.Empty).Replace(" ", string.Empty).Split(new char[] { ',' }, StringSplitOptions.RemoveEmptyEntries);
```
ShekharGupta1988, line 70: I was about to put a comment, but then saw your comment one line above :)

ApplensBackend/Services/AppSvcUxDiagnosticDataService/AppSvcUxDiagnosticDataService.cs
```diff
@@ -0,0 +1,138 @@
...
+                sd.TryGetProperty(subscriptionId, "locationPlacementId", out locationPlacementIds);
+                return locationPlacementIds;
+            }
+            catch (Exception ex)
+            {
+                throw new LocationPlacementIdException("Failed to get locationPlacementId information", ex);
+            }
+        }
+
+        private async Task Initialize()
```
cindylovescoding, line 67: We can just remove this right?

ApplensBackend/Services/AppSvcUxDiagnosticDataService/AppSvcUxDiagnosticDataService.cs
```diff
@@ -0,0 +1,138 @@
...
+        }
+
+        /// <summary>
+        /// We log subscripton data in the AppSvcUX cluster from our app service diagnostics codebase.
+        /// </summary>
+        private async Task<SubscriptionPropertiesDictionary> GetSubscriptionProperties()
+        {
+            SubscriptionPropertiesDictionary sd = null;
+            const string _appServiceDiagnosticsLocationPlacementIdQuery = @"
+ClientTelemetryNew
```
cindylovescoding, line 78: Do we want to go back to 45 days for this query? A little concerned about parsing data for 45 days records:)

cindylovescoding (approved):

ApplensBackend/Services/AppSvcUxDiagnosticDataService/AppSvcUxDiagnosticDataService.cs
hforeste: It should be fine because we sort of run this query in the background and we don't block the request to wait on this data. Then we cache the results also.

ShekharGupta1988 (approved):

PR #1112, opened 2021-04-13
Rendering table cell with markdown

XiaoxuHU:
This PR is to enable datable to support data-table to rending cells with markdown. For now, if you want  to render some logs with highlight or customized text style, I recommend to use markdown(<markdown></markdown>) instead of embedded HTML, since some styles will not show up for embedded HTML

### DIFF
--- a/AngularApp/projects/diagnostic-data/src/lib/components/data-table-v4/data-table-v4.component.ts
+++ b/AngularApp/projects/diagnostic-data/src/lib/components/data-table-v4/data-table-v4.component.ts
@@ -271,6 +271,11 @@ export class DataTableV4Component extends DataRenderBaseComponent implements Aft
     const columns = this.diagnosticData.table.columns;
     return columns.findIndex(col => col.columnName === option.name) > -1;
   }
+
+  isMarkdown(s: any) {
+    const str = `${s}`;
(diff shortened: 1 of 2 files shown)


AngularApp/projects/diagnostic-data/src/lib/components/data-table-v4/data-table-v4.component.ts
```diff
@@ -271,6 +271,11 @@ export class DataTableV4Component extends DataRenderBaseComponent implements Aft
     const columns = this.diagnosticData.table.columns;
     return columns.findIndex(col => col.columnName === option.name) > -1;
   }
+
+  isMarkdown(s: any) {
+    const str = `${s}`;
+    return str.trim().startsWith('<markdown>') && str.endsWith('</markdown>');
```
cindylovescoding, line 277: str.trim().endsWith('</markdown>')?

cindylovescoding (approved):

AngularApp/projects/diagnostic-data/src/lib/components/data-table-v4/data-table-v4.component.ts
XiaoxuHU: Updated

michimune (approved):

PR #1144, opened 2021-05-07
Add localization parameter in D&S and Applens

cindylovescoding:


ShekharGupta1988 (approved):
Looks great!!
Cindy - Can you do two more tests to make sure nothing gets broken:
a) for Products still using old Experience like ASE , Linux, Spring cloud etc
b) for Partners not passing the locale query string param from their RP (like AKS)

cindylovescoding:
> Looks great!!
> Cindy - Can you do two more tests to make sure nothing gets broken:
> a) for Products still using old Experience like ASE , Linux, Spring cloud etc
> b) for Partners not passing the locale query string param from their RP (like AKS)

Sounds good Shekhar! Will test these scenarios!

XiaoxuHU (approved):

PR #1164, opened 2021-05-19
Nmallick1/base64 partner apis

nmallick1:
Accept base64 encoded response from partners for GET detector and list detectors calls that they proxy into us. This will allow partners to publish their own swagger and not depend on ours eliminating their dependency on us.

### DIFF
--- a/AngularApp/projects/app-service-diagnostics/src/app/shared/services/arm.service.ts
+++ b/AngularApp/projects/app-service-diagnostics/src/app/shared/services/arm.service.ts
@@ -207,10 +207,74 @@ export class ArmService {
                 }));
 
             }),
-            catchError(this.handleError.bind(this))
+            map(r=> this.getDecodedDetectorResponseMessageEnvelope<T>(url, resourceUri, r)),
+            catchError(this.handleError.bind(this))            
         );
 
         return this._cache.get(url, request, invalidateCache, logData);
+    }
+
+    getDecodedDetectorResponseMessageEnvelope<T>(url:string, resourceId:string, response:ResponseMessageEnvelope<T>):ResponseMessageEnvelope<T>{
+        let isDetectorsCall = url.toLowerCase().indexOf("/detectors/") > 0 || url.toLowerCase().indexOf("/detectors?") > 0 ;
+        if(isDetectorsCall && this._genericArmConfigService && this._genericArmConfigService.isArmApiResponseBase64Encoded(resourceId)){
+            const loggingError = new Error();
+            const loggingProps = {};
+            loggingProps['Name'] = 'PartnerResponseBase64DecodingFailure';
+            if(response && response.properties && (((response.properties) as any).value)) {
+                let encodedDetectorResponse:string = ((response.properties) as any).value;
+                let decodedDetectorResponseString:string = '';
+                try {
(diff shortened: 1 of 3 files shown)


AngularApp/projects/app-service-diagnostics/src/app/shared/services/arm.service.ts
```diff
@@ -207,12 +207,76 @@ export class ArmService {
...
             }),
-            catchError(this.handleError.bind(this))
+            map(r=> this.getDecodedDetectorResponseMessageEnvelope<T>(url, resourceUri, r)),
+            catchError(this.handleError.bind(this))            
         );
 
         return this._cache.get(url, request, invalidateCache, logData);
     }
 
+    getDecodedDetectorResponseMessageEnvelope<T>(url:string, resourceId:string, response:ResponseMessageEnvelope<T>):ResponseMessageEnvelope<T>{
```
cindylovescoding, line 217: So any of the decoding process or json parse decoded string fails we will just return null right Nazeef?

cindylovescoding (approved):

ShekharGupta1988 (approved):

AngularApp/projects/app-service-diagnostics/src/app/shared/services/arm.service.ts
nmallick1: Yes @cindylovescoding, I confirmed the behavior for both GET and LIST detector calls. For LIST detectors, I handle the NULL on line#580 and the returned array of detectors will be empty. However for GET detector calls, this will end up in a JavaScript NULL Ref exception. The user will be impacted however I figured it'd the right thing to do and since we are logging the exact reason of failure, we should be good to remedy the problem too.

What do you think?

cindylovescoding: Thanks for the explanation Nazeef! In this case, for failed decrypted detector call, I guess we will show an error page for customers right? I think we should return null but not quite sure about the exception. @ShekharGupta1988  do you have any suggestions?

nmallick1: The null ref happens exception in generic-api.service.ts that is accessing the properties field of the ARM response without a NULL check. If we have to change the exception, we'll need to update that piece of code and throw a different exception.

The only reason I did not modify that piece is because tomorrow if we add a new code path that calls this method in arm.service, that will have to do the NULL check and throw the same exception to remain consistent.

The error page shown for the customer when we return a NULL for GET detectors call is 
Sorry an error has occurred

ShekharGupta1988: For error page Cindy - I think in general I feel we should have a better error page than this message - "Sorry an error occurred". How about we bring this in our next portal sync up meeting and brainstorm.

For the exception - Since this is an undesired outcome (the decryption failure), any way to detect this and alert on it is fine in my opinion. :)

nmallick1: Since we are logging a specific exception in case of a failure, we should be able to alert on it. I'll resolve this comment and merge the PR.

PR #1167, opened 2021-05-24
Diagnose and Solve Framework localization

cindylovescoding:
Diagnose and Solve Framework localization with ngx-translator:
1. Supporting 15 languages for has been requested to load Diagnose and Solve blade in last 90 days. Each of the supporting language has been added as a language json file under assets/i18n folder.
[Requested language query in last 90 days](https://ms.portal.azure.com#@72f988bf-86f1-41af-91ab-2d7cd011db47/blade/Microsoft_Azure_Monitoring_Logs/LogsBlade/resourceId/%2Fsubscriptions%2Fc1972e9d-b3c7-4de4-acb3-681773b28ced%2FresourceGroups%2FDiagnoseAndSolve%2Fproviders%2Fmicrosoft.insights%2Fcomponents%2FDiagnoseAndSolvePortal/source/LogsBlade.AnalyticsShareLinkToQuery/q/H4sIAAAAAAAAA5WQMQ7CMAxFd05hZahShFBXhjLBKSqG0FptUONUsUMBcXhCEYWFgdXvvze4jize7c9Iwos7jB0GBLEOWYwbYAum9XpTNPkM68nYpQmx9cSViqFXB6g9ibHEoLK%252BVGmOF0FqIFEoQTxLsNTqH%252FqzP5jAOO1HKx0sXyXoDbXRtAif5nz6CvPQW9FvslJrlVfF1P1DymZpCP6EtcxWOnF0zgR7Sz%252FwkUTncLx%252B%252BAP8JYbdSgEAAA%253D%253D)

customEvents
| where timestamp > ago(90d)
| where customDimensions["url"] contains "&l="
| extend url = tostring(customDimensions["url"])
| parse url with * "&l=" language 
| extend language = tostring(split(language,".")[0])
| extend language = tostring(split(language,"&")[0])
| project language
| summarize count() by language

2. With [ngx-translator](https://github.com/ngx-translate/core), we have rich API access to the library. Translation values can be accessed in both templates and typescript code files, which is not supported in Angular8 built-in i18n. 

3. No need to complete reload the whole application for locale change. AOT compilation for performance improvement.

4. Whenever a translation value is not found for the referenced translation Id, the default English value will be used.

UI Framework Localizations:
Homepage:
image
Category overview page:
image
Feedback Panel:
image


XiaoxuHU:
@cindylovescoding, if we want to localize static text in UI code in future, 
Other than import translate module and add `| translate` pipe into HTML after text, is there any other change we need to add ?

nmallick1:
@cindylovescoding , IMHO, when a translation value is not found, we should throw/log a client side exception and alert on it when we default to English.  Does the library support this?

cindylovescoding:
> @cindylovescoding, if we want to localize static text in UI code in future,
> Other than import translate module and add `| translate` pipe into HTML after text, is there any other change we need to add ?
> Will json file under assets/i18n folder update automatically or we need to manually add text into it ?

Hi Xiaoxu, I added the tool ngx-translator-extract package and updated package.json. Basically you can run "npm run extract-translations" to get keys automatically added to all the i18n/json files.

But ideally we should check and make sure translation values in each resource files are translated properly. 
This is the tool recommended to update translation files in batch: https://www.codeandweb.com/babeledit. You will get the recommended translation value for each language and you can use pre-translate feature to update all the values by one click.

But it's a good practice in general that we check and approve each translation values are changed properly.

cindylovescoding:
> @cindylovescoding , IMHO, when a translation value is not found, we should throw/log a client side exception and alert on it when we default to English. Does the library support this?

Hi Nazeef, I can add this functionablity by injecting MissingTranslationHandler module.(https://github.com/ngx-translate/core#how-to-handle-missing-translations) I think at least logging an error will help, throwing an exception may cause breaking experience though.

PR #1175, opened 2021-05-28
Enabling network checker for Consumption SKU Function Apps

sidkri:
## Summary
At present, all diagnostic tools are only supported for Function App SKUs other than consumption (SKU is not "Dynamic").  This change allows the network checker diagnostic tool to be used for consumption SKU Function Apps.

Additional changes needed to complete this change:
- [x] The App Lens detector "Diagnostic Tools" renders a list of tools based on filtering logic maintained in the detector so the logic here has been modified to include the Function App consumption SKU.

## Validation
Local validation by navigating through various WebApps and Function Apps of various SKUs

### DIFF
--- a/AngularApp/projects/app-service-diagnostics/src/app/shared/components/tools/network-checks/network-check-flows/configFailureFlow.js
+++ b/AngularApp/projects/app-service-diagnostics/src/app/shared/components/tools/network-checks/network-check-flows/configFailureFlow.js
@@ -4,6 +4,18 @@ export var configFailureFlow = {
     title: "I tried to configure VNet integration via the Azure Portal or ARM template, but it failed",
     async func(siteInfo, diagProvider, flowMgr) {
         var vnets = null, subnets = null, subscriptions = null;
+
+        siteInfo
+
+        if (siteInfo.kind.includes("functionapp") && siteInfo.sku.toLowerCase() == "dynamic") {
+            flowMgr.addView(new InfoStepView({  
+                title: "VNet integration is not supported for Consumption Plan Function Apps.",
+                infoType: 1,
+                markdown: 'For more information please review <a href="https://docs.microsoft.com/en-us/azure/app-service/web-sites-integrate-with-vnet" target="_blank">Integrate your app with an Azure virtual network</a>.'
+            }));
+            return;
+        }
+
         var getAspSitesPromise = getAspSites(diagProvider, siteInfo["serverFarmId"]);
         flowMgr.addViews(getAspSitesPromise.then(d => d.views), "Fetching App Service Plan data...");
         var aspSites = (await getAspSitesPromise).aspSites;
--- a/AngularApp/projects/app-service-diagnostics/src/app/shared/components/tools/network-checks/network-check-flows/configFailureFlow.js
+++ b/AngularApp/projects/app-service-diagnostics/src/app/shared/components/tools/network-checks/network-check-flows/configFailureFlow.js
@@ -5,8 +5,6 @@ export var configFailureFlow = {
     async func(siteInfo, diagProvider, flowMgr) {
         var vnets = null, subnets = null, subscriptions = null;
 
-        siteInfo
-
         if (siteInfo.kind.includes("functionapp") && siteInfo.sku.toLowerCase() == "dynamic") {
             flowMgr.addView(new InfoStepView({  
                 title: "VNet integration is not supported for Consumption Plan Function Apps.",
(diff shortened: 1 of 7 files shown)


AngularApp/projects/app-service-diagnostics/src/app/shared/components/tools/network-checks/network-checks.component.ts
```diff
@@ -114,31 +114,27 @@ export class NetworkCheckComponent implements OnInit, AfterViewInit {
                 flows = flows.concat(remoteFlows);
             }
             var mgr = this.stepFlowManager;
-            if (this.siteInfo.kind.includes("functionapp") && this.siteInfo.sku.toLowerCase() == "dynamic") {
-                mgr.addView(new CheckStepView({ id: "NotSupportedCheck", title: "VNet integration is not supported by Consumption Plan Function App", level: 2 }));
```
yifguoMSFT, line 118: We want to show this information when user lands to the tool from case submission flow for "VNet integration" support topic

sidkri: Got it, so should I add this right above the drop-down menu?

yifguoMSFT: For now we only redirect user to this tool for "VNet integration" topic, so we want to show "VNet integration is not supported for consumption plan" and nothing else. You can use `isSupportCenter` to decide if user comes from case submission flow or not. In the future if we redirect more topics to this tool, we can change later.

sidkri: Discussed offline and added back this check but expanded it to only match when the tool is loaded as part of the support center solution.  Changed the check step to an info step so we can provide a link to documentation around VNet integration.

AngularApp/projects/app-service-diagnostics/src/app/shared/components/tools/network-checks/network-check-flows/configFailureFlow.js
```diff
@@ -4,6 +4,18 @@ export var configFailureFlow = {
     title: "I tried to configure VNet integration via the Azure Portal or ARM template, but it failed",
     async func(siteInfo, diagProvider, flowMgr) {
         var vnets = null, subnets = null, subscriptions = null;
+
+        siteInfo
```
yifguoMSFT, line 8: remove?

yifguoMSFT (approved):

AngularApp/projects/app-service-diagnostics/src/app/shared/components/tools/network-checks/network-check-flows/configFailureFlow.js
sidkri: Good catch! Removed.  Not sure how that snuck through haha

AngularApp/projects/app-service-diagnostics/src/app/shared/components/tools/network-checks/network-check-flows/configFailureFlow.js
```diff
@@ -4,6 +4,16 @@ export var configFailureFlow = {
     title: "I tried to configure VNet integration via the Azure Portal or ARM template, but it failed",
     async func(siteInfo, diagProvider, flowMgr) {
         var vnets = null, subnets = null, subscriptions = null;
+
+        if (siteInfo.kind.includes("functionapp") && siteInfo.sku.toLowerCase() == "dynamic") {
```
cindylovescoding, line 8: Hi Sid, could we add a check to exclude single tenant logic apps? Since single tenant logic apps will have kind attribute like "functionapp, workflowapp".

Also here do we want to check siteInfo.sku for “elastic premium” plan as well?

AngularApp/projects/app-service-diagnostics/src/app/shared/components/tools/network-checks/network-checks.component.ts
```diff
@@ -114,8 +114,15 @@ export class NetworkCheckComponent implements OnInit, AfterViewInit {
...
+            if (this.isSupportCenter && 
+                this.siteInfo.kind.includes("functionapp") &&
```
cindylovescoding, line 118: Same here as above comment

cindylovescoding (approved):

AngularApp/projects/app-service-diagnostics/src/app/shared/components/tools/network-checks/network-check-flows/configFailureFlow.js
sidkri: Discussed offline.  Logic Apps are handled at a higher level and the network checker does not show up for them.  No change needed at this time.

AngularApp/projects/app-service-diagnostics/src/app/shared/components/tools/network-checks/network-checks.component.ts
sidkri: Discussed offline.  Logic Apps are handled at a higher level and the network checker does not show up for them.  No change needed at this time.

PR #1186, opened 2021-06-15
Add config files for Microsoft.Batch/batchAccounts

vincentlabonte:


yunjchoi (approved):
Approved under the condition that this is only for testing purposes. When it's time to make the detectors public to the external customers, we need to make additional changes in the config to onboard to the newest UI.

XiaoxuHU (approved):

PR #1192, opened 2021-06-23
Add health ping for D&S and Applens backend

XiaoxuHU:


nmallick1:
@XiaoxuHU , how about adding a call to google generate 204 page just like runtimehost to determine the instance's ability to communicate with the internet?

XiaoxuHU:
> @XiaoxuHU , do you think adding a call to google generate 204 page just like runtimehost to determine the instance's ability to communicate with the internet will be helpful?

Yes, @Nazeef. I can add google 204 check as well

cindylovescoding (approved):

XiaoxuHU:
> > @XiaoxuHU , do you think adding a call to google generate 204 page just like runtimehost to determine the instance's ability to communicate with the internet will be helpful?
> 
> Yes, @nmallick1 . I can add google 204 check as well

Hi @nmallick1, I add network checking for Applens backend.
Can you review the change ? 
If it is okay, I will add same code for D&S UI backend as well

ApplensBackend/Services/HealthCheckService/HealthCheckService.cs
```diff
@@ -0,0 +1,100 @@
...
+        private static HttpClient _httpClient;
+        IConfiguration _configuration;
+        bool IsOutboundConnectivityCheckEnabled = false;
+        private const string OUTBOUND_CONNECTIVITY_CACHE_KEY = "outboundconnectivitycheck";
+        private IMemoryCache cache;
+
+        public HealthCheckService(IConfiguration Configuration, IMemoryCache cache)
+        {
+            _configuration = Configuration;
+            IsOutboundConnectivityCheckEnabled = Convert.ToBoolean(_configuration["HealthCheckSettings:IsOutboundConnectivityCheckEnabled"]);
```
rekhaswaminathan, line 22: `HealthCheckSettings:IsOutboundConnectivityCheckEnabled` should be part of `appsettings.json` correct? I don't see the change for `appsettings.json` in the PR

ApplensBackend/Services/HealthCheckService/HealthCheckService.cs
```diff
@@ -0,0 +1,100 @@
...
+            if (IsOutboundConnectivityCheckEnabled)
+            {
+                OutboundConnectivityCheckUrl = _configuration["HealthCheckSettings:OutboundConnectivityCheckUrl"];
```
rekhaswaminathan, line 25: Same for the config `HealthCheckSettings:OutboundConnectivityCheckUrl`

ApplensBackend/Services/HealthCheckService/HealthCheckService.cs
XiaoxuHU: Yes, correct. 
So I can add these config into appsetting.json instead of update appsetting in portal ?

rekhaswaminathan: You can keep it part of `appsettings.json` as well. Because if the settings (at least the keys) is not checked in, if someone checks out the main branch and runs the code, it will result in `null` values.

ApplensBackend/Services/HealthCheckService/HealthCheckService.cs
XiaoxuHU: update appsettings.json

ApplensBackend/Services/HealthCheckService/HealthCheckService.cs
XiaoxuHU: update appsettings.json

rekhaswaminathan (approved):

nmallick1 (approved):

PR #1193, opened 2021-06-23
Adding AppLens Insights on CRI Correction page

SharmaAjay19:
We have already started posting solutions on ICM incidents.

This work is related to showing solutions to customers even before that when the CRI validation has failed and they visit **AppLens CRI correction page** to update the right information.
ErrorPage

Once they provide the correct information, we show them any insights that show potential issues detected with their app.
ShowInsights

It is possible customer finds this information relevant and never reactivates the CRI, thus creating deflection.


nmallick1 (approved):

khaled-zayed:
♥

PR #1208, opened 2021-07-02
Fix keystone insight image in Applens

cindylovescoding:
Fix broken image in keystone insight Applens view:
image


khaled-zayed (approved):

XiaoxuHU (approved):

AngularApp/projects/diagnostic-data/src/lib/components/solution-view-container/solution-view-container.component.ts
```diff
@@ -1,5 +1,7 @@
 
+import { Inject } from '@angular/core';
 import { Component, Input, OnInit } from '@angular/core';
+import { DiagnosticDataConfig, DIAGNOSTIC_DATA_CONFIG } from 'diagnostic-data';
```
XiaoxuHU, line 4: This import should be the root cause of build failure

cindylovescoding: Yes it has been fixed already

ShekharGupta1988 (approved):

PR #1211, opened 2021-07-09
Enable diagnostic tools only for linux web app and linux paid dedicated function app

cindylovescoding:
@puneetg1983  Putting this PR for autoheal disablement as we discussed the other day. 
In this PR, "Diagnostic Tools" category card will only show for the following type of resources in D&S homepage:
1. Linux web app
2. Linux paid dedicated function app.

Please help review/test it accordingly and see if I am missing anything here.

### DIFF
--- a/AngularApp/projects/app-service-diagnostics/src/app/resources/web-sites/services/sites-category.service.ts
+++ b/AngularApp/projects/app-service-diagnostics/src/app/resources/web-sites/services/sites-category.service.ts
@@ -237,7 +237,7 @@ export class SitesCategoryService extends CategoryService {
           createFlowForCategory: true,
           chatEnabled: false
         }
-      },
+      }
   ];
 
   constructor(private _resourceService: WebSitesService, private _websiteFilter: WebSiteFilter, private _armService: ArmService) {
@@ -266,16 +266,19 @@ export class SitesCategoryService extends CategoryService {
       );
     }
 
-    this._sitesCategories.push(this._getDiagnosticToolsCategory(this._resourceService.resourceIdForRouting));
+ //   this._sitesCategories.push(this._getDiagnosticToolsCategory(this._resourceService.resourceIdForRouting));
+    this._getDiagnosticToolsCategory(this._resourceService.resourceIdForRouting).forEach((diagnosticCategory) => {
+        this._sitesCategories.push(diagnosticCategory);
+    });
 
     this._addCategories(
       this._websiteFilter.transform(this._sitesCategories)
     );
   }
 
-  private _getDiagnosticToolsCategory(siteId: string): SiteFilteredItem<Category> {
-    return <SiteFilteredItem<Category>>{
-      appType: AppType.WebApp | AppType.FunctionApp,
+  private _getDiagnosticToolsCategory(siteId: string): SiteFilteredItem<Category>[] {
+    return <SiteFilteredItem<Category>[]> [{
+      appType: AppType.WebApp,
       platform: OperatingSystem.any,
       stack: '',
       sku: Sku.All,
@@ -290,7 +293,24 @@ export class SitesCategoryService extends CategoryService {
         createFlowForCategory: false,
         overridePath: `resource${siteId}/diagnosticTools`
       }
-    };
+    },
+    {
+        appType: AppType.FunctionApp,
+        platform: OperatingSystem.linux,
+        stack: '',
+        sku: Sku.PaidDedicated,
+        hostingEnvironmentKind: HostingEnvironmentKind.All,
+        item: {
+          id: 'DiagnosticTools',
+          name: 'Diagnostic Tools',
+          overviewDetectorId:'DiagnosticTools',
+          description: 'Run proactive tools to automatically mitigate the app.',
+          keywords: ['Auto-Heal'],
+          color: 'rgb(170, 192, 208)',
+          createFlowForCategory: false,
+          overridePath: `resource${siteId}/diagnosticTools`
+        }
+      }];
   }
 
 }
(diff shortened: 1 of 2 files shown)


AngularApp/projects/app-service-diagnostics/src/app/resources/web-sites/services/sites-category.service.ts
```diff
@@ -290,7 +293,24 @@ export class SitesCategoryService extends CategoryService {
         createFlowForCategory: false,
         overridePath: `resource${siteId}/diagnosticTools`
       }
-    };
+    },
+    {
+        appType: AppType.FunctionApp,
+        platform: OperatingSystem.linux,
```
sidkri, line 299: This should be OperatingSystem.any so windows dedicated also has diagnostic tools.

sidkri:
Based on the logic in this PR it looks like the following Product types, SKUs and OS will have diagnostic tools visible:
1. WebApp, All SKUs, All OS
2. FunctionApp, Dedicated SKU, Linux

I believe we want:
1. WebApp, All SKUs, All OS
2. FunctionApp, Dedicated SKU, All OS

puneet-gupta:
> Based on the logic in this PR it looks like the following Product types, SKUs and OS will have diagnostic tools visible:
> 
> 1. WebApp, All SKUs, All OS
> 2. FunctionApp, Dedicated SKU, Linux
> 
> I believe we want:
> 
> 1. WebApp, All SKUs, All OS
> 2. FunctionApp, Dedicated SKU, All OS

@cindylovescoding can correct me if I am wrong but I believe it is done this way because this part of the UI is only shown for resource types using the OLD D&S UI. For **Function Apps on Dedicated SKU on Windows**, we are already showing them the new UI and that gets the list of tools and categories from a different mechanism.

I am fine with this PR. 

One small note - AutoHealing works even for FREE and SHARED Web Apps (both Windows and Linux). Currently for "Windows Free Function Apps", we are not showing AutoHealing tool and with this PR, the same will happen for "Linux Free Functions". We can decide if we need to change that in future but to me it makes sense to keep both types of Free Function Apps in sync.

puneet-gupta (approved):
Tested with different combinations and the results look good.

AngularApp/projects/app-service-diagnostics/src/app/shared/models/server-farm.ts
```diff
@@ -29,10 +29,13 @@ export enum Sku {
...
     PremiumV2 = 1 << 6,
     Isolated = 1 << 7,
-    PremiumContainer = 1 << 8,
-    Paid = 504,         // 111111000
-    NotDynamic = 507,   // 111111011
-    All = 511           // 111111111
+    PremiumContainer = 1 << 8, // 100000000
+    ElasticPremium = 1 << 9, // 1000000000
+    Paid = 1016,         // 1111111000
+    Dedicated = 507,    // 111111011 Not Dynamic or ElasticPremium
```
puneet-gupta, line 35: Since this enum value includes Free and Shared, it is incorrect to call this Dedicated. Best is to rename this to NotDynamicAndElasticPremium as that matches exactly what this value represents.

sidkri:
> > Based on the logic in this PR it looks like the following Product types, SKUs and OS will have diagnostic tools visible:
> > 
> > 1. WebApp, All SKUs, All OS
> > 2. FunctionApp, Dedicated SKU, Linux
> > 
> > I believe we want:
> > 
> > 1. WebApp, All SKUs, All OS
> > 2. FunctionApp, Dedicated SKU, All OS
> 
> @cindylovescoding can correct me if I am wrong but I believe it is done this way because this part of the UI is only shown for resource types using the OLD D&S UI. For **Function Apps on Dedicated SKU on Windows**, we are already showing them the new UI and that gets the list of tools and categories from a different mechanism.
> 
> I am fine with this PR.
> 
> One small note - AutoHealing works even for FREE and SHARED Web Apps (both Windows and Linux). Currently for "Windows Free Function Apps", we are not showing AutoHealing tool and with this PR, the same will happen for "Linux Free Functions". We can decide if we need to change that in future but to me it makes sense to keep both types of Free Function Apps in sync.

Thanks for clarifying, makes sense.  We should add a comment in the code to explain this point so its clear to future code authors as it will look like Windows is not supported otherwise.

AngularApp/projects/app-service-diagnostics/src/app/resources/web-sites/services/sites-category.service.ts
sidkri: After discussion with Puneet I learned that the logic is different for Windows as it uses a newer UI.  Suggest leaving a comment in this area explaining that this logic only applies to Linux and mention the location where windows logic lives (to help future code readers/authors)

sidkri (approved):

AngularApp/projects/app-service-diagnostics/src/app/resources/web-sites/services/sites-category.service.ts
cindylovescoding: Thanks Sid! Will add the comment here for clarification

AngularApp/projects/app-service-diagnostics/src/app/shared/models/server-farm.ts
cindylovescoding: Make sense Puneet! Will rename this!

AngularApp/projects/app-service-diagnostics/src/app/resources/web-sites/services/sites-category.service.ts
cindylovescoding: While testing this again, I think @sidkri you are right about Function App running on Windows. I am gonna push a fix for it/

AngularApp/projects/app-service-diagnostics/src/app/shared/models/server-farm.ts
sidkri: how about adding two enums: "Dedicated", "FreeAndShared"?  Not a fan of enums that say "NotSomething" as it can be interpreted to include SKUs that are added in the future and is generally imprecise.

PR #1213, opened 2021-07-09
Fix bug 1922: commonRec is not defined/Cannot read property 'length' of undefined

yifguoMSFT:
fixing [Fix bug 1922: commonRec is not defined/Cannot read property 'length' of undefined](https://dev.azure.com/app-service-diagnostics-portal/app-service-diagnostics-portal/_workitems/edit/1922)

### DIFF
--- a/AngularApp/projects/app-service-diagnostics/src/app/shared/components/tools/network-checks/network-check-flows/functionsFlow.js
+++ b/AngularApp/projects/app-service-diagnostics/src/app/shared/components/tools/network-checks/network-check-flows/functionsFlow.js
@@ -212,7 +212,7 @@ async function networkCheckConnectionString(propertyName, connectionString, dnsS
         var hostPort = extractHostPortFromConnectionString(connectionString);
 
         if (hostPort.HostName != undefined && hostPort.Port != undefined) {
-            var connectivityCheckResult = await runConnectivityCheckAsync(hostPort.HostName, hostPort.Port, dnsServers, diagProvider, hostPort.HostName.length, isVnetIntegrated, failureDetailsMarkdown);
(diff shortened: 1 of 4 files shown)


AngularApp/projects/app-service-diagnostics/src/app/shared/components/tools/network-checks/network-check-flows/functionsFlow.js
```diff
@@ -334,7 +334,7 @@ async function checkNetworkConfigAndConnectivityAsync(siteInfo, diagProvider, fl
     return views;
 }
 
-async function runConnectivityCheckAsync(hostname, port, dnsServers, diagProvider, lengthLimit, isVnetIntegrated, failureDetailsMarkdown = undefined) {
+async function runConnectivityCheckAsync(hostname, port, dnsServers, diagProvider, lengthLimit = 50, isVnetIntegrated, failureDetailsMarkdown = undefined) {
```
sidkri, line 337: Could you please add a comment above this function to explain why we have a default limit of 50?  It also looks like the two callers in this file pass undefined so why not just remove this variable and have a static value of 50 for all calls?

sidkri (approved):

cindylovescoding (approved):

AngularApp/projects/app-service-diagnostics/src/app/shared/components/tools/network-checks/network-check-flows/functionsFlow.js
yifguoMSFT: It's just a magic number that will show endpoint longer than lengthLimit with ellipsis. It's not limiting the length of input

PR #1216, opened 2021-07-14
Showing 5 results from DeepSearch for external users by default

vijaysaayi:
Showing 5 results as default for both internal and external users


SharmaAjay19 (approved):
I see we are expanding to 5 results by default in public portal. Let's keep 5 for AppLens as well.

PR #1219, opened 2021-07-16
Add safety check for snapshot parameters in activatedroute

cindylovescoding:


AngularApp/projects/diagnostic-data/src/lib/components/detector-list-analysis/detector-list-analysis.component.ts
```diff
@@ -782,7 +782,7 @@ export class DetectorListAnalysisComponent extends DataRenderBaseComponent imple
                 if (this.analysisId === "searchResultsAnalysis" && this.searchTerm && this.searchTerm.length > 0) {
                     //If in homepage then open second blade for Diagnostic Tool and second blade will continue to open third blade for
                     if (this.withinGenie) {
-                        const isHomepage = !this._activatedRoute.root.firstChild.firstChild.firstChild.firstChild.snapshot.params["category"];
+                        const isHomepage = !this._activatedRoute?.root?.firstChild?.firstChild?.firstChild?.snapshot?.params["category"];
```
XiaoxuHU, line 785: See build error. I think in our current TS version it is still not supporting for `?.` syntax.

cindylovescoding: Ya I noticed that. I thought this should be supported  since typescript 2?

XiaoxuHU: Our typescript version is 3.4.5. Chain operator supporting starts 3.7.

I remember Jeff also bring this up sometime back. The blocker for updating Typescript is angular-react-fabric library will not working in higher Angular version

cindylovescoding: Yes that is what I thought too. It looks like I was confused by this: https://www.jianshu.com/p/1ccdc6a4d0ce

XiaoxuHU (approved):

PR #1225, opened 2021-07-24
Refactor connection failure flow of Network Troubleshooter 

yifguoMSFT:
1. Refactor the connection failure flow include
    - improve the checks
    - reorganize the flow's code to separate check logic and wordings
2. improve the check-step UI component


XiaoxuHU (approved):
Angular part looks good

AngularApp/projects/app-service-diagnostics/src/app/shared/components/tools/network-checks/network-check-flows/connectionFailureFlow.js
```diff
@@ -1,58 +1,49 @@
...
     var nameResolvePromise = (async function checkNameResolve() {
         var ip = null;
         var subChecks = [];
         if (diagProvider.isIp(hostname)) {
             ip = hostname;
         } else {
             for (var i = 0; i < dnsServers.length; ++i) {
+                var dns = `DNS server ${dnsServers[i]}`;
+                if(dnsServers[i] == ""){
+                    dnsServers[i] = "168.63.129.16";
```
sidkri, line 37: Could you leave a comment here explaining why we are setting the dns servers if not defined?  Would we not want the check to fail?

sidkri (approved):

AngularApp/projects/app-service-diagnostics/src/app/shared/components/tools/network-checks/network-check-flows/connectionFailureFlow.js
yifguoMSFT: This is a temporary fix as I've noticed that the logic of nameresolver.exe doesn't follow the same logic when worker resolves hostnames for user app. I am going to refactor the DNS check part later and use a DaaS controller instead of nameresolver.exe

PR #1226, opened 2021-07-26
Solution Orchestrator

SharmaAjay19:
Just for comparison, this is what the dynamic analysis looks like which currently renders under the support topic:
image

**So what does Solution orchestrator look like?**

Landing page:
image

When detector **solutions are found and have high score**:
image

<u>**The major difference that we see is that we are directly recommending solutions and are reducing the amount of information on the page thus preventing information explosion.**</u>

Customer can toggle between different solutions recommended to see details:
image

Customer can provide specific feedback on why a particular solution is not helpful:
image

Customer can click on View Supporting data to go to the detector and see details for the solutions:
image

image

Use the **Return back to main screen** link to go back to main screen:
image

There are three sections for documents, critical insights and successful checks:
image

Documents:
image

Critical Insights and Successful checks:
image

Clicking on "View Solutions" button displays all solutions provided by that insight:
image

The "View diagnostic data" links take to the detector and display details:
image

image

When detector solutions are not found, Azure Guides are displayed by default:
image

Feedback button behaves exactly as the feedback button everywhere else:
image


AngularApp/projects/app-service-diagnostics/src/app/shared-v2/services/support-topic.service.ts
```diff
@@ -83,7 +83,13 @@ export class SupportTopicService {
                         }
                     }
                     else {
-                        detectorPath = `/analysis/searchResultsAnalysis/search`;
+                        if (this.supportTopicId == "32629421") {
```
SharmaAjay19, line 86: This logic will be changed to accommodate more support topics as Solution Orchestrator is finalized after bug bash and gets ready for production.

XiaoxuHU (approved):

PR #1260, opened 2021-09-27
Enable A/B testing in Portal

XiaoxuHU:
This PR is for 
1) Enable reading slot info from Ibiza in portal.service
2) Only allow logging flow into App Insight and Kusto for Prod and Preview slot.
3) Adding banner for open Prod/Preview in new browser tab, it will hide if banner for switch between V2/V3 is showing up(for now it is only enabled for Linux App)
4) Reduce min-height for detector-container component as the ask from AKS team to not leave a big blank place when only show check list for analysis

banner shown in Prod slot,
image

banner shown in Preview slot,
image


AngularApp/projects/app-service-diagnostics/src/app/home/components/home/home.component.html
```diff
@@ -1,10 +1,15 @@
 <ng-container *ngIf="initializedPortalVersion === 'v3'&&showSwitchBanner">
-  <div *ngIf="useLegacy" class="switch2new-context-text">
+  <div *ngIf="useLegacy" class="switch2new-context-text home-banner">
```
cindylovescoding, line 2: is this "useLegacy" reused to control Production VS preview detector view? This is used to control the new and old D&S portal experience. Do we want to use a different boolean to control it?

XiaoxuHU: No, useLegacy will be only control between old and new D&S portal experience
There are two new boolean,
"enableABTesing" is to control show/hide Prod/Preview banner,
"isPreview" is to check Prod Vs Preview

AngularApp/projects/app-service-diagnostics/src/app/home/components/home/home.component.ts
```diff
@@ -377,6 +382,26 @@ export class HomeComponent implements OnInit, AfterViewInit {
...
+        } else {
+            this.abTestingBannerText = "You are currently on the old version of the Diagnose and Solve Problems. To explore the latest UI improvements in the new version";
+        }
+
+    }
+
+    openSlotInNewTab(event: Event) {
+        event.stopPropagation();
+        const url = this.abTestingService.generateSlotLink();
+        this._telemetryService.logEvent(TelemetryEventNames.OpenSlotInNewTab,{
```
cindylovescoding, line 402: Thank you for adding the logs and initial source slot here @XiaoxuHU ! This will be very useful for us to calculate the user engagement number and conversion rate!

AngularApp/projects/diagnostic-data/src/lib/services/telemetry/telemetry.service.ts
```diff
@@ -166,17 +168,18 @@ export class TelemetryService {
...
-        
+
         return productName;
     }
 
-    private addCommonLoggingProperties(properties: { [name: string]: string }):void {
+    private addCommonLoggingProperties(properties: { [name: string]: string }): void {
         properties = properties || {};
         properties["portalVersion"] = this.isLegacy ? 'v2' : 'v3';
+        properties["slot"] = this.slot;
```
cindylovescoding, line 182: Love that this slot info is populated to logging as common properties everywhere!

cindylovescoding (approved):

AngularApp/projects/diagnostic-data/src/lib/services/telemetry/telemetry.service.ts
XiaoxuHU: For local/test/staging/previewStaging, we still get logs. Just the slot name will be different

cindylovescoding: Oh got it. Do you think it makes more sense to remove logging for local and testing env?

ShekharGupta1988 (approved):

PR #1298, opened 2021-10-23
Service fabric genie v3

yunjchoi:
Migrating Service Fabric to Genie V3


yunjchoi:
@XiaoxuHU It seems like the icon becomes transparent when I upload it on Github. Any idea what might be going on here? 
Actual:
image
Expected: 
image

PR #1353, opened 2021-11-24
Correct time picker button text

cindylovescoding:


AngularApp/projects/app-service-diagnostics/src/app/home/resolvers/resource.resolver.ts
```diff
@@ -16,6 +16,12 @@ export class ResourceResolver implements Resolve<Observable<{} | ArmResource>> {
         if (!!startTime && !!endTime)
         {
             this._detectorControlService.setCustomStartEnd(startTime, endTime);
+            this._detectorControlService.updateTimePickerInfo({
```
XiaoxuHU, line 19: Just curious, why need update time picker info in resource.resoulver ?

XiaoxuHU (approved):

AngularApp/projects/app-service-diagnostics/src/app/home/resolvers/resource.resolver.ts
cindylovescoding: I think wherever we update the start time and end time from router, the label button should always show the same time range to be consistent

PR #1356, opened 2021-11-30
Fix theming for autoheal

cindylovescoding:
Change input background color with theming:
image


puneetg1983 (approved):

puneetg1983:
qq - We have other components for Autoheal that take input like Private bytes rule, Request Count rule and Request Duration rule. Will we be replicating these changes to those components as well ?

ShekharGupta1988 (approved):

cindylovescoding:
> 

Yes Puneet. The other components don't have input components and they are working good.

PR #1407, opened 2022-01-25
Fix loading issue on safari

cindylovescoding:
Safari doesn't support regex lookbehind, which is breaking Diagnose and Solve for safari users. This is the PR to temporary fix the loading experience. 
https://stackoverflow.com/questions/51568821/works-in-chrome-but-breaks-in-safari-invalid-regular-expression-invalid-group

Still need help from @yifguoMSFT and @puneetg1983  to find the alternative regex pattern to replace the negative lookbehind here:
const regex = new RegExp("(?<!\!)\[(.*?)]\((.*?)( +\"(.*?)\")?\)", "g");


yifguoMSFT (approved):
Thanks Cindy for the hotfix

ShekharGupta1988 (approved):

XiaoxuHU (approved):

PR #1432, opened 2022-02-11
Add new resource provider name for container apps

michimune:
This PR adds the new resource provider name (Microsoft.App) for container apps to the diag portal experiences.


cindylovescoding (approved):
Hey Michi, so we still need to support both Microsoft.App/ContainerApps and Microsoft.Web/containerApps for a short time right? Once it is completely switched over, we can safely delete Microsoft.Web/containerApps?

PR #1435, opened 2022-02-14
Add breadcrumb and fixing bugs

XiaoxuHU:
1. Adding breadcrumb for navigation from parent detector to child

image

2. Adding failed loading section for child detectors loading not successful
image

3. Bug Fixing
    3.1 Read startTime and endTime from url
    3.2 Focus will reset to time-picker when detector loading is finished
    3.3 Three loading dots in parent-child detector will disappeared after loading is done
    3.4 Metric below highchart will only show data without prefix points
    3.5 Revert to use stack chart as default

### DIFF
--- a/AngularApp/projects/applens/src/app/modules/dashboard/tabs/tab-common/tab-common.component.ts
+++ b/AngularApp/projects/applens/src/app/modules/dashboard/tabs/tab-common/tab-common.component.ts
@@ -41,22 +41,26 @@ export class TabCommonComponent implements OnInit {
     switch (key) {
       case TabKey.Data:
         this._router.navigate(["./"], {
-          relativeTo: this._activatedRoute
+          relativeTo: this._activatedRoute,
+          queryParamsHandling:"preserve"
         });
         break;
       case TabKey.Develop:
         this._router.navigate(["edit"], {
-          relativeTo: this._activatedRoute
+          relativeTo: this._activatedRoute,
+          queryParamsHandling:"preserve"
         });
         break;
       case TabKey.DataSources:
         this._router.navigate(["datasource"], {
-          relativeTo: this._activatedRoute
+          relativeTo: this._activatedRoute,
+          queryParamsHandling:"preserve"
         });
         break;
       case TabKey.CommitHistory:
         this._router.navigate(["changelist"], {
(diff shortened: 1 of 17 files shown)


AngularApp/projects/diagnostic-data/src/lib/components/detector-list/detector-list.component.html
```diff
@@ -110,6 +114,39 @@
     </collapsible-list-item>
   </collapsible-list-fabric>
 </div>
+<br />
+
+<div *ngIf="failedLoadingViewModels.length > 0">
+  <collapsible-list-fabric [collapsed]="false" [title]="'Failed Loading Checks (' + failedLoadingViewModels.length + ')'">
```
ShekharGupta1988, line 120: nitpick : just get a review of this tittle from Yun Jung and Hanna once and see if they have any feedback.

ShekharGupta1988 (approved):

AngularApp/projects/applens/src/app/modules/dashboard/tabs/tab-common/tab-common.component.ts
```diff
@@ -41,22 +41,26 @@ export class TabCommonComponent implements OnInit {
     switch (key) {
       case TabKey.Data:
         this._router.navigate(["./"], {
-          relativeTo: this._activatedRoute
+          relativeTo: this._activatedRoute,
```
cindylovescoding, line 44: Why do we need to add this?

cindylovescoding (approved):

AngularApp/projects/diagnostic-data/src/lib/components/detector-list/detector-list.component.html
XiaoxuHU: Yes, Shekhar !  I checked Hanna about the wording

AngularApp/projects/applens/src/app/modules/dashboard/tabs/tab-common/tab-common.component.ts
XiaoxuHU: Bug reported by Nazeef, if not add queryParmasHandling, some query params will missing when switching between the tabs

PR #1437, opened 2022-02-15
Fixing for Null Injection and metric for high-chart

XiaoxuHU:


AngularApp/projects/diagnostic-data/src/lib/components/detector-list/detector-list.component.ts
```diff
@@ -64,7 +64,7 @@ export class DetectorListComponent extends DataRenderBaseComponent {
   solutionTitle: string = "";
   loading = LoadingStatus.Loading;
 
-  constructor(private _diagnosticService: DiagnosticService, protected telemetryService: TelemetryService, private _detectorControl: DetectorControlService, private parseResourceService: ParseResourceService, @Inject(DIAGNOSTIC_DATA_CONFIG) private config: DiagnosticDataConfig, private _router: Router, private _activatedRoute: ActivatedRoute, private _portalActionService: PortalActionGenericService, private _featureNavigationService: FeatureNavigationService, private _breadcrumbService: GenericBreadcrumbService) {
+  constructor(private _diagnosticService: DiagnosticService, protected telemetryService: TelemetryService, private _detectorControl: DetectorControlService, private parseResourceService: ParseResourceService, @Inject(DIAGNOSTIC_DATA_CONFIG) private config: DiagnosticDataConfig, private _router: Router, private _activatedRoute: ActivatedRoute, private _portalActionService: PortalActionGenericService, private _featureNavigationService: FeatureNavigationService, @Optional() private _breadcrumbService?: GenericBreadcrumbService) {
     super(telemetryService);
```
cindylovescoding, line 68: What does optional do?

cindylovescoding (approved):

AngularApp/projects/diagnostic-data/src/lib/components/detector-list/detector-list.component.ts
XiaoxuHU: It will not throw an exception when there is no inject service found.
In this scenario, for external project, it will return as null instead of throwing exception since there is no implementation for breadcrumb service in portal.

PR #1514, opened 2022-04-04
Adding support for calling Linux DiagServer to the UI project

puneetg1983:
With this PR
- The diagnostic tools will invoke the API for the Linux Diag Server when stamp is ANT98 or above and when DiagServer is enabled.
- Also cleaned up existing code which was distinguishing between DAAS V1 and DAAS V2
- Removed some of the dead diagnostic tool related components.


AngularApp/projects/app-service-diagnostics/src/app/home/home.module.ts
```diff
@@ -220,35 +220,33 @@ export const HomeRoutes = RouterModule.forChild([
                             cacheComponent: true
                         }
                     },
-                    // Memory Dump
+                    // Profiler Tool for Linux (To be shown when All Instances of a Linux App are running on ANT 98)
                     {
-                        path: 'tools/memorydump',
-                        component: MemoryDumpToolComponent,
+                        path: 'tools/profilerlinuxant98',
```
ShekharGupta1988, line 225: nitpick - Changing the url might break deeplinks to this tool from detectors. I am not sure if this tool is deeplinked for Linux apps but just bringing this up in case we need to fix those detectors as well.

ShekharGupta1988: another small suggestion - instead of ant98 in the url, maybe we can add version in the url (like v2)

ShekharGupta1988 (approved):

AngularApp/projects/app-service-diagnostics/src/app/home/home.module.ts
puneetg1983: > nitpick - Changing the url might break deeplinks to this tool from detectors. I am not sure if this tool is deeplinked for Linux apps but just bringing this up in case we need to fix those detectors as well.

Good point, we need to be careful about deeplinking. Right now the tools are not deep linked anywhere but we will keep this in mind if we decide to go this route.

puneetg1983: corrected the names.

PR #1536, opened 2022-04-14
Enabling searching static web app by app name or default host name

XiaoxuHU:
This PR is for enabling searching Static Web App by app name or default host name via Observer

Now in landing page dropdown, we have a new option "Static Web App"
image

If multiple app with same name, it has the same experience as Web App and Container App,
image

In home page, it shows info fetched from Observer
image

### DIFF
--- a/AngularApp/projects/applens/src/app/modules/dashboard/dashboard-container/dashboard-container.component.ts
+++ b/AngularApp/projects/applens/src/app/modules/dashboard/dashboard-container/dashboard-container.component.ts
@@ -1,5 +1,4 @@
 import { Component, OnInit } from '@angular/core';
-import { ActivatedRoute } from '@angular/router';
 import { Observable, Subscription } from 'rxjs';
 import { ApplensGlobal } from '../../../applens-global';
 import { DiagnosticApiService } from '../../../shared/services/diagnostic-api.service';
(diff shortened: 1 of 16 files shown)


XiaoxuHU:
I will merge this PR after Jeff's change for Observer deployed into production

SharmaAjay19 (commented):
Looks alright, assuming that the Runtimehost supports the Microsoft.Web/staticsites resource type. If not, then that change will be needed first.
Edit: It's supported by ARM resource type I see. Looks good XiaoXu!

XiaoxuHU:
> Looks alright, assuming that the Runtimehost supports the Microsoft.Web/staticsites resource type. If not, then that change will be needed first.

SharmaAjay19 (approved):

AngularApp/projects/applens/src/app/modules/dashboard/dashboard-container/dashboard-container.component.ts
```diff
@@ -1,4 +1,5 @@
 import { Component, OnInit } from '@angular/core';
+import { ActivatedRoute } from '@angular/router';
```
SharmaAjay19, line 2: Needed?

XiaoxuHU: Removed

cindylovescoding (approved):

PR #1542, opened 2022-04-16
Fix chart styles for dark theming

cindylovescoding:
image


AngularApp/projects/applens/src/app/shared/components/applens-header/applens-header.component.ts
```diff
@@ -109,6 +109,7 @@ export class ApplensHeaderComponent implements OnInit {
     this._themeService.setActiveTheme(themeStr);
     this._userSettingService.isWaterfallViewSub.next(!this.smartViewChecked);
     this._userSettingService.updateUserPanelSetting(updatedSettings);
+    window.location.reload();
```
XiaoxuHU, line 112: So only after the page get reloaded, highchart can update the color, correct ?

XiaoxuHU (approved):

PR #1553, opened 2022-04-21
save on create

DarrelDonald:


AngularApp/projects/applens/src/app/modules/dashboard/onboarding-flow/onboarding-flow.component.ts
```diff
@@ -1512,13 +1512,14 @@ export class OnboardingFlowComponent implements OnInit {
...
-      this.saveFailed = true;
-      this.saveFailMessage = "A detector with this ID already exists. Please enter a new ID"
-      postSave();
-    }
-    */
+    this.idInSystem(this.saveTempId).subscribe(inSystem => {
+      if (this.mode == DevelopMode.Create && inSystem){
+        this.saveFailed = true;
+        this.saveFailMessage = "A detector with this ID already exists. Please enter a new ID"
+        this.postSave();
```
rekhaswaminathan, line 1519: Why do we need to call `postSave` if we do not want to allow Save?

DarrelDonald: this is what resets the save button after it gets disabled while executing the save attempt

XiaoxuHU (approved):

PR #1557, opened 2022-04-26
Update source location of detector and gist templates

hforeste:
- Copy over Detector and Gist templates from our Diagnostics project to here in our UI project as the new place for it to live
- Instead of reading templates using Github API, read from local disk as templates will be deployed along the UI project.
    - When we move this project to internal repo. No API changes need to be made. Everything will work as is.
- Update Angular project to use the template API for detectors and gist templates.

### DIFF
--- a/AngularApp/projects/applens/src/app/shared/services/detectorgist-template-api.service.ts
+++ b/AngularApp/projects/applens/src/app/shared/services/detectorgist-template-api.service.ts
@@ -1,0 +1,16 @@
+import { Injectable } from '@angular/core';
+import { DiagnosticApiService } from './diagnostic-api.service';
+import { Observable, of } from 'rxjs';
+import { Commit } from '../models/commit';
+import { Dependency } from '../models/package';
+import { map } from 'rxjs/operators';
+
+@Injectable()
+export class DetectorGistApiService {
+
+  constructor(private _diagnosticApiService: DiagnosticApiService) { }
+
+  public getTemplate(name: string): Observable<string> {
+    return this._diagnosticApiService.get<string>(`api/templates/${name}`, true);
+  }
+}
--- a/ApplensBackend/Services/DetectorGistTemplateService/TemplateService.cs
+++ b/ApplensBackend/Services/DetectorGistTemplateService/TemplateService.cs
@@ -1,0 +1,80 @@
+﻿// <copyright file="TemplateService.cs" company="Microsoft Corporation">
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+// </copyright>
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using AppLensV3.Helpers;
+using Microsoft.Extensions.Caching.Memory;
+
+namespace AppLensV3.Services
+{
+
+    public class TemplateService : IDetectorGistTemplateService
+    {
+        private readonly IMemoryCache memoryCache;
+        private Task templateTask;
+
+        public TemplateService(IMemoryCache detectorGistCache)
+        {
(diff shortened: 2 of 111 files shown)


AngularApp/projects/applens/src/app/shared/services/detectorgist-template-api.service.ts
```diff
@@ -0,0 +1,16 @@
...
+import { Commit } from '../models/commit';
+import { Dependency } from '../models/package';
+import { map } from 'rxjs/operators';
+
+@Injectable()
+export class DetectorGistApiService {
+
+  constructor(private _diagnosticApiService: DiagnosticApiService) { }
+
+  public getTemplate(name: string): Observable<string> {
```
XiaoxuHU, line 13: Are we going to add more APIs for github service in future ?
Just because we have a lot services so far

XiaoxuHU (approved):

AngularApp/projects/applens/src/app/shared/services/detectorgist-template-api.service.ts
XiaoxuHU: > You mean for the detectorgist-template-api service? We have the potential to add more APIs into this service in the future such as CRUD actions for detector and gist templates. let me know if you think we should take a different approach with the UI code like directly use the diagnosticApiService in the OnboardingFlowComponent instead of calling the detectorGistApiService.

Sure, Hawk ! Then let's keep detectorgist-template-api service

rekhaswaminathan:
Code changes look good. Just a thought - Have you considered using assets folder? We already use assets folder for `enabledResourceTypes.json` and getting the file from assets folder will not require an API call to AppLensBackend

rekhaswaminathan (approved):

cindylovescoding:
> Code changes look good. Just a thought - Have you considered using assets folder? We already use assets folder for `enabledResourceTypes.json` and getting the file from assets folder will not require an API call to AppLensBackend

Good suggestions! We can also put these on CDN in the future

ApplensBackend/Services/DetectorGistTemplateService/TemplateService.cs
```diff
@@ -0,0 +1,80 @@
...
+{
+
+    public class TemplateService : IDetectorGistTemplateService
+    {
+        private readonly IMemoryCache memoryCache;
+        private Task templateTask;
+
+        public TemplateService(IMemoryCache detectorGistCache)
+        {
+            memoryCache = detectorGistCache;
```
nmallick1, line 26: Do you expect us hitting throttling limits and is that why we are maintaining the cache?

hforeste:
> > Code changes look good. Just a thought - Have you considered using assets folder? We already use assets folder for `enabledResourceTypes.json` and getting the file from assets folder will not require an API call to AppLensBackend
> 
> Good suggestions! We can also put these on CDN in the future

I also like this suggestion and I was considering it but I didn't like the HTTP request would always hit the filesystem aka assets folder for the templates. Then I also thought, it would be nice if in the future we can enable a scenario for partners where they can run CRUD actions on templates and the experience would all be done in the UI.

ApplensBackend/Services/DetectorGistTemplateService/TemplateService.cs
hforeste: I implemented the cache just to avoid having to hit FileSystem every time for each template call. Trying to reduce dependency on the FileSystem as much as possible.

hforeste: The templates also rarely change so no need to do IO every time for something that doesn't really change that much.

nmallick1: I actually thought we were hitting DevOps API's to fetch detector templates like we used to do with GitHub. Yeah, this makes sense. Thanks!

SharmaAjay19 (approved):

PR #1584, opened 2022-05-11
Fix theming for network troubleshooter and case submission flow searchbox

cindylovescoding:
image


XiaoxuHU (approved):

sidkri (approved):
Thanks for fixing this!

PR #1620, opened 2022-06-03
Store recent resource query params and update time string format

XiaoxuHU:
1. Store query params (startTime,endTime, caseId, ect...) when updating recent resource. 
    When click recent resource it will route to time period which fetching from cosmosDB instead of last 24 hours
2. Change time string format to **"YYYY-MM-DD HH:mm"** in time-picker and downtime selection to be consistency. And also allow to type as this format in date-picker
Todo, after this PR deployed. Also need to update time string format for downtime gist(DowntimesGist) and overview metric detector(applensoverviewmetrics_appservices)

image

### DIFF
--- a/AngularApp/projects/diagnostic-data/src/lib/services/detector-control.service.ts
+++ b/AngularApp/projects/diagnostic-data/src/lib/services/detector-control.service.ts
@@ -330,9 +330,8 @@ export class DetectorControlService {
     if (updatedInfo && updatedInfo.selectedKey !== TimePickerOptions.Custom) {
       this.timePickerStrSub.next(updatedInfo.selectedText);
     } else {
-      const timeFormat = 'M/D/YY HH:mm';
-      const st = moment(this.startTimeString).format(timeFormat);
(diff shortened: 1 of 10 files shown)


AngularApp/projects/diagnostic-data/src/lib/services/detector-control.service.ts
```diff
@@ -75,6 +75,9 @@ export class DetectorControlService {
 
   constructor(@Inject(DIAGNOSTIC_DATA_CONFIG) config: DiagnosticDataConfig) {
     this.internalClient = !config.isPublic;
+    this._duration = this.durationSelections[2];
```
SharmaAjay19, line 78: You might want to make this safe against index error. Can check not null and array size before indexing.

SharmaAjay19: Please ignore this, just verified it's a declared array above.

SharmaAjay19 (approved):

cindylovescoding (approved):

PR #1624, opened 2022-06-13
internal passthrough

DarrelDonald:


AngularApp/projects/applens/src/app/modules/dashboard/onboarding-flow/onboarding-flow.component.ts
```diff
@@ -1463,14 +1494,20 @@ export class OnboardingFlowComponent implements OnInit {
       `/${this.id.toLowerCase()}/package.json`
     ];
 
-    if (this.autoMerge) {
-      this.Branch = this.defaultBranch;
+    if(Object.keys(this.DevopsConfig.appTypeReviewers).length > 0 || Object.keys(this.DevopsConfig.platformReviewers).length > 0){
+      gradPublishFiles.push("delete owners.txt");
+      gradPublishFileTitles.push(`/${this.id.toLowerCase()}/owners.txt`);
```
rekhaswaminathan, line 1499: I didn't quite understand this part. Why do we need to delete the owners.txt file?

DarrelDonald: This is for when the detector is being deleted. removing all of the files in the detector folder

rekhaswaminathan:
Code changes look good. Thanks @DarrelDonald for working on this! 
Can you please make sure the following scenarios are tested & make sure we are not introducing any regressions?
[the list below is not exhaustive] 
- Test with Resource which does not have reviewer or internal pass-through flag
- Test with Resource which has  internal pass through flag + only App Type filter (eg. WebApp, FunctionApp)
- Test with Resource which has internal pass through flag + only Platform Type filter (eg. Windows, Linux)
- Test with Resource which has detector created already with owners.txt. (Make sure existing owner aliases do not get removed)
- Test with Resource which has only internal pass through flag but no reviewers. Which means, internal only detectors directly get pushed but internal + external gets PR created
- Test with Resource which has detector code changed from internal to external (Make sure we honor the code change & add reviewers)
- Test with Resource which has detector code changed from external to internal

rekhaswaminathan (approved):

PR #1673, opened 2022-07-28
Fix app insights integration national clouds

puneetg1983:
Correct AppInsights endpoint for National clouds

### DIFF
--- a/Backend/Services/AppInsightsService/AppInsightsService.cs
+++ b/Backend/Services/AppInsightsService/AppInsightsService.cs
@@ -74,7 +74,7 @@ namespace Backend.Services
             {
                 return AppInsightsEndpointAzureGov;
             }
-            else if (siteHostName.Contains(".azurewebsites.cn", StringComparison.OrdinalIgnoreCase))
+            else if (siteHostName.Contains(".chinacloudsites.cn", StringComparison.OrdinalIgnoreCase))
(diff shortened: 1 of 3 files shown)


Backend/Services/AppInsightsService/AppInsightsService.cs
```diff
@@ -54,5 +57,29 @@ public async Task<bool> Validate(string appInsightsAppId, string encryptedKey)
...
+        /// Gets the name of the correct AppInsights default endpoint based on the hostname where the UI project is hosted
+        /// </summary>
+        /// <param name="siteHostName"></param>
+        /// <returns></returns>
+        private string GetAppInsightsDefaultEndpoint(string siteHostName)
+        {
+            if (string.IsNullOrWhiteSpace(siteHostName))
+            {
+                return AppInsightsEndpointPublicAzure;
+            }
```
hforeste, line 71: You don't need to add this conditional statement since you are by default returning AppInsightsEndpointPublicAzure in line 82

hforeste (approved):

Backend/Services/AppInsightsService/AppInsightsService.cs
puneetg1983: updated the PR. Thanks!

PR #1684, opened 2022-08-09
Self-serve feature for managing ICM automation templates

SharmaAjay19:
First screen displays the list of team templates that user has access to
image

As user selects a team, the template is displayed in JSON format as below:
image

After modifying the template, user has to click on "Test Template" button, to validate the template against real incident (s).
On the Test Incident blade, user sees a list of recent incidents on the team whose template they are modifying:
image

They can use any of these incidents, in the Incident Id field, as well as provide any other incident id. Then they click on the "Test Incident" button:
image

We run the validations on the data from that incident and show them the results:
image

After confirming expected results, they can then update the template using "Update Template" button:
image

If there are no templates that the user has access to, they will see this message:
image


AngularApp/projects/applens/src/app/modules/incidentassist/components/template-management/template-management.component.ts
```diff
@@ -0,0 +1,276 @@
...
+      (err) => {
+        this.displayLoader = false;
+        this._telemetryService.logEvent(TelemetryEventNames.ICMTeamTemplateUpdate, {"TeamId": this.selectedTeam.teamId, "IncidentType": this.selectedTeam.incidentType.toString(), "Status": "failed", "ErrorMessage": err.error, userId: this.userId});
+        this.footerMessage = `Failed to update team template because of ${err.error}`;
+        this.footerMessageType = "error";
+      });
+    }
+  }
+
+  resetLoadedTeam(){
```
rekhaswaminathan, line 242: Empty method?

nmallick1 (approved):

rekhaswaminathan (approved):

PR #1704, opened 2022-08-30
Changing AppInsights Encryption Service logic to use X509Certificate2

puneetg1983:
With this PR, the encryption service will now use X509Certificates to Encrypt and decrypt the AppInsights API key. The new encryption service logic will be enforced **ONLY** by change to the following settings.
- **AppInsights:UseCertficates** -> True in Backend Project.
- **useAppSettingsForAppInsightEncryption** in appinsights.service.ts -> true

Both of the above should be true for the latest logic to pick up. This should be done however after ANT99 when the dynamic environment variable **WEBSITE_APPINSIGHTS_ENCRYPTEDAPIKEY** takes effect

On the backend side, a new service (CertificateService) is added that pulls the certificate from KeyVault every hour and constructs an X509Certificate if the thumbprint doesn't match with what is already loaded in memory. This is done to avoid recreating the certificates again and again. This service loads all the certificates (valid and expired) with the Certificate Name (specified by **AppInsights:CertificateName** app setting). When a cert rotation happens, we decrypt the hash using the latest or the last 5 expired certificates (can be controlled using **AppInsights:MaxExpiredCertificateCount**). If the existing hash is encrypted using an older certificate, the UI makes a call to update the encrypted blob automatically and this is how the encrypted hash stays valid.

Additional app settings for all UI projects should be created before merging this PR
- AppInsights:CertificateName
- Secrets:DevKeyVaultName
- Secrets:StagingKeyVaultName
- Secrets:ProdKeyVaultName

We would also need a PR to RuntimeHost but wanted to get this validated before doing further work.


ShekharGupta1988 (approved):

rekhaswaminathan:
One thing I'd make sure if Portal backend Webapps has the necessary access control policies on the keyvault before deploying this.

rekhaswaminathan (approved):

XiaoxuHU:
Looks good to me, please resolve conflict before merge PR

PR #1721, opened 2022-09-08
Update onboarding-flow.component.ts

DarrelDonald:


AngularApp/projects/applens/src/app/modules/dashboard/onboarding-flow/onboarding-flow.component.ts
```diff
@@ -1581,8 +1584,8 @@ export class OnboardingFlowComponent implements OnInit {
       this.publishingPackage.packageConfig
     ] : [
       this.code,
-      "",
-      ""
+      this.existingMetadata,
+      this.existingPackage
```
rekhaswaminathan, line 1588: Do we really need new variables? We already have` this.allUtterances` and `this.configuration`

DarrelDonald: those variables have the parsed data from the files but we have to pass the actual strings that they were parsed from here

rekhaswaminathan: Will JSON.stringify work?

rekhaswaminathan: Just trying to assess the need for new variables here since this class already have a TON of variables and would be difficult to maintain going forward

DarrelDonald: it works for the package but not for metadata. what if we just leave them empty? it will be corrected after the successful run, which is required before the pull request can be made through applens anyway

rekhaswaminathan: There could be a state where someone can just raise a PR through devops directly after saving on AppLens. It'll be fine to leave them empty for new detectors but not existing detectors wiping out the package and metadata might have other downstream effects.
The previous code does https://github.com/Azure/Azure-AppServices-Diagnostics-Portal/blob/c2f579fcf9324ae8828c4688549e8e479506ecaf/AngularApp/projects/applens/src/app/modules/dashboard/onboarding-flow/onboarding-flow.component.ts#L1774 `JSON.stringify({ "utterances": this.allUtterances })`
Doesn't  `JSON.stringify({ "utterances": this.allUtterances })` work?

DarrelDonald: updated

rekhaswaminathan (approved):

PR #1725, opened 2022-09-12
Send location header

hforeste:
Send the location of ARM resource to the RP and eventually to our backend. Detectors consume this value during runtime and will save our backend from having to make an extraneous call to get location information.


AngularApp/projects/app-service-diagnostics/src/app/shared/services/arm.service.ts
```diff
@@ -133,7 +134,7 @@ export class ArmService {
         return uri;
     }
 
-    getResource<T>(resourceUri: string, apiVersion?: string, invalidateCache: boolean = false): Observable<{} | ResponseMessageEnvelope<T>> {
+    getResource<T>(resourceUri: string, apiVersion?: string, invalidateCache: boolean = false, armResource?: ArmResource): Observable<{} | ResponseMessageEnvelope<T>> {
```
ShekharGupta1988, line 137: the getResource method signature feels a bit counter-intuitive as its now taking an ARM resource as optional input. 
This might confuse the developers. 

Instead of passing the armResource, one small suggestion would be to pass optional additionalRequestHeaders. You can just add them to the request headers (and dedupe as well). In that way, its pretty extensible for future.

What do you think?

AngularApp/projects/app-service-diagnostics/src/app/shared/services/generic-api.service.ts
```diff
@@ -73,8 +74,16 @@ export class GenericApiService {
             if (additionalQueryParams != undefined) {
                 path += additionalQueryParams;
             }
-            return this._armService.getResource<DetectorResponse>(path, null, refresh).pipe(
-                map((response: ResponseMessageEnvelope<DetectorResponse>) => response.properties));
+
+            return this._armService.getArmResource<ArmResource>(resourceId).pipe(
```
ShekharGupta1988, line 78: Can you verify once if this is hitting the cache and not again making an outbound ARM call to getResource. (Since we already make this call during the start up as part of resource service)

hforeste: Already verified during development

XiaoxuHU (approved):

ShekharGupta1988 (approved):

PR #1726, opened 2022-09-12
Fix for missing searchTerm for query params calling keystone detector

XiaoxuHU:


ShekharGupta1988 (approved):

AngularApp/projects/app-service-diagnostics/src/app/shared-v2/services/support-topic.service.ts
```diff
@@ -227,7 +227,9 @@ export class SupportTopicService {
                         let fakeStartTimeString = startTime.format('YYYY-MM-DD HH:mm');
                         let fakeEndTimeString = endTime.format('YYYY-MM-DD HH:mm');
 
-                        return this._diagnosticService.getDetector(keystoneDetectorId, fakeStartTimeString, fakeEndTimeString, false, false, "&extractKeywords=true").pipe(map(keystoneRes => {
+                        let formQueryParmas  ="&extractKeywords=true";
+                        if(searchTerm) formQueryParmas = formQueryParmas + `&searchTerm=${searchTerm}`;
```
SharmaAjay19, line 231: I think this part is just to fetch the keywords from the detector to match on the UI for whether keystone detector should be triggered. This call does not require the search term.

For actual solutions, there is another call to the keystone detector in integrated solutions component.

SharmaAjay19: However checking the code and current experience on windows this does not need to be explicitly done since it's already working there without this.

XiaoxuHU:
Close this PR as discussed with Ajay for getting keyword list we do not need to send request with searchTerm query parameter

PR #1763, opened 2022-10-04
add Observer client cert auth for Applens backend

yifguoMSFT:


SharmaAjay19 (commented):
We may be able to reuse the GenericCertLoaderService instead of creating a dedicated one.
https://github.com/Azure/Azure-AppServices-Diagnostics-Portal/blob/main/ApplensBackend/Services/GenericCertLoaderService/GenericCertLoader.cs

ApplensBackend/Configuration/Startup.cs
```diff
@@ -58,6 +58,7 @@ public void ConfigureServices(IServiceCollection services)
             services.AddSingleton(Configuration);
 
             GenericCertLoader.Instance.Initialize();
+            SupportObserverCertLoader.Instance.Initialize(Configuration);
```
SharmaAjay19, line 61: Quick check - we may need to update the Startup file for NationalCloud as well.

yifguoMSFT:
> We may be able to reuse the GenericCertLoaderService instead of creating a dedicated one. https://github.com/Azure/Azure-AppServices-Diagnostics-Portal/blob/main/ApplensBackend/Services/GenericCertLoaderService/GenericCertLoader.cs

Hi Ajay, I already used the GenericCertLoader in the SupportObserverCertLoader, see [here](https://github.com/Azure/Azure-AppServices-Diagnostics-Portal/blob/4b877d51be82881116bfdeb08dc24726d1e1382b/ApplensBackend/Services/ObserverClientService/SupportObserverCertLoader.cs#L22). It's just for reading the Observer appsetting and decide whether load the cert or not and throw an exception when cert is not found

ApplensBackend/Configuration/Startup.cs
yifguoMSFT: Didn't see the configuration files for sovereign cloud in the project, will check with Hawk regarding to this.

And I haven't enabled the client cert yet, so SupportObserverCertLoader here won't load anything which should be fine.

yifguoMSFT: chatted with Hawk, created appsettings.<cloud>.json for different cloud settings and added cert loader to StartupNationalCloud.cs

SharmaAjay19 (approved):

PR #1798, opened 2022-11-02
add new UI component: client script component

yifguoMSFT:
- Create a new `ClientScriptView` UI component for triggering a client script based check in diagnostics portal to do some diagnostics that cannot be done in a normal detector script, e.g. run a tcpping test
- Use `ClientScriptView` to implement a simple VNet integration check

image


AngularApp/package.json
```diff
@@ -5,7 +5,7 @@
   "scripts": {
     "ng": "ng",
     "start": "ng serve",
-    "applens": "npm run fixAngularReact && node --max_old_space_size=8192 node_modules/@angular/cli/bin/ng serve applens",
+    "applens": "npm run fixAngularReact && node --max_old_space_size=8192 node_modules/@angular/cli/bin/ng serve --host 0.0.0.0 applens",
```
XiaoxuHU, line 8: Just curious why want to add 0.0.0.0 for host

yifguoMSFT: I use docker container to run the angular project, so that I need to expose its port to make it accessible from outside of the docker container

XiaoxuHU: Can you also test if it is still working when running "npm run applens" without container, will this command still work as before ?

XiaoxuHU: Test on my end, 0.0.0.0 working as usual

XiaoxuHU (approved):

PR #1801, opened 2022-11-02
Nmallick1/allow multiple detectors with same support topic

nmallick1:
Enable the UI to handle the scenario when multiple detectors are mapped to the same support topic. In such a scenario, an analysis view will be shown merging all the detectors that were tagged. 
If one of the detectors that were tagged is an Analysis, then all the children of that analysis will be shown as well.
This hierarchy is only one level deep.
In the below screenshot, app down analysis, ARRAffinity detector and AlwaysOnCheck detector were tagged to the support topic. The resulting view contains a combined analysis view of all the participating detectors.

image

image


XiaoxuHU (approved):

SharmaAjay19 (approved):
You're playing with hot molten lava.

ShekharGupta1988 (approved):

PR #1806, opened 2022-11-08
Fix input and add L3 support topic for all detector list

XiaoxuHU:
- Fix for input in Applens where cannot delete last character 
- Update logic for showing L2 and L3 support topic in all detector list

image


ShekharGupta1988 (approved):

ShekharGupta1988:
@XiaoxuHU - the screenshot is showing same values in L2 and L3 column

XiaoxuHU:
> @XiaoxuHU - the screenshot is showing same values in L2 and L3 column

Yes, @ShekharGupta1988. Value in screenshot is just **mock up** as we are not able to get real support topic list in local

nmallick1 (approved):

PR #1817, opened 2022-11-16
Add bread crumb to go back in portal

XiaoxuHU:
Add Go back button to enable go back from child detector to analysis or parent detector

In Diag&Solve blade,
image

In case submission
image


nmallick1:
@XiaoxuHU , does this preserve query params for Forms, Downtimes, CaseTitle etc?

nmallick1 (approved):

PR #1850, opened 2022-12-10
National cloud

hforeste:
## Overview

Applens UI and backend UI project has evolved into a state that it wasn't compatible with soveriegn clouds and there also runtime errors for services that were not injected. This change updates the project to be compatible and it also provides a better framework for developers to contribute to the project while thinking of sovereign clouds.

## Related Work Item

Internal DevOps

## Type of change

### This PR:

- Cleanly separate setting up of configurations in `StartupPublicAzure` and `StartupNationalCloud` by calling `AddConfigurations`
- Enable clientCertAuth with Runimehost in all clouds.
  - Add `DiagnosticRole:clientCertEnabled` setting for flexibility
- We've disabled the UI's Adal functionality in non-PublicAzure and replaced with dSTS. This has been the case for a long time but with the UserSettings feature
- Add placeholder API for getuserinfo when user_id is blank
   - UI makes a GET to getuserinfo API with no user_id because Adal doesn't run in non-PublicAzure scenario. This was previously breaking the request flow with a 415 Method Not Found status code.
- Update dSTS auth flow to challenge the request in the case of an unauthenticated user. This is a more industry standard approach as opposed to the redirects that was there previously.
   - Currently when the session cookie expires for dSTS and a user is still in the middle of using detectors. The detectors will immediately fail with no warning as to why. This is because when the UI code is calling the backend, the API challenges the unauthenticated request but the UI code doesn't recognize the redirect to dSTS. We need future bug fix for this but its not blocking the PR.
- Create a base appsettings.json file for national cloud scenarios so as to not repeat settings in individual cloud appsettings.

- Local Testing
- Testing dSTS auth flow with SSL enabled
- Verified PublicAzure and non-PublicAzure scenarios are working fine
  - Load detectors
  - Ran applicationcrashes and another detector 

- [N/A] I have ensured that my changes support accessibility and can be accessed with the keyboard alone


ApplensBackend/Configuration/Startup.cs
```diff
@@ -119,8 +114,10 @@ public void ConfigureServices(IServiceCollection services)
                 GraphTokenService.Instance.Initialize(Configuration);
             }
 
-            // If we are using runtime host directly
-            DiagnosticClientToken.Instance.Initialize(Configuration);
+            if (!Configuration.GetValue<bool>("DiagnosticRole:clientCertEnabled", true))
+            {
+                DiagnosticClientToken.Instance.Initialize(Configuration);
```
rekhaswaminathan, line 119: Should we clean `DiagnosticClientToken` up since it is not needed anymore?

hforeste: I wanted to move everything to DiagnosticClientToken in national clouds first before doing this

ApplensBackend/Configuration/StartupNationalCloud.cs
```diff
@@ -1,24 +1,69 @@
 ﻿using System.IO;
```
yifguoMSFT, line 1: now that we have cloud specific app settings, do we still need an extra startup.cs for national cloud?

yifguoMSFT (approved):

ApplensBackend/Configuration/StartupNationalCloud.cs
hforeste: Honestly, we can rely on just one startup.cs file we'd just have to add conditional statements for different scenarios. I prefer to hide implementation details by classes versus conditional logic.

rekhaswaminathan (approved):

PR #1866, opened 2022-12-21
National cloud

hforeste:
## Overview

This change updates out cert loaders to be dependency injected versus be used as static classes. This change was needed so that we can use the built-in asp.net core dependency injected ILogger infrastructure. We need this for debugging purposes.

## Related Work Item

This is related to the bigger work item of getting applens functional in some sovereign clouds.

## Type of change

- [X] New feature (non-breaking change which adds functionality)
- [X] Refactor

### This PR:

- Removes some level of tight coupling between classes and the cert loader helper
- Dependency Injects cert loader as opposed to reference via static class
- Add logging capabilities to the cert loader class
- Abstract logging behind ILogger so that tomorrow we can add logging providers like EventSource
- Add File Logging as a custom logging provider. Writes to a diagnostic log file and adheres to the built-in asp.net core logging.LogLevel configuration settings
- File Logging is behind a feature flag. Default is disabled

Tested on local machine and a private web app of applens


yifguoMSFT (approved):

ApplensBackend/Program.cs
```diff
@@ -50,6 +61,29 @@ public static IHostBuilder CreateHostBuilder(string[] args)
...
+                });
+            }
+
+            if (config.GetValue("ASPNETCORE_ENVIRONMENT", "Production").Equals("Production", StringComparison.CurrentCultureIgnoreCase))
+            {
+                webHostBuilder.ConfigureLogging((logging) =>
+                {
+                    logging.ClearProviders();
+
+                    if (config.GetValue<bool>("FileLogging:Enabled"))
```
ShekharGupta1988, line 80: Is the plan to enable File Logging only for S-clouds or Public cloud as well?

hforeste: its just temporarily logging that can be enabled with any cloud. Right now I have it enabled for s-clouds but I will later disable it.

ShekharGupta1988 (approved):

ApplensBackend/Program.cs
hforeste: What I am excited about though is adding the plumbing for event source. This work sets us up for that.

ApplensBackend/Configuration/Startup.cs
```diff
@@ -66,8 +66,9 @@ public void ConfigureServices(IServiceCollection services)
 
             services.AddSingleton(Configuration);
 
-            GenericCertLoader.Instance.Initialize();
-            SupportObserverCertLoader.Instance.Initialize(Configuration);
+            services.AddSingleton<GenericCertLoader>();
+
+            services.AddSingleton<SupportObserverCertLoader>();
```
nmallick1, line 71: Do you think we need SupportObserverCertLoader anymore?

ApplensBackend/Program.cs
```diff
@@ -50,6 +61,29 @@ public static IHostBuilder CreateHostBuilder(string[] args)
...
+                    logging.AddConsole();
+                    logging.AddDebug();
+                });
+            }
+
+            if (config.GetValue("ASPNETCORE_ENVIRONMENT", "Production").Equals("Production", StringComparison.CurrentCultureIgnoreCase))
+            {
+                webHostBuilder.ConfigureLogging((logging) =>
+                {
+                    logging.ClearProviders();
```
nmallick1, line 78: Will AppInsight logging continue to work? I forgot if our backend project sends logs to AppInsights or is it just the Angular code that does the logging.

ApplensBackend/Configuration/Startup.cs
hforeste: To be honest, I just kept it because I didn't want to break the current flow it had with SupportObserverClient. Jeff can take that call whether he wants to keep it or not.

nmallick1 (approved):

ApplensBackend/Program.cs
hforeste: Thats a good question and not sure. I know our UI jarvis dashboard is based on Angular AppInsights though. As far as the applens backend, just to be on the safe side I've added the below, `logging.AddApplicationInsights()` .

PR #1869, opened 2023-01-05
National cloud

hforeste:
## Overview

This change fixes a few issues that has occurred in non-Public clouds. 

1. Our UI code does not use Adal in non-PublicClouds and it attempts to fetch userid from adal but when its empty it results in 405 status code from the API. This code change fixes that assumption.
2. Added additional logging to the GenericCertLoader class.
3. Some certs have their subjectname as CN=...,OU=....,OU=...,DC=... etc. I written a small utility which will extract the common name from the subject name so that no matter the subject name the expected behavior will continue to happen.

## Related Work Item

Ping internally for DevOps work item.

## Type of change

Build applens UI project with `npm run build-applens-ncloud-prod` and simply list detectors to hit the generic cert loader code path as well as the usersettings controller

N/A no UI change

N/A no UI change

## Extra Notes

Ping me if any more details needed


ApplensBackend/Helpers/Utilities.cs
```diff
@@ -142,5 +143,15 @@ public static string GetUserIdFromToken(string authorizationToken)
 
             return userId;
         }
+
+        /// <summary>
+        /// Get the common name of the certificate subject name that is the part that part in the subject name that starts with CN=.
+        /// </summary>
+        /// <param name="certificate">certificate.</param>
+        /// <returns>Common Name</returns>
+        public static string GetSubjectCommonName(this X509Certificate2 certificate)
```
nmallick1, line 152: @hforeste , would be a good idea to include this utility in the runtimeHost project too.

nmallick1 (approved):

ShekharGupta1988 (approved):

PR #1890, opened 2023-01-25
Handle case number needed error gracefully

SharmaAjay19:
## Overview
Since we are planning to expand Durian to more products, we will have some scenarios where Durian configuration is not in sync between Applens and Runtimehost. This can cause some bad situations where the user is stuck. This PR adds a very critical logic to handle error gracefully even if there is a configuration mismatch. This particular change is to handle case number needed error by showing the case number field to the user.

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## Solution description
Durian is enabled for any given product from Runtimehost and Applens separately. This can lead to situations where Runtimehost and Applens are not updated synchronously and while the Runtimehost needs a case number, Applens is not asking the user for one (as the frontend config is not yet updated). To handle this - **if the Runtimehost throws an error asking for a case number, the frontend should assume Durian is enabled for the product and should show the field for the user to provide the case number**. This solution achieves that.

A CSS user is asked for case number while accessing AppLens for certain products. This change can be tested by enabling Durian for a product in backend (Runtimehost) but not in Frontend (AppLens). Initially when user comes to AppLens, they will not see case number field since frontend doesn't have Durian enabled for the product. However, once they enter information and try to move further, backend will throw an error asking for case number (since Durian is enabled for the product on the backend). The above solution should handle that error gracefully and show the user case number field to fill in the data.

- [x] I have looked over the diffs.
- [x] I have run the linter to catch any errors.
- [x] There are no errors from my changes on this branch.
- [x] I have ensured that my changes support accessibility and can be accessed with the keyboard alone
- [x] I have requested review on this PR.

Initial Screen (Durian is not enabled on frontend for AKS resources, so the user is not asked for case number)
image

However when they try to enter Applens:
image

Here we can see user is seeing the error message about case number but no field to provide it. This is due to configuration mismatch between frontend (Applens) and backend (Runtimehost).

After this fix, user will get an option to provide case number when the error occurs
image

### DIFF
--- a/AngularApp/projects/applens/src/app/modules/main/main/main.component.ts
+++ b/AngularApp/projects/applens/src/app/modules/main/main/main.component.ts
@@ -105,7 +105,7 @@ export class MainComponent implements OnInit {
       this.resourceTypes = this.resourceTypes.filter(resourceType => !resourceType.caseId);
     }
 
-    if (this._activatedRoute.snapshot.queryParams['caseNumber'] && this._activatedRoute.snapshot.queryParams['caseNumber'] !== "undefined") {
+    if (this._activatedRoute.snapshot.queryParams['caseNumber'] && this._activatedRoute.snapshot.queryParams['caseNumber'] !== "null" && this._activatedRoute.snapshot.queryParams['caseNumber'] !== "undefined") {
       this.caseNumber = this._activatedRoute.snapshot.queryParams['caseNumber'];
     }
     if (this._activatedRoute.snapshot.queryParams['resourceName']) {
--- a/AngularApp/projects/applens/src/app/modules/main/main/main.component.ts
+++ b/AngularApp/projects/applens/src/app/modules/main/main/main.component.ts
@@ -105,7 +105,7 @@ export class MainComponent implements OnInit {
       this.resourceTypes = this.resourceTypes.filter(resourceType => !resourceType.caseId);
     }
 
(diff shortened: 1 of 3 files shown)


AngularApp/projects/applens/src/app/modules/main/main/main.component.ts
```diff
@@ -104,7 +105,7 @@ export class MainComponent implements OnInit {
       this.resourceTypes = this.resourceTypes.filter(resourceType => !resourceType.caseId);
     }
 
-    if (this._activatedRoute.snapshot.queryParams['caseNumber'] && this._activatedRoute.snapshot.queryParams['caseNumber'] !== "undefined") {
+    if (this._activatedRoute.snapshot.queryParams['caseNumber'] && this._activatedRoute.snapshot.queryParams['caseNumber'] !== "null" && this._activatedRoute.snapshot.queryParams['caseNumber'] !== "undefined") {
```
puneetg1983, line 108: You are checking for a string who value is "null" ?

SharmaAjay19: Unfortunately yes, the browser parameters treats the value null as the string "null"

SharmaAjay19: Btw found a better way of dealing with this.

nmallick1 (approved):

h-mangipudi (approved):

PR #1933, opened 2023-03-01
Allow configuring workflow users by storing them in cosmosDB

puneetg1983:
## Overview
Instead of changing the `Workflow:Users` app setting on all the endpoints, storing the workflow users in CosmosDb and giving a small UI to configure them in AppLens.

Only those users that are part of the existing `Workflow:Users` app setting will be able to add new users to CosmosDB.

### DIFF
--- a/ApplensBackend/Configuration/Startup.cs
+++ b/ApplensBackend/Configuration/Startup.cs
@@ -72,6 +72,7 @@ namespace AppLensV3
 
             services.AddSingleton<IObserverClientService, SupportObserverClientService>();
             services.AddSingleton<IDiagnosticClientService, DiagnosticClient>();
+            services.AddSingleton<IWorkflowUsersCacheService, WorkflowUsersCacheService>();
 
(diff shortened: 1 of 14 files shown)


XiaoxuHU (approved):

ApplensBackend/Configuration/Startup.cs
```diff
@@ -103,6 +104,8 @@ public void ConfigureServices(IServiceCollection services)
 
             services.AddSingletonWhenEnabled<ICosmosDBUserSettingHandler, CosmosDBUserSettingHandler, NullableCosmosDBUserSettingsHandler>(Configuration, "UserSetting");
 
+            services.AddSingletonWhenEnabled<ICosmosDBWorkflowUsersHandler, CosmosDBWorkflowUsersHandler, NullableCosmosDBWorkflowUsersHandler>(Configuration, "UserSetting");
```
SharmaAjay19, line 107: This value should not be "UserSetting", it should probably be the name of the collection i.e. WorkflowUsers

puneetg1983: Good catch. Changed.

PR #1970, opened 2023-03-27
Enable Workflows on 'detector' routes only

puneetg1983:
## Overview
To avoid making a change to our GeoMaster, we are enabling list workflow and execute workflow node on existing /detectors/ route only.

For workflows an additional parameter diagnosticEntityType=workflows will be passed in query string and backend will return detectors vs workflows based on this query parameter and other query parameters passed for executing workflow node Id.

After this change, we will not need to update GeoMaster code to enable Workflows in Diagnose and Solve.

### DIFF
--- a/AngularApp/projects/applens/src/app/shared/services/diagnostic-api.service.ts
+++ b/AngularApp/projects/applens/src/app/shared/services/diagnostic-api.service.ts
@@ -78,21 +78,17 @@ export class DiagnosticApiService {
     internalView: boolean = true, additionalQueryParams?: string, userInputs?: any):
     Observable<workflowNodeResult> {
     let timeParameters = this._getTimeQueryParameters(startTime, endTime);
-    let path = `${version}${resourceId}/workflows/${workflowId}?${timeParameters}`;
-    if (workflowExecutionId && nodeId) {
-      path = `${version}${resourceId}/workflows/${workflowId}/${workflowExecutionId}/${nodeId}?${timeParameters}`;
-    }
-
+    let path = `${version}${resourceId}/detectors/${workflowId}?${timeParameters}&diagnosticEntityType=workflow&workflowNodeId=${nodeId}&workflowExecutionId=${workflowExecutionId}`;
+    
     if (additionalQueryParams != undefined) {
       path += additionalQueryParams;
     }
 
-    let body = null;
     if (userInputs != null) {
-      body = { ResourcePostBody: null, UserInputs: userInputs };
+      path += `&userInputs=${JSON.stringify(userInputs)}`;
     }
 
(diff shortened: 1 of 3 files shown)


AngularApp/projects/applens/src/app/shared/services/diagnostic-api.service.ts
```diff
@@ -112,10 +108,13 @@ export class DiagnosticApiService {
...
   public getWorkflows(version: string, resourceId: string, body?: any, queryParams?: any[], internalClient: boolean = true): Observable<DetectorMetaData[]> {
-    let path = `${version}${resourceId}/workflows`;
-    if (queryParams) {
-      path = path + "?" + queryParams.map(qp => qp.key + "=" + qp.value).join("&");
+    let path = `${version}${resourceId}/detectors`;
+    if (queryParams == null) {
+      queryParams = [];
     }
+
+    queryParams.push({ key: 'diagnosticEntityType', value: 'workflow' });
```
nmallick1, line 116: Do you want to check if this query param was already present just in case the supplied queryParams were not empty?

nmallick1 (approved):

AngularApp/projects/applens/src/app/shared/services/diagnostic-api.service.ts
```diff
@@ -75,24 +75,20 @@ export class DiagnosticApiService {
...
+    
     if (additionalQueryParams != undefined) {
       path += additionalQueryParams;
     }
 
-    let body = null;
-    if (userInputs != null) {
-      body = { ResourcePostBody: null, UserInputs: userInputs };
+    if (workflowUserInputs != null) {
+      path += `&workflowUserInputs=${JSON.stringify(workflowUserInputs)}`;
```
SharmaAjay19, line 88: Should we encode this json string?

SharmaAjay19 (approved):

AngularApp/projects/applens/src/app/shared/services/diagnostic-api.service.ts
puneetg1983: The caller code `(/components/workflow-accept-userinput/workflow-accept-userinput.component.ts)` is correctly html encoding the value

AngularApp/projects/applens/src/app/shared/services/diagnostic-api.service.ts
puneetg1983: Good point. Will add that check!

puneetg1983: done.

PR #1979, opened 2023-04-01
Fix for undefined subscription id when calling ARM 

XiaoxuHU:
## Overview

This PR is for 
- Fix for race condition when retrieving location with subscription id
- Improving logging in arm.service for error handling

## Type of change

Go to portal -> Select "High CPU Analysis" -> Check if still having "https://management.azure.com/subscriptions/undefined" call

### DIFF
--- a/AngularApp/projects/app-service-diagnostics/src/app/resources/web-sites/services/site-feature.service.ts
+++ b/AngularApp/projects/app-service-diagnostics/src/app/resources/web-sites/services/site-feature.service.ts
@@ -18,6 +18,9 @@ import { CategoryService } from '../../../shared-v2/services/category.service';
 import { VersionTestService } from '../../../fabric-ui/version-test.service';
 import { ArmService } from '../../../shared/services/arm.service';
 import { SubscriptionPropertiesService } from '../../../shared/services/subscription-properties.service';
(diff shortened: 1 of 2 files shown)


AngularApp/projects/app-service-diagnostics/src/app/resources/web-sites/services/site-feature.service.ts
```diff
@@ -38,76 +40,82 @@ export class SiteFeatureService extends FeatureService {
       order: ['appdownanalysis', 'perfanalysis', 'webappcpu', 'memoryusage', 'webapprestart']
     }];
     this._authService.getStartupInfo().subscribe(startupInfo => {
-      this.subscriptionId = startupInfo.resourceId.split("subscriptions/")[1].split("/")[0];
```
nmallick1, line 41: Why was this wrong? I thought we always have subscriptionId in the resourceId

nmallick1 (approved):

AngularApp/projects/app-service-diagnostics/src/app/resources/web-sites/services/site-feature.service.ts
XiaoxuHU: Nothing wrong with it, remove it since no using it any other places in this service

PR #2001, opened 2023-04-21
Revert change for max input length limitation

XiaoxuHU:
## Overview

This is the revert change for [PR 1993](https://github.com/Azure/Azure-AppServices-Diagnostics-Portal/pull/1993) to add back maxLength up to 150. Besides adding error message to prompt if length > 150 characters

image

## Type of change

### DIFF
--- a/AngularApp/projects/diagnostic-data/src/lib/components/form/form.component.html
+++ b/AngularApp/projects/diagnostic-data/src/lib/components/form/form.component.html
@@ -13,7 +13,7 @@
           </label>
           <div class="col-sm-8 col-md-6">
             <input type="text" class="form-control" [id]=formInput.internalId [(ngModel)]="formInput.inputValue"
-              (keyup)="inputChanged(formInput)">
+              (keyup)="inputChanged(formInput)" maxlength="150">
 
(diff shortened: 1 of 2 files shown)


AngularApp/projects/diagnostic-data/src/lib/components/form/form.component.html
```diff
@@ -13,7 +13,7 @@ <h3 class="form-title">{{element.formTitle}}</h3>
           </label>
           <div class="col-sm-8 col-md-6">
             <input type="text" class="form-control" [id]=formInput.internalId [(ngModel)]="formInput.inputValue"
-              (keyup)="inputChanged(formInput)">
+              (keyup)="inputChanged(formInput)" maxlength="150">
```
yifguoMSFT, line 16: I think we should provide error message instead of simply trimming the user input. And I am still checking if we want to change this limitation to a higher number

yifguoMSFT (changes requested):

yifguoMSFT (approved):

yifguoMSFT (approved):

rekhaswaminathan (approved):

PR #2011, opened 2023-04-28
Making APIM as 1P service in Applens

ShekharGupta1988:
## Overview

The change is to:
- Add support for services in future to become 1P easily (i.e. we are able to identify arm url from the resource name)
- Since we dont have ARM/Resource Graph access, APIM has a kusto query that would help with the arm url resolution.
- In future, if any other service also has this support, we should be able to very quickly add them,

- This is needed to strengthen the security of our system with resource validation.

image


AngularApp/projects/applens/src/app/modules/armresourceurlfinder/armresourceurlfinder.component.html
nmallick1, line 1: This is very cool!

nmallick1 (approved):

XiaoxuHU (approved):

PR #2065, opened 2023-06-02
Resource Validation for Failed to Create/Deleted Resource Scenario (FE)

osshad:
## Overview
AKS and some other RPs may face this problem where the resource was recently deleted or failed to created. In this case ACIS would return 404 for the resource Id, however the ARM cluster, HttpIncomingRequests table would have the information if the resource was recently deleted or failed creation.

One solution is to first ask the AppLens user whether they think this resource was deleted/failed creation and what was the timestamp for that. Then we could search the ARM cluster for 10 minutes before and after that timestamp to validate the resource Id. If the status is validated, we should store that in the Redis cache so we do not make successive calls to the same resource.

## Related Work Item

https://dev.azure.com/app-service-diagnostics-portal/app-service-diagnostics-portal/_workitems/edit/3304

## Type of change
- [x] New feature (non-breaking change which adds functionality)

### This PR:
 - adds new functions to validate resource in ARM cluster
 - updates the UI to allow users validate AKS resources in ARM cluster
 - fixes a pre-existing issue when selected date in the datetime picker is null

 Due to permission restrictions when accessing the ARM cluster locally, this feature can only be tested in the staging/production environment.
To test
 - Go to Applens homepage
 - Select ARM Resource ID in the Service type dropdown
 - Provide a deleted resource ID
 - Click Continue
 - Because the resource wasn't found, the homepage is re-rendered with an error message "ResourceNotFound: The provided resource is not a valid Azure resource." and two buttons (YEs/No) asking the user if the resource they're trying to find was deleted or failed to create recently
 - If the user clicks the Yes button, the page is updated and the user is asked to provide the time the resource was deleted or failed to create
 - When the user provides the event time, a call is made to RuntimeHost to validate the resource in the ARM cluster. 
 - If the resource exists, it is added to the Redis cache and the user is directed to the Dashboard page
 
- [x] I have looked over the diffs.
~- [ ] I have run the linter to catch any errors.~
- [x] There are no errors from my changes on this branch.
- [x] I have ensured that my changes support accessibility and can be accessed with the keyboard alone
- [x] I have requested review on this PR.

image

image


AngularApp/projects/applens/src/app/modules/main/main/main.component.ts
```diff
@@ -520,6 +525,61 @@ export class MainComponent implements OnInit {
...
+    const dateStringFormat : string = TimeUtilities.fullStringFormat;
+    const eventTime = this.deletedOrCreationFailedResourceEventTime.format(dateStringFormat);
+    const { urlPath, queryParams } = this.extractUrlPathAndQueryParams(targetPathBeforeError);
+    let navigationExtras: NavigationExtras = {
+      queryParams: {
+        ...queryParams
+      }
+    }
+
+    this._diagnosticApiService.validateResourceExistenceInArmCluster(this.queryParams.resourceId, eventTime, "MICROSOFT.CONTAINERSERVICE", "MANAGEDCLUSTERS").subscribe(() => {
```
ShekharGupta1988, line 548: Lets pick the RP and service name from service dropdown

osshad: @ShekharGupta1988 I'm not sure the names in the dropdown will work for this scenario. Since this scenario is only applicable to when users select the ARM Resource ID as the service type, the resource type in the dropdown is `ARMResourceId`
image
In the future, if we need to apply this solution to other scenarios where other service types are selected, then I think we can pick the RP and type from the service dropdown. 
For now, I added some code to extract the resource id and provider from the subscription id. Kindly let me know what you think about this. Thanks

AngularApp/projects/applens/src/app/modules/main/main/main.component.ts
```diff
@@ -520,6 +525,66 @@ export class MainComponent implements OnInit {
...
+    const eventTime = this.deletedOrCreationFailedResourceEventTime.format(dateStringFormat);
+    const { urlPath, queryParams } = this.extractUrlPathAndQueryParams(targetPathBeforeError);
+    let navigationExtras: NavigationExtras = {
+      queryParams: {
+        ...queryParams
+      }
+    }
+
+    const match = this.queryParams.resourceId.match(regex);
+    const { providerName: targetResourceProvider, resourceType: targetResourceType } = match?.groups || {}
```
rekhaswaminathan, line 550: I believe we already have a `UriUtilities `class that does the same thing of extracting resource provider and resource type from the ARM Uri. If its not there, you can move this logic to any Utility class.

AngularApp/projects/applens/src/app/modules/main/main/main.component.ts
```diff
@@ -64,6 +65,10 @@ export class MainComponent implements OnInit {
   status = HealthStatus.Critical;
   targetPathBeforeError: string = "";
   isNoResource:boolean = false;
+  isArmResourceRelatedError : boolean;
```
rekhaswaminathan, line 68: do you want to init these to false?

ShekharGupta1988 (approved):

AngularApp/projects/applens/src/app/modules/main/main/main.component.ts
osshad: Good catch. I found out that `ResourceDescriptor.parseResourceUri` does this already. Will update it now. Thanks

rekhaswaminathan (approved):

PR #2089, opened 2023-06-19
Update LinkInterceptorService to handle categories properly for hyperlinks

puneetg1983:
## Overview
Detectors today are using the method `DetectorUtils.GetDetectorLink` and this method is passing a URL back to the UI code which gets intercepted via the LinkInterceptorService but the click of the link is not working properly for a lot of detectors. The most visible example being the **Application Logs** link shown for Linux under Linux Web App Down.

The logic for `GetDetectorLink ` was written way before categories were introduced due to which the detectors are just passing a string like 'detectors/{detectorId}. When an end user clicks on such a hyperlink, they will not be able to the category menu on the left and this breaks the user experience.

## Type of change

## Solution description
This PR will append 'category/{categoryId} based on the right detector or workflow or analysisId to the route path thus ensuring that the link clicked is displayed on the right route and this ensures that the left category menu gets displayed properly.

A lot of functions detectors are passing the link as "../../" or including the right categoryId so the change we did is also supporting all those type of URLs.

image

image

With this PR, the click to Application Logs takes us to the right Application Logs detector with categories displayed on the left.

image


AngularApp/projects/app-service-diagnostics/src/app/shared-v2/services/link-interceptor.service.ts
```diff
@@ -0,0 +1,128 @@
...
+  // we manage to find a category for this detector or analysis and
+  // if the existing url does not contain '/category/' at the right
+  // location.
+  //
+
+  addCategoryIdIfNeeded(linkURL: string) {
+    const entityTypes = ['detectors', 'analysis', 'workflows'];
+    for (let index = 0; index < entityTypes.length; index++) {
+      const entityType = entityTypes[index];
+      if (linkURL.indexOf('/' + entityType + '/') > -1) {
```
nmallick1, line 95: Is there a possibility that someone might give only the detectorId in the link and not mention whether it is Detector/Analysis/Workflow?

nmallick1 (approved):

AngularApp/projects/app-service-diagnostics/src/app/shared-v2/services/link-interceptor.service.ts
puneetg1983: So we should disallow that. 

The DetectorUtils.GetDetectorLink should actually be replaced by a method in the ScriptUtilities class where these kind of validations can be enforced to ensure a consistent coding experience for detector authors.

PR #2121, opened 2023-06-30
Enable workflows back and handle duplication

puneetg1983:
## Overview
This PR enables workflows back and filters out the duplicate items for scenarios where resource types may not be passing query params to AppLens backend

## Type of change
- [x] Bug fix (non-breaking change which fixes an issue)

@SharmaAjay19 - As you know how to reproduce the AKS scenario, can you validate that this PR is indeed fixing the issue that AKS ran in to ?

### DIFF
--- a/AngularApp/projects/app-service-diagnostics/src/app/shared/services/generic-api.service.ts
+++ b/AngularApp/projects/app-service-diagnostics/src/app/shared/services/generic-api.service.ts
@@ -20,11 +20,11 @@ export class GenericApiService {
 
     useLocal: boolean = false;
 
-    useApolloApi:boolean = false;
+    useApolloApi: boolean = false;
 
     effectiveLocale: string = "";
 
-    constructor(private _http: HttpClient, private _armService: ArmService, private _authService: AuthService, private _genericArmConfigService : GenericArmConfigService) {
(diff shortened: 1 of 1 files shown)


SharmaAjay19 (approved):

AngularApp/projects/app-service-diagnostics/src/app/shared/services/generic-api.service.ts
```diff
@@ -43,7 +43,7 @@ export class GenericApiService {
         // Enabling this flag to call both detectors and workflows
         //
 
-        let fetchDetectorsAndWorkflows = false;
+        let fetchDetectorsAndWorkflows = true;
```
hforeste, line 46: Can this be set as an appsetting versus hardcoded?

rekhaswaminathan: or how about enabling only for Web apps for now as we don't know how many other Partners are passing query params or not.

puneetg1983: > Can this be set as an appsetting versus hardcoded?

My past experience says it's extremely hard to update appsetting in UX project as we have too many endpoints. This might become easier once we start using KeyVault but till then it is easier to perform a deployment than update all endpoints.

PR #2162, opened 2023-07-20
Adding vector DB capabilities in AppLens Copilots

SharmaAjay19:
## Overview
This PR adds two capabilities:
1. Integration of AppLens Copilots with documents stored in Vector DB.
2. Support for GPT 3.5 turbo chat models that can be used as an economically viable option where GPT 4 is not needed.

### **Vector DB Integration**
**Context**: Today Applens copilots use GPT-3 and GPT-4 functionalities with prompt engineering. Prompts can get large.

We are now introducing the capability to add content documents to vector db and fetch only the relevant ones to add into the prompt. This will:
1. Provide capability to add much more context to the copilots using larger set of documents.
3. Ensure that only relevant documents are added to the prompt (instead of all documents).
4. Shorten the size of the prompt thus improving performance of the copilots.

**How to use this feature**
For using this feature you need to add two things into your copilot JSON file:
1. DocumentSearchSettings: IndexName - the name of the index or collection where documents are added. This is usually unique for a copilot scenario e.g. RCACopilot can have an index name as rca-copilot-index etc.
image

2. A placeholder <<DOCUMENT_CONTENT_HERE>> inside your prompt in the location where you want the documents to be populated.
image

### **AskAppLens: An application of this feature**
This PR also adds a chatbot for answering general questions about AppLens - **AskAppLens**

This chatbot answers questions from the **AppLens documentation site** (wordpress posts) together with providing reference links for the actual documents.

image

image
image


ApplensBackend/Services/CognitiveSearchService/CognitiveSearchQueryService.cs
```diff
@@ -0,0 +1,43 @@
...
+            if (queryClient != null)
+            {
+                SearchOptions options = new SearchOptions();
+                options.Size = numDocuments;
+                options.IncludeTotalCount = true;
+                options.Select.Add("AdditionalMetadata");
+                SearchResults<CognitiveSearchDocumentWrapper> results = await queryClient.SearchAsync<CognitiveSearchDocumentWrapper>(query, options);
+                if (results != null && results.TotalCount > 0)
+                {
+                    return results.GetResults().ToList().Where(result => result.Score >= minScore).Select(x => JsonConvert.DeserializeObject<CognitiveSearchDocument>(x.Document.AdditionalMetadata)).ToList();
```
ShekharGupta1988, line 37: this .GetResults() method is async ?

SharmaAjay19: The method is sync.

ShekharGupta1988 (approved):

PR #2184, opened 2023-07-31
Adds Dev Keyvault For AppLens

rekhaswaminathan:
## Overview
- Adds keyvault to pull common config settings 
- Unblocks a lot of scenarios where settings are not properly shared across appsettings.Development.json file
- Enabled only on Development Env

## Type of change

- [ ] New feature (non-breaking change which adds functionality)

## Solution description
In StartupPublicAzure, if the hosting env is Development, it adds the Keyvault client and pulls the config. Any other config can be overridden in AppSettings.Development.json

Order of precedence (when running in development mode)

Environment Variables >> appsettings.Development.json >> appsettings.PublicAzure.json >> Keyvault >> appsettings.json

## How Can This Be Tested? 
Once the user is signed in Visual Studio, secrets are auto pulled from keyvault 

- [ ] I have looked over the diffs.
- [ ] There are no errors from my changes on this branch.
- [ ] I have added tests for this feature (if applicable)
- [ ] I have incorporated telemetry into this code to capture and analyze relevant metrics (if applicable)
- [ ] I have requested review on this PR.


ApplensBackend/appsettings.json
```diff
@@ -12,6 +12,7 @@
       }
     }
   },
+  "KeyVault": "https://applenskvdev.vault.azure.net/",
```
michimune, line 15: Did you want to add this to appsettings.Development.json?

rekhaswaminathan: Nope this has to be here as we need to find out which keyvault to connect to. Appsettings.Development.json is loaded at the very end to override any values

ApplensBackend/Configuration/StartupPublicAzure.cs
```diff
@@ -19,11 +23,20 @@ public void AddCloudSpecificServices(IServiceCollection services, IConfiguration
...
 
-        public void AddConfigurations(ConfigurationBuilder builder, IWebHostEnvironment env, string cloudDomain)
+        public void AddConfigurations(ConfigurationBuilder builder, IWebHostEnvironment env, string cloudDomaintempConfig, IConfigurationRoot tempConfig)
         {
             builder.SetBasePath(env.ContentRootPath)
-                    .AddJsonFile("appsettings.json", optional: false, reloadOnChange: true)
-                    .AddJsonFile($"appsettings.PublicAzure.json", optional: false, reloadOnChange: true)
+                    .AddJsonFile("appsettings.json", optional: false, reloadOnChange: true);
+            if (env.IsDevelopment() && !string.IsNullOrWhiteSpace(tempConfig["KeyVault"]))
+            {
```
h-mangipudi, line 31: Just adding a comment to validate that this code looks like it follows order of precedence stated in the pr description:

"Order of precedence (when running in development mode)

Environment Variables >> appsettings.Development.json >> appsettings.PublicAzure.json >> Keyvault >> appsettings.json"

ApplensBackend/Configuration/StartupPublicAzure.cs
```diff
@@ -19,11 +23,20 @@ public void AddCloudSpecificServices(IServiceCollection services, IConfiguration
...
         }
 
-        public void AddConfigurations(ConfigurationBuilder builder, IWebHostEnvironment env, string cloudDomain)
+        public void AddConfigurations(ConfigurationBuilder builder, IWebHostEnvironment env, string cloudDomaintempConfig, IConfigurationRoot tempConfig)
         {
             builder.SetBasePath(env.ContentRootPath)
-                    .AddJsonFile("appsettings.json", optional: false, reloadOnChange: true)
-                    .AddJsonFile($"appsettings.PublicAzure.json", optional: false, reloadOnChange: true)
+                    .AddJsonFile("appsettings.json", optional: false, reloadOnChange: true);
+            if (env.IsDevelopment() && !string.IsNullOrWhiteSpace(tempConfig["KeyVault"]))
```
hforeste, line 30: This is fine but if you want to move away from having to use the IConfigurationRoot maybe you can put the KeyVault appsetting value instead in launchSettings.json https://github.com/Azure/Azure-AppServices-Diagnostics-Portal/blob/f71e01c377e138874e3d8c7bfd62e826fddb254d/ApplensBackend/Properties/launchSettings.json#L18-L23

Then you can just read the KeyVault value from environment variables and you don't have to modify the method signature and all the other stuff for the IConfigurationRoot.

ApplensBackend/Configuration/Startup.cs
```diff
@@ -40,7 +40,7 @@ public Startup(IWebHostEnvironment env)
             }
 
             var builder = new ConfigurationBuilder();
-            cloudEnvironmentStartup.AddConfigurations(builder, env, tmpConfiguration.GetCloudDomain());
+            cloudEnvironmentStartup.AddConfigurations(builder, env, tmpConfiguration.GetCloudDomain(), tmpConfiguration);
```
hforeste, line 43: I know we're passing the IConfigurationRoot because we need the specific keyvault value in order for the individual startup classes to build the configuration object but I don't see too many appsettings that tmpConfiguration will have that the builder will not be able to get so maybe just pass the string value of KeyVault directly instead of the the IConfigurationRoot or follow the suggestion of adding the KeyVault URI in the launchSettings.json file 

https://github.com/Azure/Azure-AppServices-Diagnostics-Portal/blob/f71e01c377e138874e3d8c7bfd62e826fddb254d/ApplensBackend/Properties/launchSettings.json#L18-L23

And you won't have to change the signatures and you can reference the KeyVault URI as an environment variable inside StartupPublicAzure.

hforeste (approved):
Looks good to me, I added some comments that I think could make things a little nicer and simple.

ApplensBackend/Configuration/StartupPublicAzure.cs
hforeste: appsettings.json > KeyVault > appsettings.PublicAzure.json > appsettings.Development.json

hforeste: LOL @h-mangipudi , I order it from least to greatest precedence.

ApplensBackend/Configuration/Startup.cs
rekhaswaminathan: Makes sense. I did not think of launchSettings.json, I will make the change

rekhaswaminathan:
> Looks good to me, I added some comments that I think could make things a little nicer and simple.

Can you please take a look, I made the changes

nmallick1 (approved):

hforeste (approved):
Looks great!

PR #2213, opened 2023-08-23
Update search algorithm for Applens side-nav

XiaoxuHU:
## Overview

This is to use another search algorithm see if it can reduce searching time
## Related Work Item

[Performance - Filtering the side nav on AppLens](https://dev.azure.com/app-service-diagnostics-portal/app-service-diagnostics-portal/_workitems/edit/3181/)
## Type of change


SharmaAjay19 (approved):
And this is why they still ask DS&ALGO in interviews. :P This is great XiaoXu! Are we going to do performance comparison by deploying on staging slot and comparing with prod slot?

XiaoxuHU:
> And this is why they still ask DS&ALGO in interviews. :P This is great XiaoXu! Are we going to do performance comparison by deploying on staging slot and comparing with prod slot?

I deployed into staging before, and let Cristhian and other folks test. Seems like slightly better, so I am thinking to make it into PROD. If we have any complains about slowness regression in the future, we can easily revert back to use previous searching algorithm

